PR #13, opened 2017-03-23
Added "mqtt-lwt" to the MESSAGING group

ppatierno:


ppatierno:
Closed because OpenShift doesn't support grouping based on "Deployment" but only "Service".

PR #18, opened 2017-03-31
add console to admin pod; add support for a sasldb mode

grs:


lulf:
@grs One thing I noticed was that there are no resource limits on the console container. It would be good to add some due to strict defaults in places like openshift online preview account.

grs:
@lulf how does that work exactly? if no limit is specified will it not be allowed to run? The memory used will obviously vary with the size of the router network and the number of the addresses and connections. How would you choose the limit to declare?

lulf:
@grs Setting this limits basically tells kubernetes "this container consumes N units of this resource of my quota". There are typically quota limits on the number of pods, the number of containers within a pod, and  also the max memory available to a pod. The default memory given to containers on some OpenShift instances are 512MB, while the max memory might be 1 GB.

Since the admin pod is shared between 4 containers now, it is good if we can keep the total memory required for the pod below 1GB. The defaults should still work now that the address controller is not part of the admin pod, but I think it's nice to not consume too much resources by default, at least to give a better first impression :)

With the other components I've basically experimented how much it needs in a simple setup. We can always increase the limits if it turns out they are too low. The javascript components seem to be very light weight (ragent is only 64MB limit).

grs:
Ok, so its a way of reducing the memory footprint as compared with the default if nothing is specified, is that right?

grs:
Can a container grow above the initial limit it specifies?

lulf:
Yes, that's what it is. If you go beyond I think  your container might get killed. For Java, we specify a max heap that is half the container memory or something

luksa:
There are resources _requests_ and there are resource _limits_. Requests define the minimum amount the container requires (that amount of CPU/memory is thus guaranteed). Limits define the maximum the container is allowed to consume. The container will never use more CPU than defined in the limit (CPU is throttled). Since memory isn't compressible, if the container tries to use more than the memory limit, it will get killed. 

If limit is not set, it defaults to the request.

PR #28, opened 2017-04-28
added inter-router port to the messaging service

grs:
Needed if you want to exten the router network beyond the local openshift cluster.


lulf:
Looks good!

PR #33, opened 2017-05-26
WIP: Initial work on load balancer

### LINKED ISSUE #32: Create Kubernetes config with service type=LoadBalancer
This makes it easier to deploy EnMasse on public cloud providers that do not have an ingress controller.

lulf:
This fixes #32


include/enmasse-kubernetes.jsonnet
```diff
@@ -10,7 +10,7 @@ local flavorConfig = import "flavor.jsonnet";
     "items": [ templateConfig.generate(use_sasl, with_kafka, false),
                addressController.deployment("enmasseproject/address-controller", "false", 0),
                restapiRoute.ingress(""),
-               addressController.service,
+               addressController.service("LoadBalancer"),
```
ppatierno: As you proposed during our chat should we have a different kubernetes tenmplate for cloud provider (i.e. Azure) so that the current one using minikube can work with ingresses ?

lulf: Yes, I'm planning to do that also, as I think most people want to use Ingress in a production setup, but need to think a bit about if we should make a completely separate template, or make the ingress part an 'add-on' template.

lulf: @ppatierno I'm actually wondering if the appropriate thing to do here is to create additional services that have type=LoadBalanced. The messaging service for instance, contains 2 additional ports that are used for 1. interrouter communication and 2. brokers. I'm pondering whether it is a feature to have them exposed or not.

lulf: @ppatierno Actually, I like this approach:

1. Let the current services stay as they are, they can be considered 'internal'
2. We create a new set of services explicitly targeted for exposing, 'messaging-external', 'mqtt-extrnal' and 'console-external'. These additional services will be put in a separate file that can be kubectl create'd for the non-ingress use cases. 

We can consider renaming the 'internal' services to 'messaging-internal' etc as well, which would perhaps match their actual use better.

wdyt?

lulf: I'm going to propose a new directory structure for the templates to cope better with 'addon' templates:

install/openshift/enmasse.json
install/openshift/enmasse-with-kafka.json
install/openshift/enmasse-with-sasldb.json

install/kubernetes/enmasse.json
install/kubernetes/addons/external-lb.json

PR #38, opened 2017-06-08
Add multitenancy support to console

### LINKED ISSUE #37: Support multitenancy for console
Right now the console connects to the address-controller using the service environment. These environment variables are not available when they run in separate namespaces. 

Work has already been done to prepare the console to use API_SERVER_HOSTNAME environment if available. The remaining work is to expose this environment variable so that it will be used by the console.

lulf:
This fixes #37

### DIFF
--- a/include/admin.jsonnet
+++ b/include/admin.jsonnet
@@ -87,8 +87,8 @@ local common = import "common.jsonnet";
                         "value": "5672"
                       },
                       {
-                        "name": "API_SERVER_HOSTNAME",
-                        "value": "${API_SERVER_HOSTNAME}"
+                        "name": "ADDRESS_SPACE_SERVICE_HOST",
+                        "value": "${ADDRESS_SPACE_SERVICE_HOST}"
                       },
                       {
                         "name": "INSTANCE",
--- a/include/admin.jsonnet
+++ b/include/admin.jsonnet
@@ -85,7 +85,16 @@ local common = import "common.jsonnet";
                       {
                         "name": "CONFIGURATION_SERVICE_PORT",
                         "value": "5672"
-                      }]),
+                      },
+                      {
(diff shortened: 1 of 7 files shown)


include/admin.jsonnet
```diff
@@ -85,7 +85,16 @@ local common = import "common.jsonnet";
                       {
                         "name": "CONFIGURATION_SERVICE_PORT",
                         "value": "5672"
-                      }]),
+                      },
+                      {
+                        "name": "API_SERVER_HOSTNAME",
```
grs: Minor suggestion on naming... perhaps ADDRESS_CONTROLLER_HOST would be more obvious? Or else ADDRESS_SERVICE_HOST (to emulate loosely the kubernetes naming convention)?

lulf: So my reason for using this name is that the API  would eventually be part of the enmasse API server which would be a 3rd party API server in kubernetes. So the name is following the kubernetes convention in that sense.

lulf: So I think ADDRESS_CONTROLLER_HOST is wrong in that sense. ADDRESS_SERVICE_HOST is better, but it's not only about addresses anymore... But i agree _HOST is better than _HOSTNAME

grs: Fair points, and it is only a name so not a huge issue at this point. API_SERVER is a little too ambiguous/generic for my taste (though as an env var I can certainly live with it :). What about MESSAGING_CONTROL or MESSAGING_ADMIN or something a little more descriptive of its actual function?

lulf: MESSAGING_ADMIN_API_HOST  ?

grs: I prefer MESSAGING_ADMIN_SERVICE_HOST, but I would be fine with your suggestion .

PR #42, opened 2017-06-26
*WIP DO NOT MERGE* Initial proposal of new conceptual model

lulf:
I propose that we use this document as a base for discussing and refining the conceptual model/address model in EnMasse. I've written an initial proposal based on our previous discussions and the currently supported address types in EnMasse.

### DIFF
--- a/documentation/address-model/resource-definitions.md
+++ b/documentation/address-model/resource-definitions.md
@@ -22,24 +22,21 @@ The `AddressSpace` resource defines the type of address space and endpoint for c
         "type": "standard", // Optional. Default is 'standard'
         "plan": "small", // Required. This is the aggregate 'size' of the address space (i.e. controls how
                          // many addresses and how many resources each address may consume
-        "endpoints": [ // Optional. If not provided, the default generated hostnames will be provided
-            {
-                "type": "messaging", // Required. The available types are defined by the address space type
+        "endpoints": { // Optional. If not provided, the default generated hostnames will be provided
+           "messaging" { // Required. The available types are defined by the address space type
                 "host": "messaging.example.com", // Optional. The host name for this endpoint to use for routing by the load balancer
                 "cert": { // Optional. Self-signed certificates are generated if not specified
                     "provider": "secret", // Provider of certs may be through secrets or ACME. For 'secret', the secret name must be passed
                     "secretName": "mysecret"
                 }
             },
-            {
-                "type": "mqtt",
+            "mqtt": {
                 "host": "mqtt.example.com"
             },
-            {
-                "type": "console",
+            "console": {
                 "host": "console.example.com"
             }
-        ]
+        }
     }
 }
 ```
--- a/documentation/address-model/resource-definitions.md
+++ b/documentation/address-model/resource-definitions.md
@@ -23,7 +23,7 @@ The `AddressSpace` resource defines the type of address space and endpoint for c
         "plan": "small", // Required. This is the aggregate 'size' of the address space (i.e. controls how
                          // many addresses and how many resources each address may consume
         "endpoints": { // Optional. If not provided, the default generated hostnames will be provided
-           "messaging" { // Required. The available types are defined by the address space type
+           "messaging": { // Required. The available types are defined by the address space type
                 "host": "messaging.example.com", // Optional. The host name for this endpoint to use for routing by the load balancer
                 "cert": { // Optional. Self-signed certificates are generated if not specified
                     "provider": "secret", // Provider of certs may be through secrets or ACME. For 'secret', the secret name must be passed
--- a/documentation/address-model/resource-definitions.md
+++ b/documentation/address-model/resource-definitions.md
@@ -1,0 +1,69 @@
+# EnMasse resource definition
+
+This document defines the resources consumed and produced by APIs and can be considered a base for writing the schema for the data formats. The [conceptual model](model.md) defines address spaces and addresses which translates to these resources.
+
+The resources defined herein may be created, deleted and updated through different APIs. Either the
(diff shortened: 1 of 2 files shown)


lulf:
@rgodfrey @grs @ppatierno 

I've added another file, resource-definitions.md, that attempts to specify the resources that can be manipulated through the API. There are a few parts of the address space that we've not discussed yet like how the endpoints are configured. 

Below is a proposal that maps to what we have today. The available endpoint names are determined by the address space, and is used by the implementation to create the route to the internal service name (today we have messagingHost, mqttHost and consoleHost on the v3.Instance resource). I think that ideally there would be only one endpoint for an address space (which supports multiple protocols), but this is not possible today. There is also the question on what we would do with the console endpoint.

Certificates for the endpoint can be specified using different cert providers. 'secret' would be one where the user has manually put the certs into a secret, whereas one can envison 'acme' as another provider where acme account info could be specified.

grs:
I would not aim for S, M & L as the values for all plans yet (it might be nice to get there at some point for consistency, but we have a lot more thinking to do on how plans would actually work). I think it is important to state in the model that the plans available depend on the particular address- and address-space- type in question.

I think it might be worth explicitly stating that there are descriptions for address- and address-space- types (and plans). (Perhaps short and long descriptions even).

As a first step (and this may or may not be appropriate detail for your document), perhaps we could have a 'standard' plan available for the standard address-space type and the direct address types (these would in reality all be separate plans, with the same name but perhaps different descriptions).

documentation/address-model/resource-definitions.md
```diff
@@ -0,0 +1,65 @@
...
+    "kind": "AddressSpace",
+    "metadata": {
+        "name": "myspace" // Required
+    },
+    "spec": {
+        "type": "standard", // Optional. Default is 'standard'
+        "plan": "small", // Required. This is the aggregate 'size' of the address space (i.e. controls how
+                         // many addresses and how many resources each address may consume
+        "endpoints": { // Optional. If not provided, the default generated hostnames will be provided
+           "messaging": { // Required. The available types are defined by the address space type
```
grs: I don't understand why 'messaging' is required, when the endpoints aren't and neother are either of the fields within the messaging element.

documentation/address-model/resource-definitions.md
```diff
@@ -0,0 +1,65 @@
...
+```
+{
+    "apiVersion": "enmasse.io/v1", // Required
+    "kind": "Address", // Required
+    "metadata": {
+        "name": "myqueue", // Required
+        "address_space": "myspace" // Optional. Defaults to 'default' address space
+    },
+    "spec": {
+        "address": "my_queue", // Optional. Default is .metadata.name
```
grs: What is the distinction between address and name (from metadata)? I.e. if they don't match, what is each used for?

documentation/address-model/resource-definitions.md
```diff
@@ -0,0 +1,65 @@
...
+    "apiVersion": "enmasse.io/v1",
+    "kind": "AddressSpace",
+    "metadata": {
+        "name": "myspace" // Required
+    },
+    "spec": {
+        "type": "standard", // Optional. Default is 'standard'
+        "plan": "small", // Required. This is the aggregate 'size' of the address space (i.e. controls how
+                         // many addresses and how many resources each address may consume
+        "endpoints": { // Optional. If not provided, the default generated hostnames will be provided
```
grs: In terms of the model, would we consider the endpoint details to be defined within the properties?

documentation/address-model/resource-definitions.md
lulf: If we use TPRs to represent this format, there will be a restriction on length and allowed characters in the .metadata.name. The name could be a mix of uuid and sanitized address, while the address would contain the 'raw' address and be used when propagating the info further to router agent for instance.

documentation/address-model/resource-definitions.md
lulf: Good point, it should be possible to have an empty "endpoints" map or having it partially defined.

documentation/address-model/resource-definitions.md
lulf: I don't think we've considered properties on an address space, but it could be useful if we can think of other uses of it perhaps.

documentation/address-model/resource-definitions.md
grs: Makes sense. So the address is always the logical address, it is just populated automatically by the name if left blank, and the name is a kubernetes specific attribute (i.e. part of the physical data not the logical model).

One thing related to your comment, are TPRs the right way to go? I notice that the servicecatalog uses the approach of having its own api-server (I think that may be registered with the central api server at some point) and is being talked about as being an emerging example of using the kubernetes extension points. (kubectl works on the servicecatalog). Maybe another option for us?

documentation/address-model/resource-definitions.md
grs: True, we didn't explicitly have them for address-spaces. If these are to be passed from the user then I think it makes sense to add them to the model, and doing so as properties would allow us to evolve without breaking the model.

documentation/address-model/resource-definitions.md
lulf: Actually, I think TPR (being renamed to CRD these days...) and api-servers complement each other. TPRs define the resource types, and you can use these resources as the 'data types' internally. It's not clear from the docs what the relationship between TPRs and the API server is. Found https://kubernetes.io/docs/concepts/api-extension/custom-resources/

@luksa Perhaps you have better definition :)

lulf:
Closing, as we've come to an agreement and the documents have been committed.

PR #43, opened 2017-07-03
Define a subject for AMQP messages sent out by configserv 

grs:
(to identify the format of the body)


grs:
@rgodfrey @lulf

lulf:
Looks good!

PR #45, opened 2017-07-04
update model.md

tkratky:
missing http:// prefix for external links and wrong link to "MQTT over AMQP" documentation


lulf:
Thanks @tkratky !

PR #71, opened 2017-07-04
Skip creating secret for instance if it already exists

### LINKED ISSUE #70: Precreating instance secret causes exception in address controller
See https://www.redhat.com/archives/enmasse/2017-July/msg00004.html for reference

lulf:
This fixes #70


lulf:
The fix is now in the latest snapshot: https://dl.bintray.com/enmasse/snapshots/latest/enmasse-latest.tgz

PR #74, opened 2017-07-06
Use new address model for address spaces

lulf:
* Remove Instance*
* Rename Instance -> AddressSpace
* Add SecretCertProvider
* Rename instancecontroller -> standardcontroller
* Merge address controller into standardcontroller

* Change serialization code and introduce serializers on model type instead
* Configure resteasy with custom object mapper
* Configure endpoints according to address space
* Remove creation of routes in templates and create them in address controller


lulf:
@grs @rgodfrey I don't expect you to review all this code before I merge it, but at least you can see whats going on.

PR #80, opened 2017-07-24
Added authentication to OSB API

luksa:


luksa:
@lulf Can you check the credentials for the image registry? (that's why the CI build is failing)

lulf:
The auth error is a travis issue afaict. Looks good to me. @rgodfrey @grs How does this fit with the authentication design?

grs:
Does this mean there is one user defined that can use the service broker? How does that work? Is the service catalog then doing some authorisation based on the actual openshift user requesting the service, and then using this single identity to access the service broker instance?

luksa:
When you register a broker with the Service Catalog, you supply a Secret containing a username and password. The catalog then uses this to authenticate with the broker through Basic HTTP auth (token based auth is coming soon). 

For retrieving the catalog, that's the only auth the catalog sends to the broker. When provisioning and binding a service, the broker also sends information about the user performing the operation (the username, namespace, service account name).

PR #84, opened 2017-07-31
Added proposal for generic spec about QoS and retain in AMQP 1.0

ppatierno:


documentation/mqtt-over-amqp/docs/amqp_qos_retain.md
```diff
@@ -0,0 +1,94 @@
...
+* a new "desired-capabilities" named `qos-policy` during the link attachment from the receiver
+* a link "property" specifying the value for the above `qos-policy`
+* the source should support such "capability"
+* a link "property" specifying the value for the desired QoS on receiving
+
+The `qos-policy` could be something like :
+
+* _exact_ : the receiver wants to receive messages published only with the same QoS as the original message
+* _enforce minimum_ : the receiver wants to receive message published with QoS = min(published QoS, requested QoS)
+* _enforce maximum_ : the receiver wants to receive message published with QoS = max(published QoS, requested QoS)
```
grs: To avoid ambiguity, I would rephrase these a little, e.g.  "the receiver wants messages to be transferred to it using the same QoS as used in the original transfer from publisher to intermediary" (I know that sound long-winded, and probably a more succinct wording could be found, the key is to make it clear that it is not filtering messages in anyway, merely determining the QoS with which they are delivered).

documentation/mqtt-over-amqp/docs/amqp_qos_retain.md
```diff
@@ -0,0 +1,94 @@
...
+With AMQP protocol we could have :
+
+* a "delivery annotation" named `x-opt-retain` on the sent message
+* a "capability" on the target which describes that it can retain messages
+
+Here the problem is how to encode an "empty" message for deleting the last retain message because as AMQP 1.0 spec :
+
+"The body consists of one of the following three choices: one or more data sections, one or more amqp-sequence
+sections, or a single amqp-value section."
+
```
rgodfrey: One option here would be to have a second annotation 'x-opt-retain-empty' used on a message which you wish to use to clear the current retained message.  The actual body of such a message would be ignored

documentation/mqtt-over-amqp/docs/amqp_qos_retain.md
```diff
@@ -0,0 +1,94 @@
...
+* _link level_ , when the receiver attaches a link specifying the desired QoS
+* _message level_, when the sender sends the message with such QoS specified in the message itself
+
+At _link level_, we could have :
+
+* a new "desired-capabilities" named `qos-policy` during the link attachment from the receiver
+* a link "property" specifying the value for the above `qos-policy`
+* the source should support such "capability"
+* a link "property" specifying the value for the desired QoS on receiving
```
grs: My initial reaction here is that the AMQP spec already defines a way for peers to select a QoS for the link, through the sender- and receiver- settlement modes. While a sender-settlement mode of 'mixed' indicates a per-message choice, there isn't a way - as far as I can tell - to decide on the receiver settlement mode per message. That means that the same link cannot be used for receiving both at-most-once/at-least-once aka Qos 0/1 and exactly-once aka QoS 2 messages.

Personally, I'm not sure I see sufficient value in the MQTT approach to make it worth trying to define something this complex. The primary value I see is that the receiver is made aware of the fact that a message may have been published with a QoS less than it itself negotiated, which may effect the end-to-end delivery guarantee.

Perhaps a simpler solution would just be to provide some way of indicating the QoS/delivery guarantee under which the original publication was made. The receiver would use the same pattern for its own acknowledgement of all messages, but would be able to determine where the expectations between itself and the publisher differ.

documentation/mqtt-over-amqp/docs/amqp_qos_retain.md
grs: It could be an AMQP null value.

documentation/mqtt-over-amqp/docs/amqp_qos_retain.md
rgodfrey: > My initial reaction here is that the AMQP spec already defines a way for peers to select a QoS for the link, through the sender- and receiver- settlement modes. While a sender-settlement mode of 'mixed' indicates a per-message choice, there isn't a way - as far as I can tell - to decide on the receiver settlement mode per message.

I saw this too, and am trying to rack my brains as to why it is worded the way it is.  In general I think that either side should be able to settle any time they like.  

> Personally, I'm not sure I see sufficient value in the MQTT approach to make it worth trying to define something this complex. The primary value I see is that the receiver is made aware of the fact that a message may have been published with a QoS less than it itself negotiated, which may effect the end-to-end delivery guarantee.

The issue with doing things this way is that we would be breaking the MQTT contract.  If the message was published with QoS 0, then there is effectively an end-to-end at most once guarantee; whereas if we effectively use QoS 1 for the subscriber then that end-to-end contract is broken (since it becomes at-least-once).

grs: > The issue with doing things this way is that we would be breaking the MQTT contract. If the  essage was published with QoS 0, then there is effectively an end-to-end at most once guarantee; whereas if we effectively use QoS 1 for the subscriber then that end-to-end contract is broken (since it becomes at-least-once).

In terms of bridging MQTT over AMQP, the bridge could recognise the original QoS of the publication and alter how it sends the message to the MQTT consumer.

I guess I'm not clear on what the objective here is. With e.g. retain or lwt its clear that it is about exposing some useful semantic defined by MQTT to general AMQP use. I assumed this was something similar (though to me the feature in question is not all that compelling). If it is just about not losing information when bridging MQTT->AMQP->MQTT, that is a different thing.

rgodfrey: > In terms of bridging MQTT over AMQP, the bridge could recognise the original QoS of the publication and alter how it sends the message to the MQTT consumer.

Sure, we could ensure in the case of such a message that the transfer from AMQP -> Bridge is settled before the message is transferred from Bridge -> MQTT client (thus ensuring that the end-to-end at most once contract is guaranteed).

Personally, if I was imagining potentially useful semantics of end-to-end QoS, the MQTT ones are not what I would choose (for instance I could sort of see a use for defining a contract that the QoS for the message MUST be respected for all hops across the network... but that's not what MQTT does).

If we consider the primary (only) use case for this annotation is for the mqtt bridge, then I don't think we need to define anything on the links, but then the annotation is really not anything generic, but only makes sense for messages you believe might be travelling over the mqtt-bridge (since that is the only intermediary that is going to act on them)

ppatierno: > In terms of bridging MQTT over AMQP, the bridge could recognise the original QoS of the publication and alter how it sends the message to the MQTT consumer.

This is what already happens in the MQTT gateway.

rgodfrey: > This is what already happens in the MQTT gateway.

Sure - so, as above, that means that the QoS annotation is basically just a signal to the MQTT bridge and not part of a "generic" AMQP definition; or, at most, the generic AMQP definition for the annotation is that it is "informational" but that some intermediaries may choose to act on this information in terms of how the message is transferred onward through the network.

ppatierno: The first intention was something like that then we move to think about having a sort of mechanism to define a policy for and end-to-end delivery through more intermediaries and so doing something like MQTT does. Having the receiver specifying a policy like the proposal should be on this path but then what @grs pointed out about the impossibility to settle in different way from the receiver PoV.

grs: > If we consider the primary (only) use case for this annotation is for the mqtt bridge, then I don't think we need to define anything on the links, but then the annotation is really not anything generic, but only makes sense for messages you believe might be travelling over the mqtt-bridge (since that is the only intermediary that is going to act on them)

Agreed.

documentation/mqtt-over-amqp/docs/amqp_qos_retain.md
ppatierno: What if the client "will" message brings headers which are meaningful but not body ? Maybe the additional annotation `x-opt-retain-empty` proposed by @rgodfrey is a good solution.

documentation/mqtt-over-amqp/docs/amqp_qos_retain.md
gemmellr: > > My initial reaction here is that the AMQP spec already defines a way for peers to select a QoS for the link, through the sender- and receiver- settlement modes. While a sender-settlement mode of 'mixed' indicates a per-message choice, there isn't a way - as far as I can tell - to decide on the receiver settlement mode per message.

> I saw this too, and am trying to rack my brains as to why it is worded the way it is. In general I think that either side should be able to settle any time they like.

Is this what rcv-settle-mode on transfer is for?
http://docs.oasis-open.org/amqp/core/v1.0/os/amqp-core-transport-v1.0-os.html#type-transfer

rgodfrey: @gemmellr I realise it has been about a month since you left this comment, but for some reason I missed it until now :-)

To me the issue is that the way the attach definition is worded around "rcv-settle-mode": 

> When set at the receiver this indicates the actual settlement mode in use" in combination 

with the definition of the mode "second" : 

> The receiver will only settle after sending the disposition to the sender and receiving a disposition indicating settlement of the delivery from the sender.

Makes it appear that the receiver never has discretion to actually settle first.  As per your reference to the rcv-settle-mode on transfer, obviously here receiver is being instructed to settle first (even though the mode is second).  And, in general, I think either side can choose to settle a delivery at any point in time... therefore the receiver choosing rcv-settle-mode "second" is not precluding the receiver actually settling first where either it is instructed to by the sender, or because it has it's own reasons why it needs to settle immediately upon achieving a terminal state.  The wording could do with clarifying though,

gemmellr: @rgodfrey On the point about what attach says, I believe thats just related to the fact the peer receiving the attach might reply with different value depending on what it supports (i.e something that doesnt support 'second' probably shouldn't respond with 'second' but could respond with 'first'), with the next sentence there essentially also covering that.

The definition for Transfer seems to me to say its value if set overrides the link attach value but is only useful at all for setting 'first' on a transfer on a link originally attached with 'second'. Thus you could use mixed+second to support sending at-most-once settled, exactly-once unsettled, and at least once unsettled transfers on the link.

I don't think the spec really allows for either side to settle at any point in general when 'second' is used, with the above instead being the mechanism to make particular messages use 'first', though it seems to me there are some obvious cases there would effectively be no choice for a receiver but to do so....so, yeah..

lulf:
@ppatierno Can this be merged/closed?

lulf:
Closing due to inactivity. Reopen if discussion is to continue (or move discussion to an issue).

PR #85, opened 2017-08-02
Use new address model in getting started documentation

lulf:


documentation/getting-started/openshift.md
```diff
@@ -120,64 +120,57 @@ EnMasse is configured with a set of addresses that you can use for messages. Cur
...
+      "spec": {
+        "type": "multicast"
+      }
+    }
+  ]
 }
 ```
 
 Each address that set store-and-forward=true must also refer to a flavor. See below on how to create
 your own flavors. To deploy this configuration, you must currently use a http client like curl:
```
ppatierno: Does this statement still make sense ? It's speaking about "store-and-forward" which isn't in the JSON and about "flavor".

lulf: Agreed. I've updated the doc now.

PR #87, opened 2017-08-02
Use common base image with node runtime

lulf:
This saves some time and space for building docker images that all
install the same RPMs.

@grs Wdyt? Some travis builds hang on RPM install, and I was hoping this would fix that as well.

### DIFF
--- a/console/Dockerfile
+++ b/console/Dockerfile
@@ -1,9 +1,4 @@
-FROM centos:7
-
-RUN yum -y update  \
-    && yum -y install epel-release cyrus-sasl-lib \
-    && yum -y install nodejs npm \
-    && yum clean all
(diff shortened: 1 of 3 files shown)


grs:
Yes, seems sensible to me... where is the base image defined/built?

grs:
Ignore last question, I see a notification of a new repo in my mail

grs:
One point is that the current base doesn't include the cyrus-sasl-lib rpm, which is needed for the sasldb based user management in the console. Though ultimately we will remove that, perhaps we could add that back in to the base image for now and remove it when we have the keycloak based solution working>

lulf:
@grs Agreed, will add that.

PR #89, opened 2017-08-03
FIX build.gradle - docker authentication

tkratky:
Default dockerUser and dockerPassword changed from "unset" to null due to authentication fails. Now dockerUser and dockerPassword should be set only via System properties.


lulf:
LGTM!

PR #104, opened 2017-08-08
Create none auth service and enable auth for router and console

rgodfrey:


lulf:
@grs This is without TLS on the auth-service connection, but we plan to address that in a later task.

PR #108, opened 2017-08-09
Replace oadm with oc adm

luksa:


lulf:
Looks good, filed an issue for the unstable test that caused the build to  fail: https://github.com/EnMasseProject/enmasse/issues/109

PR #114, opened 2017-08-10
Add -s switch to deploy-openshift.sh, which sets up the Service Catal…

luksa:
…og when used with the all-in-one option


lulf:
Great, thanks Marko!

PR #126, opened 2017-08-11
Updated to use host in favour of the deprecated addr.

grs:
This avoids https://issues.apache.org/jira/browse/DISPATCH-804 (though that is now resolved).


lulf:
Thanks for resolving!

PR #132, opened 2017-08-15
Update router

grs:
This uses the router base image from https://github.com/EnMasseProject/router-docker-base/pull/2, which should be updated to the proper tag once that is merged. This is an experimental PR to test the build status with just these changes (works locally for me now).


lulf:
@grs I noticed you change the base image in this PR to be enmasseproject/qdrouterd and not enmasseproject/qdrouterd-base. The router-docker-base still uses qdrouterd-base, so you might want to change one of them.

I also fixed a bug in the router-docker-base travis build to do the tagging properly, so enmasseproject/qdrouterd-base:1.0.0-alpha2 is now built: https://hub.docker.com/r/enmasseproject/qdrouterd-base/tags/

lulf:
Changed it to qdrouterd-base just to see how the build goes :)

grs:
@lulf, thanks! Silly me :-(

lulf:
@grs LGTM. Thank for digging and finding the cause!

PR #148, opened 2017-08-23
Enable TLS for authentication services

lulf:
This enables TLS for authentication services and enables it in the router authentication service plugin for the TLS-enabled listener.


lulf:
@grs @rgodfrey Propose that we merge this now that things are in a working state.

rgodfrey:
+1 on merging

console/lib/auth_service.js
```diff
@@ -15,14 +15,20 @@
  */
 'use strict';
 
+var path = require('path');
+var fs = require('fs');
 var rhea = require('rhea');
 var Promise = require('bluebird');
+var ca_dir = "/opt/console/authservice-ca";
```
grs: I think it would be better to pass this in, rather than hardcode it. Otherwise it makes the code hard to use outside the specific pod.

console/lib/auth_service.js
```diff
@@ -15,14 +15,20 @@
...
 
 function authenticate(user) {
     return new Promise(function(resolve, reject) {
         var options = {
-            "host": process.env.AUTHENTICATION_SERVICE_HOST,
-            "port": process.env.AUTHENTICATION_SERVICE_PORT
+            host: process.env.AUTHENTICATION_SERVICE_HOST,
+            port: process.env.AUTHENTICATION_SERVICE_PORT,
+            transport: 'tls',
+            ca: [ fs.readFileSync(path.resolve(ca_dir, 'tls.crt')) ],
```
grs: Possibly worth a try/catch around the reading of the cert to make it obvious is there is any problem there(?)

grs:
See https://github.com/EnMasseProject/enmasse/pull/152 for a suggested fix to keep the tests working (and add coverage for tls).

lulf:
Thanks for adding the test @grs !

PR #157, opened 2017-08-25
Use tls for mqtt lwt

lulf:


lulf:
The plan ( @rgodfrey  correct me if i'm wrong) is to do TLS-only between all components.

ppatierno:
And because I'll move the mqtt-lwt to be a sort of amqp-presence/connectivity service usable outside EnMasse as well, do we want that ? I'm just asking :-) 
I could have same problem with amqp-kafka bridge which is currently in the EnMasse project but could be used outside and in that case I would prefer to have a choice to have TLS (if inside EnMasse) or not having TLS (if someone doesn't want it when outside EnMasse). Any thoughts ?

lulf:
Thats a good point. I guess it would have to have some more configuration knobs to support both outside-enmasse and within-enmasse deployment. I that we do that as part of the 'connectivity service' rewrite/refactor. 

There is not anytning enmasse-specific about the TLS support. It just requires client key + client cert + trusted CA cert.

rgodfrey:
@ppatierno I think that the rule should be "if it is a component running within enmasse then it uses TLS / the distribute client certs.  If it is a service connecting in from outside of enmasse, then it should be provisioned its own credentials and connect via an appropriately secure mechanism".  For the the lwt service itself, it should obviously be configurable how it sets up its connections (maybe including whether it acts as a server that the network connects into, or it initiates the connection into the router network).

ppatierno:
@rgodfrey @lulf so we are assuming that current mqtt-lwt (futue amqp-presence/connectivity) will run only inside EnMAsse ? If yes the PR LGTM otherwise we need a way to configure TLS or not or maybe doing that when I'll change from mqtt-lwt to amqp-presence/connectivity ?

rgodfrey:
@ppatierno no - I think lwt should be capable of running outside enmasse

I think it should be configurable so thatif we want to deploy it in enmasse then we can configure it to use TLS / our certificate.  If someone want to run it outside enmasse then they can configure it in a different way

lulf:
@ppatierno I can modify it so that if it doesn't find certDir or certDir is not set, it will not use TLS.

ppatierno:
@lulf I'm ok with that ! :) Sorry for my comments but I was just thinking about the component usage outside of EnMasse ... so my doubts on forcing TLS.

PR #159, opened 2017-08-28
Run without junit reporter by default

lulf:


lulf:
@grs This also defines runMocha at toplevel so that it doesn't need to be defined in each JS module.

PR #164, opened 2017-08-28
switch router->ragent traffic over to TLS

grs:


templates/install/openshift/enmasse-with-sasldb.yaml
```diff
@@ -40,6 +40,8 @@ objects:
               value: "${ADDRESS}"
             - name: CLUSTER_ID
               value: "${CLUSTER_ID}"
+            - name: PN_TRACE_FRM
+              value: '1'
```
lulf: Perhaps we should remove this file, it is not used (or usable for that sake) anymore.

ragent/tls_options.js
```diff
@@ -0,0 +1,45 @@
+/*
+ * Copyright 2016 Red Hat Inc.
```
lulf: Update year.

ragent/tls_options.js
```diff
@@ -0,0 +1,45 @@
...
+
+var path = require('path');
+var fs = require('fs');
+
+function get_client_options(config) {
+    var options = config || {};
+    var cert_dir = (process.env.CERT_DIR !== undefined) ? process.env.CERT_DIR : '/etc/enmasse-certs';
+    var ca_path = path.resolve(cert_dir, 'ca.crt');//TODO: allow setting via env var
+    var crt_path = path.resolve(cert_dir, 'tls.crt');//TODO: allow setting via env var
+    var key_path = path.resolve(cert_dir, 'tls.key');//TODO: allow setting via env var
```
lulf: IMHO setting the CERT DIR is sufficient, and in doing testing so far I've not had any use for setting them individually.

I'd say just fix them or remove them?

ragent/tls_options.js
grs: done

ragent/tls_options.js
grs: done

templates/install/openshift/enmasse-with-sasldb.yaml
grs: done

PR #168, opened 2017-08-29
Require clients to authenticate on internal TLS based listeners.

grs:


lulf:
The QueueTest and PublishTest(mqtt) fails with this, and I don't think its instability in the test from the looks of it. Could there be something missing for the servers that connect to that router-container port?

grs:
The issue here is the brokers connector service. It appears not to use a SASL layer, which is why the authenticatePeer: yes change causes it to fail to connect. However that is the only way to ensure that the router requests client certs to be presented.

lulf:
Closing this, as I've cherry picked your commit into #175

PR #169, opened 2017-08-29
Use TLS for 'router-container' AMQP port in Mqtt-Lwt

lulf:


mqtt/mqtt-lwt/src/main/java/enmasse/mqtt/MqttLwt.java
```diff
@@ -129,22 +131,7 @@ private void connect(Future<Void> startFuture) {
...
-                        .setHostnameVerificationAlgorithm("")
-                        .setPemTrustOptions(new PemTrustOptions()
-                                .addCertPath(new File(certDir, "ca.crt").getAbsolutePath()))
-                        .setPemKeyCertOptions(new PemKeyCertOptions()
-                                .addCertPath(new File(certDir, "tls.crt").getAbsolutePath())
-                                .addKeyPath(new File(certDir, "tls.key").getAbsolutePath()));
-            };
-
-            this.client.connect(internalConnOptions, this.messagingServiceHost, port, done -> {
+            this.client.connect(options, this.messagingServiceHost, options.isSsl() ? messagingServiceAmqpsNormalPort : messagingServicePort, done -> {
```
grs: My 2 cents: 'messagingServicePort'/messagingServiceInternalPort are slightly confusing (both should be internal). 

Suggestion:
messagingServiceAmqpBrokerPort,
messagingServiceNormalPort,
messagingServiceAmqpBrokerPortTLS and
messagingServiceNormalPortTLS

grs: (I'd probably also consider dropping some or all of messagingService in these, and using the saving to change from BrokerPort to RouteContainerPort)

mqtt/mqtt-lwt/src/test/resources/qdrouterd.conf
```diff
@@ -50,8 +50,9 @@ listener {
 # listener for services (i.e. Last Will and Testament and Subscription services)
 listener {
   host: 0.0.0.0
-  port: 55673
-  authenticatePeer: no
```
grs: Without the authenticatePeer, you don't need to use a SASL layer. I recognise this is only a test though, so perhaps not significant here.

mqtt/mqtt-lwt/src/main/java/enmasse/mqtt/MqttLwt.java
lulf: The reason they are named this way is so that they match the environment variables defined by the service. 

However, I think a workaround is to add a mapping in the deployment specification, so I will do that.

lulf: could be that changing the annotation on the setters is sufficient( spring magic)

mqtt/mqtt-lwt/src/test/resources/qdrouterd.conf
lulf: I can add it for completeness, though I think we want to refactor the test to run without the routet

PR #173, opened 2017-08-30
Resolve the default authentication service type based on environment

lulf:


lulf:
I expect this to fail building, as systemtest clients will not be authenticated. I'm pondering if the ideal way to do this is to skip deploying keycloak for the systemtests until we have systemtest support for creating the roles etc.

systemtests/src/main/java/enmasse/systemtest/OpenShift.java
```diff
@@ -117,7 +117,7 @@ public int getExpectedPods() {
         if (environment.isMultitenant()) {
             return 5; // admin, qdrouterd, subscription, mqtt gateway, mqtt lwt
         } else {
-            return 9; // address-controller, keycloak, none-authservice, admin, qdrouterd, subscription, mqtt gateway, mqtt lwt
+            return 7; // address-controller, none-authservice, admin, qdrouterd, subscription, mqtt gateway, mqtt lwt
```
grs: Minor question: Not clear what the change is here. Only one thing - keycloak - is removed from the comment, but the count is decremented by 2. (Other way of asking the question is why it was 9 previously when only 8 items were listed in the comment).

lulf: keycloak and keycloak-controller. The comment was out of date :)

PR #175, opened 2017-08-30
Queue scheduler server tls

lulf:


admin/queue-scheduler/src/main/java/io/enmasse/queue/scheduler/QueueScheduler.java
```diff
@@ -65,7 +72,19 @@ private static String getGroupId(ProtonConnection connection) {
 
     @Override
     public void start() {
-        server = ProtonServer.create(vertx);
+        ProtonServerOptions options = new ProtonServerOptions();
+        if (certDir != null) {
+            options.setSsl(true)
+                    .setClientAuth(ClientAuth.NONE)
```
grs: What does ClientAuth.NONE do? We need clients to authenticate using TLS, right?

admin/queue-scheduler/src/main/java/io/enmasse/queue/scheduler/Main.java
```diff
@@ -23,11 +23,12 @@ public static void main(String [] args) {
...
         String configHost = getEnvOrThrow("CONFIGURATION_SERVICE_HOST");
         int configPort = Integer.parseInt(getEnvOrThrow("CONFIGURATION_SERVICE_PORT"));
-        int listenPort = 55667;
+        int listenPort = Integer.parseInt(getEnvOrThrow("LISTEN_PORT"));
 
         QueueScheduler scheduler = new QueueScheduler(
                 connection -> Artemis.create(vertx, connection),
-                listenPort);
+                listenPort,
+                getEnvOrThrow("CERT_DIR"));
```
rgodfrey: Do we need to throw here? - this precludes running without TLS even though the rest of the code looks like it would work without it

admin/queue-scheduler/src/main/java/io/enmasse/queue/scheduler/QueueScheduler.java
rgodfrey: Yes, this should be  setClientAuth(ClientAuth.REQUIRED);

PR #176, opened 2017-08-30
convert router<->subserv communication to TLS

grs:


subserv/subserv.js
```diff
@@ -261,12 +262,27 @@ var connection_properties = {product:'subserv', container_id:process.env.HOSTNAM
 
 log.info("Starting subserv");
 
+var options;
+try {
+    options = tls_options.get_server_options({port:5672, properties:connection_properties});
+} catch (error) {
+    options = {port:5672, properties:connection_properties};
```
rgodfrey: I think we should probably standardise between the various components how we trigger user of TLS vs. not using TLS.  I'm not sure I'm terribly keen on "something went wrong trying to set up TLS, let's open an unsecured port" as a strategy, though until we agree on a common pattern my objection is not strenuous.

lulf:
Sorry , I closed it by mistake :)

subserv/subserv.js
grs: I agree. There seem to be a few stability issues at present in various PRs (possibly even master?), and the PRs backing up overlap a little. I'll create a merge of this, the queue-scheduler and the config-serv changes, then try to get to the bottom of the instability. After that I'll revise the approach to more explicitly choose TLS or not, further reduce the duplicated code and add some unit test coverage.

lulf: @grs the queue scheduler Pr is not ready for merge, I have to pick some things from the.configserv PR ( the external sasl authenticator)

PR #178, opened 2017-08-31
Use TLS certs to authenticate with configserv

rgodfrey:


lulf:
@rgodfrey LGTM. I'll probably move the externalsaslauthenticator to common-Lib so it can be reused for queue scheduler server in my PR.

console/lib/router_stats.js
```diff
@@ -20,24 +20,10 @@ var path = require('path');
...
-        options.ca = [fs.readFileSync(ca_path)];
-        options.key = fs.readFileSync(client_key_path);
-        options.cert = fs.readFileSync(client_crt_path);
-        options.enable_sasl_external = true;
-        options.transport = 'tls';
-        options.rejectUnauthorized = false;
-    } catch (error) {
-        console.warn('Error setting connection options ' + error);
-    }
+    var options = { host: process.env.MESSAGING_SERVICE_HOST, port: process.env.MESSAGING_SERVICE_PORT_SECURE_INTERNAL};
```
lulf: This should be: MESSAGING_SERVICE_PORT_AMQPS_NORMAL

PR #180, opened 2017-08-31
allow path for auth-service ca to be overridden

grs:
This makes it possible to run the console server outside of openshift (while still connecting to the other components), which is very useful for dev.


lulf:
LGTM!

PR #181, opened 2017-08-31
Display tooltips extracted from asciidocs for the add address wizard

ErnieAllen:
This displays tooltips on the create address wizard. The tooltips are stored in the console/www/tooltips.json file. They were extracted from the asciidocs using the parse.py utility located in the doc-utils dir.


console/www/tooltips.json
```diff
@@ -0,0 +1,63 @@
...
+        "queue": {
+            "label": "A store and forward queue",
+            "plan": {
+                "inmemory": {
+                    "label": "In memory only",
+                    "tooltip": "Creates a standalone broker cluster for queues. Messages are not persisted on stable storage."
+                },
+                "persisted": {
+                    "label": "Saved to storage",
+                    "tooltip": "Creates a standalone broker cluster for queues. Messages are persisted on stable storage."
```
lulf: Is this the file that is generated by the script after looking at the docs? I assume this file would essentially be reused for the address controller /v1/schema as well, and the names (at least tooltip) has a very UI-specific meaning. Does it make sense to rename label -> shortDescription, tooltip -> longDescription or someting like that?

ErnieAllen: Yes, this is the file that is generated by the script from parsing the docs. 
And yes, shortDescription and longDescription sound like better names. I'll make that change.

lulf:
Looks good!

PR #184, opened 2017-09-04
Address controller auth

lulf:
This enable authentication for using address controller API. The 'user database' is backed by a secret that is appended with <address space>:<random password> whenever a new address space is created.

The secret is looked up every time a user is authenticated. 

There is also the question of what to do with defaults wrt. passwords and if the user should be able to set it when creating the address space for instance.


lulf:
@grs @rgodfrey This is ready for review.

console/lib/address_ctrl.js
```diff
@@ -15,46 +15,50 @@
...
     return new Promise(function (resolve, reject) {
         var options = {
-            hostname: ctrlr.host,
-            port: ctrlr.port
+            hostname: self.host,
+            port: self.port
         };
         options.path = path;
         options.method = method || 'GET';
+        options.headers = {};
```
grs: Might be cleaner to do:
  options.headers = headers || {};

(Minor issue though).

console/lib/address_ctrl.js
```diff
@@ -15,46 +15,50 @@
...
             options.headers = headers;
         }
 
-        var req = http.request(options, function(res) {
+        if (self.auth_string) {
+            options.headers['Authorization'] = self.auth_string
+        }
+
+        if (self.ca_file) {
+            options.ca = fs.readFileSync(self.ca_file);
```
grs: Doing a sync read on every request feels a little wrong in general. I guess you could make the argument that it won't be that frequently called. If we want to re-read it every request perhaps an async read would be nicer. Again not a blocking issue, and something that can also be changed later as needed.

console/lib/address_ctrl.js
```diff
@@ -113,11 +117,11 @@ function create_address(address) {
...
         ]
     };
-    return request(addr_path, 'POST', {'content-type': 'application/json'}, JSON.stringify(data));
+    return this.request(this.addr_path, 'POST', {'content-type': 'application/json'}, JSON.stringify(data));
 }
 
-function delete_address(address) {
-    return request(addr_path + address.address, 'DELETE');
+AddressCtrl.prototype.delete_address = function(address) {
+    return this.request(this.addr_path + address.address, 'DELETE');
```
grs: Minor: Arguably the path needn't be passed in, since all current uses use the instance member variable.

console/lib/address_ctrl.js
```diff
@@ -134,10 +138,23 @@ function address_types(text) {
...
+    if (process.env.ADDRESS_SPACE_SERVICE_HOST) {
+        host = process.env.ADDRESS_SPACE_SERVICE_HOST;
+    } else if (process.env.ADDRESS_CONTROLLER_SERVICE_HOST) {
+        host = process.env.ADDRESS_CONTROLLER_SERVICE_HOST;
+        if (process.env.ADDRESS_CONTROLLER_SERVICE_PORT_HTTPS) {
+            port = process.env.ADDRESS_CONTROLLER_SERVICE_PORT_HTTPS;
+        }
+    }
+
+    var ca_file = undefined; //process.env.ADDRESS_CONTROLLER_CA;
```
grs: Is this meant to be commented out?

lulf: No! :)

console/lib/address_ctrl.js
lulf: Leaving as is since this.addr_path is not used by get_address_types(). For delete_address, the address would have to be appended to the path, but not for create_address. So there's basically 3 different uses of it.

PR #185, opened 2017-09-04
Brokered address space address controller and templates

lulf:
This is a prototype of a brokered address space. It has:

* Basic types for the brokered address space in the model: queue and topic
* Modifications to address controller to monitor address spaces with type=brokered, and instantiate the brokered instance template
* A brokered instance template, consisting of a single Artemis broker at the moment
* A brokered address controller that creates/deletes queues on the artemis broker whenever addresses are created.

I've yet to test if it works, but it is along the lines I though initially. There is also no support for topics at present. Another question is what to do with authentication. IMO, it would be nice if we could get have a SASL forwarding plugin in Artemis as well, because it would allow users to get the same experience wrt. setting up authentication.

Wdyt?


lulf:
Remove prototype now that we have reimplemented most of it.

PR #187, opened 2017-09-04
Deploy external loadbalancers and keycloak credentials into the correct namespace

scholzj:
When deploying on Kubernetes, the external load balancers and keycloak credentials are deployed without namespace. As a result they end up in the default namespace and that causes problems.


lulf:
Looks good, thanks Jakub!

PR #199, opened 2017-09-11
Add TLS support for AMQP-Kafka bridge

scholzj:
In master the AMQP-Kafka bridge doesn't work because the bridge doesn't use TLS and the right port when connecting to the router. 

This PR adds the annotation to the bridge deployment to generate the secret with TLS certificates and maps this secret as a volume. A second PR  EnMasseProject/amqp-kafka-bridge#75 adds TLS support for the bridge it self.

### DIFF
--- a/templates/include/amqp-kafka-bridge.jsonnet
+++ b/templates/include/amqp-kafka-bridge.jsonnet
@@ -2,6 +2,7 @@ local common = import "common.jsonnet";
 {
   deployment(addressSpace, image_repo)::
     {
+      local certSecretName = "amqp-kafka-bridge-internal-cert",
       "apiVersion": "extensions/v1beta1",
       "kind": "Deployment",
       "metadata": {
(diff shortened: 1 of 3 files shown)


templates/include/amqp-kafka-bridge.jsonnet
```diff
@@ -29,12 +31,66 @@ local common = import "common.jsonnet";
...
+              }
+            ],
+            "volumes": [
+              {
+                "name": certSecretName,
+                "secret": {
+                  "secretName": certSecretName
+                }
+              }
+             ]
```
ppatierno: maybe we could try to extract a function like the `clientContainer` as before to be less verbose here ?

lulf: IMHO, I'd say this is fine for now. We have an outstanding task to look at ksonnet and get a better structure of the jsonnet files in place soon. Although verbose, it's a bit easier to understand.

scholzj: I probably took a bit of a shortcut here :-). I did a quick search and it seems that the clientContainer function from `common.jsonnet` is not used for anything anymore. So it should be easy to add some support for volume mounts into it.

scholzj: I updated the PR so that it uses the function again. Have a look.

PR #206, opened 2017-09-12
Convert from json to yaml using python in order to get more control over output format

lulf:
With this change, hopefully users on different platform should now generate the same yaml format.


lulf:
Please test @rgodfrey @scholzj @ppatierno

lulf:
The expected behavior is that 'git diff' shows nothing :)

scholzj:
Yeah, it looks like @ppatierno is adding quotes (see 78419b4d0ab04c5cb7a7d1fbc4810eef7baf699f) and you seem to be removing them. I think we should agree on one way and keep it that way :-)

ppatierno:
@scholzj I made last commit adding quotes because I saw that all the other yaml files have them. Then @lulf explained me that we have same problem with @rgodfrey and it's due to Ruby version used on Mac. For this reason @lulf had the good idea to move from Ruby to Python to see if we can achieve a common format which in this case should be without quotes as Python generation. My only concern is about values that could have a space.

scholzj:
@ppatierno Ahh, ok. I noticed in one of the PRs I submitted, that on my Mac it regenerated all the yaml files with quotes. I was wondering if it is some kind of platform issue or not but didn't asked and simply backported the changes manually into the original YAML files. :-) Now it s a bit more clear to me why this change. Thanks.

rgodfrey:
@lulf - works for me.

One question I've had for a while though is why the resultant files are checked in at all - conceptually how are these different from other artefacts generated from the build such as jar files?

lulf:
@rgodfrey It is purely for convenience, but also due to the fact that we didn't have a way to store artifacts on bintray like we have now.

I guess another nice feature is that you can see what actually gets generated and compare it with what was generated before, which I find is a useful way to catch some bugs.

PR #213, opened 2017-09-13
Tls enable router metrics

lulf:


systemtests/scripts/run_test_component.sh
```diff
@@ -23,6 +23,8 @@ popd
 
 $SYSTEMTESTS/scripts/collect_logs.sh $OC_PATH $ARTIFACTS_DIR
 
+oc get pods
+
 if [ $failure -gt 0 ]
```
rgodfrey: why is this here?

lulf: Useful to see if things crash :) i'd like to keep it in

PR #215, opened 2017-09-14
Update HACKING.md

tkratky:
Added missing requirement and dependency

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -11,7 +11,7 @@ To build EnMasse, you need
     * Node.js
     * npm
     * docker
-    * maven
+    * maven >=3.1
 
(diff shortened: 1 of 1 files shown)


HACKING.md
```diff
@@ -11,6 +11,7 @@ To build EnMasse, you need
     * Node.js
     * npm
     * docker
+    * maven >=3.1
```
ppatierno: out of curiosity, what's the need for a specific minimum version ?

lulf: The maven requirement is from building a patched version of Artemis, and Artemis has a lower bound on 3.1 for maven versions.

PR #225, opened 2017-09-18
Publish test reports on bintray

lulf:


lulf:
@grs @rgodfrey @tkratky This stores test reports and snapshots of _failed_ builds (both master and PR) on bintray. If successful, nothing is stored, unless it is build from 'master', in which case a 'latest' snapshot is stored.

This ensures that the bintray quota will not fill up as long as tests pass :)

Artifacts can be found here:

https://dl.bintray.com/enmasse/snapshots/

The number is the travis build number. 'latest' contains the latest snapshot.

Issue #86, opened 2017-08-02
Create REST API for creating/updating/deleting address spaces


lulf:
As documented in https://github.com/EnMasseProject/enmasse/blob/master/documentation/address-model/resource-definitions.md#addressspace

tkratky:
@lulf : PUT method is tracked separately https://github.com/EnMasseProject/enmasse/issues/604, so I suggest to close this issue

PR #229, opened 2017-09-18
Implement REST API for listing, creating and deleting address spaces

lulf:
This fixes issue #86


lulf:
The 'update' operation is not exposed since we haven't defined how that should work yet.

PR #233, opened 2017-09-19
moved existing documentation files into a new subfolder 

jenmalloy:
and added new files for new docs content and docs build


lulf:
The stuff in the logo folder is referenced from the top level README, so the link should be updated (also on the github pages repo) . Would it make sense to move the logo folder to the root ?

documentation/design_docs/README.adoc
```diff
@@ -3,29 +3,29 @@ EnMasse
 -------
 
 Documents relating to EnMasse project. For setting up EnMasse, see
-link:getting-started/README.adoc[Getting Started]. To setup EnMasse with
+link:design_docs/getting-started/README.adoc[Getting Started]. To setup EnMasse with
```
grs: I don't think these links should have the design_docs prefix, as the doc they are in is itself in that design_docs folder. It is the toplevel README, i.e. directly under enmasse, that needs its links updated.

jenmalloy: @grs, got it, thanks, Gordon! Will revert this one and fix the top-level README file.

lulf:
Closing as I believe this has been merged in a different PR

PR #238, opened 2017-09-20
Checking in model.adoc that has parsible comments and surrounding changes

ErnieAllen:
Checking in model.adoc that has parsible comments. Changed expected.json and console code to reflect the removal of the 'plan' descriptions. Generated a new tooltips.json


grs:
Why are we removing the plan descriptions?

grs:
@ErnieAllen ok, I understand now. I think we need to include plan descriptions. I'll dig out the commit to the doc that removed them and figure out why that happened. Worst case we can add in another doc that has the plans in them for now.

grs:
@jenmalloy no, that wouldn't change the content.

@ErnieAllen, did that doc ever have anything on plans? When you generated the current tooltips, was it only from that doc or was there some other source? (I can't find any change that removed info on plans).

ErnieAllen:
@grs The original model.md doc had the plan descriptions. To generate the current tooltips I converted just that file into ascii doc and added the parsible comments. I didn't check in my version of model.adoc since all the original .md documents were going to be batch converted.

I have all the original plan descriptions. I can put them into model.adoc or wherever @jenmalloy wants them after PR: #233 is merged.

ErnieAllen:
@grs @jenmalloy Updated this pull request to only contain the parsible comments in the model.adoc.
This file contains the plan descriptions so no changes to the expected.json or console code is needed.
 
With this file you can run the parse program from the doc-utils/ directory like so:
./parse.py -i expected.json -o tooltips.json -v ../documentation

Once the documentation directory is moved/renamed you wold use the new location instead of ../documentation at the end of that command.

PR #240, opened 2017-09-22
Optional authservice

lulf:
This makes both standard and none auth services deployable independently of eachother. I'd like to get this in before I tag the 0.13.0 release.


templates/install/deploy-kubernetes.sh
```diff
@@ -24,22 +24,22 @@ fi
...
-KEYCLOAK_TEMPLATE=$SCRIPTDIR/kubernetes/addons/keycloak.yaml
-KEYCLOAK_TEMPLATE_PARAMS=""
+KEYCLOAK_TEMPLATE=$SCRIPTDIR/kubernetes/addons/standard-authservice.yaml
+NONE_TEMPLATE=$SCRIPTDIR/kubernetes/addons/none-authservice.yaml
 DEFAULT_NAMESPACE=enmasse
 GUIDE=false
 
-while getopts dgk:lm:n:t:vh opt; do
+while getopts a:dgk:lm:n:t:vh opt; do
     case $opt in
```
rgodfrey: should k not be removed from the available options here?

templates/install/deploy-openshift.sh
```diff
@@ -25,18 +25,22 @@ fi
...
 TEMPLATE_PARAMS=""
+AUTH_SERVICES="none"
 
 DEFAULT_USER=developer
 DEFAULT_NAMESPACE=myproject
 OC_ARGS=""
 GUIDE=false
 
-while getopts c:dgk:m:n:p:st:u:yvh opt; do
+while getopts a:c:dgk:m:n:p:st:u:yvh opt; do
```
rgodfrey: Again - should k not be removed

PR #245, opened 2017-09-22
improved coverage for ragent unit tests

grs:


Makefile
```diff
@@ -21,6 +21,8 @@ clean_java:
 
 clean: clean_java
 
+coverage:
```
lulf: I think you can remove this, as it is part of the $(BUILD_TARGETS) rule below

console/Makefile
```diff
@@ -7,3 +7,10 @@ package:
 	tar -czf build/console.tgz bin lib www package.json
 
 test: npm_test
+
+coverage: npm_install
```
lulf: Maybe move the 'implementation' of this rule into Makefile.common, and rename it node_coverage ?

grs: The issue there is that not all the subdirs that use Makefile.common have a package.json yet (and so can't use npm_install).

PR #254, opened 2017-09-27
Perform fewer requests and pass timeout budget to address wait loop

lulf:


systemtests/src/main/java/enmasse/systemtest/TestUtils.java
```diff
@@ -228,28 +228,51 @@ public static boolean isAddressReady(JsonObject address) {
      *
      * @param apiClient    instance of AddressApiClient
      * @param instanceName name of instance
+     * @param budget       the timeout budget for this operation
      * @param destinations variable count of destinations
-     * @throws Exception IlegalStateException if destionation is not ready within timeout
+     * @throws Exception IlegalStateException if destinations is not ready within timeout
```
rgodfrey: <pedantic>destinations *are* not ready</pedantic>

systemtests/src/main/java/enmasse/systemtest/TestUtils.java
```diff
@@ -228,28 +228,51 @@ public static boolean isAddressReady(JsonObject address) {
...
+                } else {
+                    numReady++;
                 }
-                Logging.log.info("Waiting until destination: " + destination.getAddress() + " will be in ready state");
             }
+            isReady = (numReady == destinations.length);
             if (!isReady) {
-                throw new IllegalStateException("Address " + destination.getAddress() + " is not in Ready state within timeout.");
+                Thread.sleep(5000);
+            }
```
rgodfrey: If we get to this point and timeLeft < 5 seconds, won't the method always report failure (since we don't actually test readiness again). Potentially we could sleep min(5000, budget.timeLeft() - small number) or something?

PR #260, opened 2017-09-27
initial stab at some console help and tooltips generated by build

grs:


console/Makefile
```diff
@@ -1,9 +1,14 @@
...
-	npm install
+build: npm_install www/tooltips.json www/help.html
 
-package:
+www/tooltips.json: ../doc-utils/parse.py ../doc-utils/expected.json ../documentation/modules/*
+	../doc-utils/parse.py -i ../doc-utils/expected.json -o www/tooltips.json -v ../documentation/modules/
+
+www/help.html:  ../documentation/master.adoc ../documentation/modules/*.adoc ../documentation/common/*.adoc
+	asciidoctor -d book -B ../documentation/ -o ../console/www/help.html ../documentation/master.adoc
+
```
rgodfrey: presumably we should add the requirement for asciidoctor to HACKING.md?

lulf: And mark it as 'optional' in the sense that you only need it to build the docs, which I think some devs won't be interested in building.

lulf:
LGTM after skimming through it

PR #288, opened 2017-10-02
simple prototype of broker controller and console server

grs:


console/lib/broker_controller.js
```diff
@@ -0,0 +1,125 @@
...
+    this.addresses = addresses.reduce(function (map, a) { map[a.address] = a; return map; }, {});
+    this.check_broker_addresses();
+};
+
+BrokerController.prototype.retrieve_stats = function () {
+    var self = this;
+    if (this.broker !== undefined) {
+        this.broker.getAllQueuesAndTopics().then(function (results) {
+            for (var name in results) {
+                self.console_server.address_list.update_stats(name, results[name]);
```
lulf: This dependency on the console 'object' seems a bit misplaced... Is there a way to have some sort of 'listener' interface that the console could implement and this controller could call? Seems like it would be easier to unit test this way.

console/bin/console_server.js
```diff
@@ -15,136 +15,28 @@
...
+var address_source = new AddressSource();
+address_source.on('addresses_defined', address_list.addresses_defined.bind(address_list));
+
+var console_server = new ConsoleServer(address_list, address_ctrl.create());
+console_server.listen();
+
+if (process.env.ADDRESS_SPACE_TYPE === 'broker') {
+    var BrokerController = require('../lib/broker_controller.js');
+    var bc = new BrokerController(console_server);
+    //TODO: have broker connect to controller? or have controller connect to broker?
```
lulf: IMO have controller connect to broker seems to be the simplest since one does not have to add the outgoing connector + listener in the broker controller.

console/bin/console_server.js
```diff
@@ -15,136 +15,28 @@
...
+    //bc.listen();
+    bc.connect({host:process.env.BROKER_SERVICE_HOST, port:process.env.BROKER_SERVICE_PORT});
+    address_source.on('addresses_defined', bc.addresses_defined.bind(bc));
+} else {
+    //assume standard address space for now
+    var StandardStats = require('../lib/standard_stats.js');
+    var stats = new StandardStats();
+    stats.init(console_server);
 }
```
lulf: I was envisioning to have these plugins more 'configurable'/injected from the outside, but that might be overcomplicating it.

An approach I think might be better as a way to isolate standard from brokered would be to create 2 distinct 'main.js' entrypoints (in 2 separate modules that depends on this 'common' module). One for the standard, and one for the brokered agent.

grs: I'm fine with two separate executables. I'm less convinced about having separate modules.

console/bin/console_server.js
grs: Yes, I think I agree. The standard address space will be the other way round, but there is really no reason they should be the same.

console/lib/broker_controller.js
grs: I agree a simpler interface is needed here.

console/bin/console_server.js
lulf: Lets go with a single executable as discussed, and consider splitting up in the future.

lulf:
Rebased this against https://github.com/EnMasseProject/enmasse/pull/329 and readded 'standard' plans.

PR #296, opened 2017-10-03
Add types for brokered address space

lulf:
This fixes issue #273

### DIFF
--- a/address-model-lib/src/main/java/io/enmasse/address/model/types/brokered/BrokeredType.java
+++ b/address-model-lib/src/main/java/io/enmasse/address/model/types/brokered/BrokeredType.java
@@ -1,0 +1,70 @@
+/*
+ * Copyright 2017 Red Hat Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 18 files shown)


address-model-lib/src/main/java/io/enmasse/address/model/types/brokered/BrokeredType.java
```diff
@@ -0,0 +1,70 @@
...
+
+    QUEUE {
+        @Override
+        public String getName() {
+            return "queue";
+        }
+
+        @Override
+        public String getDescription() {
+            return "A queue that supports selectors, message grouping and transactions.";
```
grs: Minor point, but these description could be taken to mean that the brokered topic does not support selectors or transactions.

Ideally the descriptive text would come from one source wherever it is displayed, and would be easy to edit.

That does not need to happen as part of this change or even this sprint, just noting it as it occurred to me when looking through the new code.

lulf: Agreed, perhaps an additional task for this sprint would be to fetch the type descriptions from the json generated from the documentation.

PR #302, opened 2017-10-04
UPD added new version of mocha for ragent and console

tkratky:
mocha-junit-reporter version dependency issue was fixed, so we are able to use latest version of mocha

### DIFF
--- a/console/package.json
+++ b/console/package.json
@@ -19,7 +19,7 @@
     },
     "devDependencies": {
         "eslint":"",
-        "mocha":"^3.0.0",
+        "mocha":"^4 || ^3.0.0",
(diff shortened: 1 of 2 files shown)


tkratky:
I tried it locally with the same result - build is stuck on    
    ...
    ✓ uses CA_PATH as path for ca
    ✓ uses CERT_PATH as path for cert
    ✓ uses KEY_PATH as path for key
    ✓ retrieves client options
    ✓ retrieves server options


  58 passing (42s)

issue opened for this: https://github.com/EnMasseProject/enmasse/issues/305

tkratky:
closed without merge because of #305

PR #327, opened 2017-10-09
version of mocha for ragent and console

tkratky:


tkratky:
thanks @kornys for your help

systemtests/src/test/java/enmasse/systemtest/TestBase.java
```diff
@@ -74,7 +74,7 @@ public void setup() throws Exception {
     protected void createAddressSpace(String addressSpace, String authService) throws Exception {
         if (!TestUtils.existAddressSpace(addressApiClient, addressSpace)) {
             Logging.log.info("Address space " + addressSpace + "doesn't exist and will be created.");
-            addressApiClient.createAddressSpace(addressSpace, authService);
+            addressApiClient.createAddressSpace(addressSpace, authService, environment.isBrokered() ? "brokered" : "standard");
```
lulf: I don't think it makes sense to resolve this from the environment, since we won't expect the same tests to work in different address spaces. Instead, I'd say we should provide the address space type as an argument to this method (and add another one that uses 'standard' by default), so that we can write tests that explicitly creates brokered address spaces instead.

The standard address space is supposed to be the default anyway.

tkratky: commits wrt system-tests removed from this branch

PR #329, opened 2017-10-09
Resolve address default plan and types when needed

### LINKED ISSUE #314: Address Controller REST API is tied to schema for standard address space
Meaning that the addresses and plans used in any other address space type (e.g. brokered) must be the same as something that exists in the standard address space, which is undesirable.

lulf:
* This makes it possible to use different plan names for addresses in different address spaces

This fixes #314

### DIFF
--- a/address-controller/src/main/java/io/enmasse/controller/standard/AddressController.java
+++ b/address-controller/src/main/java/io/enmasse/controller/standard/AddressController.java
@@ -1,5 +1,6 @@
 package io.enmasse.controller.standard;
 
+import io.enmasse.address.model.AddressType;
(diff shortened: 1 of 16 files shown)


address-controller/src/main/java/io/enmasse/controller/standard/AddressController.java
```diff
@@ -1,5 +1,6 @@
 package io.enmasse.controller.standard;
 
+import io.enmasse.address.model.AddressResolver;
 import io.enmasse.amqp.SyncRequestClient;
```
rgodfrey: is this import used?

PR #347, opened 2017-10-11
Artemis sasl delegation

rgodfrey:
Ports are probably still wrong - need to test connecting from outside openshift


artemis/config_templates/brokered/broker.xml
```diff
@@ -96,8 +96,10 @@ under the License.
       <global-max-size>104857600</global-max-size>
 
       <acceptors>
-         <acceptor name="amqp">tcp://0.0.0.0:5672?protocols=AMQP</acceptor>
-         <acceptor name="amqps-internal">tcp://0.0.0.0:5673?protocols=AMQP;sslEnabled=true;keyStorePath=${KEYSTORE_PATH};keyStorePassword=enmasse;trustStorePath=${TRUSTSTORE_PATH};trustStorePassword=enmasse;verifyHost=false;needClientAuth=true</acceptor>
+         <acceptor name="amqp">tcp://0.0.0.0:5672?protocols=AMQP;saslMechanisms=PLAIN</acceptor>
+         <acceptor name="amqps">tcp://0.0.0.0:5671?protocols=AMQP;sslEnabled=true;saslMechanisms=PLAIN;keyStorePath=${KEYSTORE_PATH};keyStorePassword=enmasse</acceptor>
+         <!-- SASL Mechanism should be EXTERNAL not anonymous... but Artemis doesn't support EXTERNAL -->
+         <acceptor name="amqps-internal">tcp://0.0.0.0:55671?protocols=AMQP;saslMechanisms=ANONYMOUS;sslEnabled=true;keyStorePath=${KEYSTORE_PATH};keyStorePassword=enmasse;trustStorePath=${TRUSTSTORE_PATH};trustStorePassword=enmasse;verifyHost=false;needClientAuth=true</acceptor>
```
lulf: Should be amqps-normal

rgodfrey: I assume the names don't actually matter here, but I shall change it presently

rgodfrey:
@lulf Agree on tests - which I'll work on today; we can merge if not having the proper auth is holding up any of the other testing that we'd want to do against the brokered address space otherwise we can wait for the tests to be done :)

PR #354, opened 2017-10-13
Using pipe for showing "none" and "standard" as exclusive options

ppatierno:


ppatierno:
Ah I didn't know about that. And how it decides which auth services is used if you deploy both ?

lulf:
@ppatierno It uses standard by default, but you can select when you create an address space

ppatierno:
Ah ok. It was not clear to me. So I'm going to close this PR without merging it. Thanks for the clarification ;)

PR #358, opened 2017-10-16
retrieve address information directly from kubernetes

grs:
removes configserv from brokered address space


lulf:
@grs 

In order for this to work reliably in the event of disconnections to kubernetes, there needs to be a periodic 'list and watch' to ensure events are not missed. Basically watches are only useful to populate a local cache and should not solely relied upon. This list + watch logic is common in kubernetes controllers written in Go. I also implemented that in the address controller, which seemed to solve some instabilities wrt. missing updates.

Also, It should be possible to provide namespace and token from environment variables for testing changes locally.

console/lib/kubernetes.js
```diff
@@ -0,0 +1,157 @@
...
+    var request = https.get(list_options(this.resource, this.options), function(response) {
+	console.log('STATUS: ' + response.statusCode);
+	response.setEncoding('utf8');
+        var data = '';
+	response.on('data', function (chunk) { data += chunk; });
+	response.on('end', function () {
+            try {
+                var list = JSON.parse(data);
+	        self.objects = list.items;
+                self.emit('updated', self.objects);
```
lulf: indentation

console/lib/kubernetes.js
```diff
@@ -0,0 +1,157 @@
...
+}
+
+util.inherits(Watcher, events.EventEmitter);
+
+Watcher.prototype.list = function () {
+    var self = this;
+    var request = https.get(list_options(this.resource, this.options), function(response) {
+	console.log('STATUS: ' + response.statusCode);
+	response.setEncoding('utf8');
+        var data = '';
```
lulf: indentation

lulf:
@grs I've created #369 for the second change. Review when you are back :)

PR #364, opened 2017-10-17
Added folders and a master.adoc file for each guide

jenmalloy:
@nengard, should the "modules/" part be deleted?


jenmalloy:
@grs @rgodfrey Any objections to this structure?

documentation/address_space_admin/master.adoc
```diff
@@ -0,0 +1,16 @@
+include::common/attributes.adoc[]
+
+= {Title}
+
+include::modules/.adoc[leveloffset=+1]
+
```
rgodfrey: should these just be commented out (we have no modules sub directory, and obviously we wouldn't expect a doc called just ".adoc")

documentation/service_admin/master.adoc
```diff
@@ -0,0 +1,16 @@
+include::common/attributes.adoc[]
+
+= {Title}
+
+include::modules/.adoc[leveloffset=+1]
+
```
rgodfrey: should these just be commented out/removed (we have no modules sub directory, and obviously we wouldn't expect a doc called just ".adoc")

documentation/tenant/master.adoc
```diff
@@ -0,0 +1,16 @@
+include::common/attributes.adoc[]
+
+= {Title}
+
+include::modules/.adoc[leveloffset=+1]
+
```
rgodfrey: should these just be commented out/removed (we have no modules sub directory, and obviously we wouldn't expect a doc called just ".adoc")

documentation/messaging_app_developer/master.adoc
```diff
@@ -0,0 +1,16 @@
+include::common/attributes.adoc[]
+
+= {Title}
+
+include::address-model.adoc[leveloffset=+1]
+
+include::address-space.adoc[leveloffset=+1]
+
```
rgodfrey: Do the address-model / address-space docs really belong here?  They are common concept type things that are probably most relevant to the tenant, but really likely to be info we present across a number of books.  A messaging app developer might choose to ignore the conceptual model part as they *only* care about connecting

documentation/address_space_admin/master.adoc
jenmalloy: @rgodfrey, yes, those were intended solely as stubs. Will comment it out

documentation/messaging_app_developer/master.adoc
jenmalloy: @rgodfrey, I'll remove them here and add them to the tenant master.adoc file instead. Also, since they are likely to be used in other guides as well in the future, I'll move those two files to the common folder to facilitate easier reuse.

PR #375, opened 2017-10-20
Initial version of Swagger/OpenAPI spec

luksa:


luksa:
I've added the AddressSpaceType type and referenced it in the AddressSpace definition (previously it referenced the AddressType incorrectly).

But, I don't think swagger supports having two different AddressType enums (e.g. BrokeredAddressType and StandardAddressType) and have them references in the Address definition (and somehow specify which one should be used based on the value of the AddressSpace.type the Address is being created in). So, at least for now, AddressType should include a union of all possible options. The same will also apply to plans.

lulf:
Issue for autogenerating: https://github.com/EnMasseProject/enmasse/issues/419

address-controller/src/main/resources/swagger.json
```diff
@@ -0,0 +1,640 @@
...
+                        "required": true,
+                        "schema": {
+                            "$ref": "#/definitions/AddressSpace"
+                        }
+                    }
+                ],
+                "responses": {
+                    "405": {
+                        "description": "Invalid input"
+                    }
```
mtaylor: You probably want to return a 201 on successful creation, or a 202 if it's async.  Probably also want to return the created resource URI.

address-controller/src/main/resources/swagger.json
```diff
@@ -0,0 +1,640 @@
...
+                        }
+                    },
+                    "404": {
+                        "description": "AddressSpace not found"
+                    }
+                }
+            }
+        },
+        "/addresses/{addressSpace}": {
+            "get": {
```
mtaylor: Wouldn't this make more sense as: /addressspaces/{name}/addresses

address-controller/src/main/resources/swagger.json
```diff
@@ -0,0 +1,640 @@
...
+                        "type": "string"
+                    }
+                ],
+                "responses": {
+                    "200": {
+                        "description": "successful operation",
+                        "schema": {
+                            "$ref": "#/definitions/AddressList"
+                        }
+                    }
```
mtaylor: 404 if the address space doesn't exist?

address-controller/src/main/resources/swagger.json
```diff
@@ -0,0 +1,640 @@
...
+                    "application/json"
+                ],
+                "parameters": [
+                    {
+                        "name": "addressSpace",
+                        "in": "path",
+                        "description": "Name of AddressSpace",
+                        "required": true,
+                        "type": "string"
+                    },
```
mtaylor: does this mean you post to /addresses/addresspace/{name}?  If so I would say this is the same thing as the PUT operation below.

address-controller/src/main/resources/swagger.json
```diff
@@ -0,0 +1,640 @@
...
+                        "required": true,
+                        "schema": {
+                            "$ref": "#/definitions/AddressList"
+                        }
+                    }
+                ],
+                "responses": {
+                    "405": {
+                        "description": "Invalid input"
+                    }
```
mtaylor: Successful response codes?

address-controller/src/main/resources/swagger.json
```diff
@@ -0,0 +1,640 @@
...
```
mtaylor: Same comment as before, successful response codes.

address-controller/src/main/resources/swagger.json
```diff
@@ -0,0 +1,640 @@
...
+                        }
+                    },
+                    "404": {
+                        "description": "AddressSpace not found"
+                    }
+                }
+            }
+        },
+        "/addressses/{addressSpace}/{address}": {
+            "get": {
```
mtaylor: I'm a little confused by this hierarchy.  I would have thought it would be something like.

AddressSpaces/{addressSpace}/addressess/address

address-controller/src/main/resources/swagger.json
```diff
@@ -0,0 +1,640 @@
...
+            }
+        },
+        "AddressType": {
+            "type": "string",
+            "description": "Type of address (queue, topic, ...)",
+            "enum": [
+                "QUEUE",
+                "TOPIC",
+                "ANYCAST",
+                "MULTICAST"
```
mtaylor: I guess this is already an EnMasseism.  But can you remind me what the difference is between a Topic and Mulitcast address.  In Artemis a Topic is based on a multicast address.

address-controller/src/main/resources/swagger.json
```diff
@@ -0,0 +1,640 @@
...
+                        }
+                    },
+                    "404": {
+                        "description": "Address or AddressSpace not found"
+                    }
+                }
+            }
+        }
+    },
+    "securityDefinitions": {
```
mtaylor: Is there a purpose for the meta-data.  I presume that the swagger library can ingest this data and spit out a client?  Or is it mostly for documentation purposes?

address-controller/src/main/resources/swagger.json
lulf: Yes, we'll probably change that in the future.

address-controller/src/main/resources/swagger.json
lulf: Multicast is what we've named the 'direct' multicast supported by the qpid dispatch router.

address-controller/src/main/resources/swagger.json
luksa: It currently returns 200. I've added 200 to the spec now. Thanks

address-controller/src/main/resources/swagger.json
luksa: Thanks. Added.

address-controller/src/main/resources/swagger.json
luksa: Added it now. Thanks.

address-controller/src/main/resources/swagger.json
luksa: Done.

address-controller/src/main/resources/swagger.json
luksa: Yup, paths need to be cleaned up.

address-controller/src/main/resources/swagger.json
luksa: Forgot to remove this part. Will need to be re-added later with correct information. I believe it's used both for documentation and generating clients.

PR #381, opened 2017-10-20
Add discoverable URIs from /v1 root

### LINKED ISSUE #359: Addresscontroller /v1 endpoint returns 404
It should return a list of links to the addresses and addressspaces endpoints. See kubernetes/v1 for an example.

lulf:
This fixes #359


luksa:
I'd also add a handler for /, which would include links to /v1, /osbapi, /healthz, etc.

PR #384, opened 2017-10-23
FIX TestUtils - added "wait until all addresses are deleted"

tkratky:


lulf:
@tkratky LGTM, but have you actually seen this happen that you got 200 OK but addresses were not deleted? Because I think if the delete was successful, the address should not be listed, and would otherwise be a bug.

tkratky:
@lulf I wasn't able to reproduce exactly the same issue as I saw before in standard.QueueTest#testRestApiForQueue which was Expected: is <[]>
     but: was <[queue2]>

but I was able to reproduce restart of address-controller: #385

tkratky:
shouldn't be merged before: https://github.com/EnMasseProject/enmasse/pull/425

PR #386, opened 2017-10-23
Add a java router to be used in testing

lulf:
* This removes the need to run the qpid dispatch router to test
mqtt-gateway and mqtt-lwt


lulf:
@rgodfrey Judging from the amount of code that had to be added, I'd really like for the test to be rewritten somehow, but I think this is an improvement at least...

Issue #280, opened 2017-09-29
brokered address space: Update address status based on status of address in Broker


rgodfrey:

grs:
implemented in https://github.com/EnMasseProject/enmasse/pull/391

grs:
Further fix in https://github.com/EnMasseProject/enmasse/pull/426

PR #391, opened 2017-10-24
update status in config maps

grs:


grs:
Fixes https://github.com/EnMasseProject/enmasse/issues/280

templates/include/brokered-space-infra.jsonnet
```diff
@@ -176,6 +176,7 @@ local hawkularBrokerConfig = import "hawkular-broker-config.jsonnet";
           }
         },
         "spec": {
+          "serviceAccount": "enmasse-service-account",
```
lulf: This service account is not available in the address space namespace in a multitenant deployment so this must be created by the address controller. I’ll have a look.

PR #399, opened 2017-10-25
Allow service account within a tenant to edit its namespace resources

lulf:


luksa:
One thing you should be aware of: users usually aren't allowed to modify their namespaces (and shouldn't be), because some features, which should be outside of the control of a user, are configured through annotations on namespaces.

lulf:
@luksa

The address space namespaces can be regarded as 'internal' to the messaging service though, do you think we should deal with this differently? The reason we need 'edit' permission is so that the 'Address' status can be updated by the controller running within the address space (this is a design change where we want to make the address space more autonomous).

lulf:
Merging for now to unblock @grs

PR #400, opened 2017-10-25
FIX AddressApiClient - new handling of response from address-controller

tkratky:


tkratky:
I agree, I'll do that

PR #401, opened 2017-10-25
Add support for maintaining endpoints and address space readiness for brokered address space

lulf:


lulf:
As part of this change, the hardcoded (1) route -> service and (2) service -> port mapping is removed, so that (1) is defined for each address space type and (2) is defined for for each service.

address-controller/src/main/java/io/enmasse/controller/ControllerHelper.java
```diff
@@ -165,14 +157,12 @@ private StandardResources createResourceList(AddressSpace addressSpace) {
             }
 
             // Step 3: Create missing secrets if not specified
-            Set<String> secretsToGenerate = new HashSet<>();
             for (String service : availableServices) {
                 String secretName = getSecretName(service);
 
-                    // TODO: Remove console clause when it supports https
+                // TODO: Remove console clause when it supports https
```
rgodfrey: Do we have an issue covering this TODO?  If we do we should remove the TODO.. if we don't then we should create an issue... and then remove the TODO :-)

templates/include/admin.jsonnet
```diff
@@ -28,7 +29,7 @@ local common = import "common.jsonnet";
     self.service("ragent", addressSpace, [{"name": "amqp", "port": 5671, "targetPort": 55671}]),
     self.service("configuration", addressSpace, [{"name": "amqps", "port": 5671}]),
     self.service("queue-scheduler", addressSpace, [{"name": "amqp", "port": 5672, "targetPort": 55667}]),
-    self.service("console", addressSpace, [{"name": "amqp-ws", "port": 5672, "targetPort": 56720}, {"name": "http", "port": 8080}])
+    self.service("console", addressSpace, [{"name": "amqp-ws", "port": 5672, "targetPort": 56720}, {"name": "http", "port": 8080}], "http")
```
grs: I don't believe we need the amqp-ws port mapping here, it could just be http.

address-controller/src/main/java/io/enmasse/controller/ControllerHelper.java
lulf: https://github.com/EnMasseProject/enmasse/issues/402 :) Removing the TODO

Issue #398, opened 2017-10-25
remove express dependency


grs:
express has itself quite a lot of transitive dependencies and our usage doesn't really justify all that extra baggage, so it would be nice to remove that and slim down the runtime dependencies for the agent

PR #410, opened 2017-10-25
remove dependency on express and basic_auth

grs:
Fixes issue #398

### DIFF
--- a/agent/lib/console_server.js
+++ b/agent/lib/console_server.js
@@ -16,10 +16,10 @@
 'use strict';
 
 var log = require("./log.js").logger();
-var express = require('express');
-var basic_auth = require('basic-auth');
+var fs = require('fs');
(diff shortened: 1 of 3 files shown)


agent/lib/console_server.js
```diff
@@ -93,26 +93,78 @@ ConsoleServer.prototype.ws_bind = function (server) {
...
-    var auth = function (req, res, next) {
-        var authrequired = function() {
-            res.set('WWW-Authenticate', 'Basic realm=Authorization Required');
-            return res.sendStatus(401);
-        };
-        var user = basic_auth(req);
-
-        auth_service.authenticate(user, auth_service.default_options(env)).then( next ).catch( authrequired );
-    };
+function basic_auth(request) {
```
lulf: This function is a good candidate for a unit test

PR #428, opened 2017-10-27
Add kubernetes documentation and master.adoc for getting started

lulf:
The purpose of this change is to add an install guide for kubernetes and to be able to expose the getting started guides on the website.


lulf:
@jenmalloy @grs Could it be useful to add some sort of generic category for 'getting started' guides? I can keep referring to the design_docs/getting-started folder on the website, but it feels like it should be put somewhere.

I think many newcomers look for this kind of 'self-contained' quick-start guides

documentation/service_admin/installing-kubernetes.adoc
```diff
@@ -0,0 +1,90 @@
...
+Since minikube runs in a virtual machine, you need to enable the ingress controller:
+
+[source,options="nowrap"]
+....
+minikube addons enable ingress
+....
+
+.Procedure 
+
+. Download one of the releases from https://github.com/EnMasseProject/enmasse/releases and unpack it.
```
grs: Minor point, but if we are parameterising the product name, we should also parameterise the link to releases. (Not a blocking issue on this PR though).

documentation/design_docs/getting-started/openshift.adoc
```diff
@@ -43,135 +43,6 @@ up EnMasse will take a while, usually depending on how fast it is able
 to download the docker images for the various components. In the
 meantime, you can start to create your address configuration.
 
-[[deploying-enmasse-manually]]
```
grs: Personally I think the description of how to install manually is valuable. I agree it may not be relevant to a getting started guide, but it might be worth keeping it for more indepth readers. Though you can always read the script, its nice to see not only the manual steps but also their purpose, as a way of understanding how it all works.

lulf: @grs I agree, but it's a bit hard to maintain those steps and making them up to date. Btw, you can actually generate the list of manual steps by passing -g to the deployment script. Maybe mentioning that should be sufficient? Another approach would be if we as part of build ran the script with -g, put the output into a manual_steps.adoc, and referred to that.

Wdyt?

documentation/design_docs/getting-started/aws.adoc
```diff
@@ -103,14 +103,11 @@ EnMasse.
 You can find the latest version of EnMasse
 https://github.com/EnMasseProject/enmasse/releases/latest[here]. To
 deploy EnMasse, it is recommended to use the deploy script together with
-a template of the latest version. At the time of writing, the latest
-version is 0.10.0, which can be deployed as follows:
+a template of the latest version.
 
-....
-curl -L https://github.com/EnMasseProject/enmasse/releases/download/0.10.0/enmasse-deploy.sh | sh - -c "https://openshift.<yourdomain>:8443" -p enmasse -u <myuser>
-....
```
grs: (It might be useful to show the script options for a typical openshift on aws deployment(?))

PR #439, opened 2017-10-31
improved test coverage for agent

grs:


agent/lib/console_server.js
```diff
@@ -51,17 +52,12 @@ function ConsoleServer (address_ctrl) {
     this.amqp_container.on('sender_open', function (context) {
         self.subscribe(context.connection.remote.open.container_id, context.sender);
     });
-    this.amqp_container.on('sender_close', function (context) {
+    function unsubscribe (context) {
         self.unsubscribe(context.connection.remote.open.container_id);
```
lulf: The condition if (context.connection.remote.open)  is missing from the disconnected case, is it ok?

agent/lib/internal_address_source.js
```diff
@@ -61,14 +64,15 @@ AddressSource.prototype.update_status = function (record, ready) {
...
-    kubernetes.update('configmaps/' + record.name, update).then(function (result) {
+    return kubernetes.update('configmaps/' + record.name, update, this.config).then(function (result) {
         if (result === 200) {
             record.ready = ready;
             log.info('updated status for %j: %s', record, result);
         } else {
             log.error('failed to update status for %j: %s', record, result);
         }
     }).catch(function (error) {
+        console.log(error);
```
lulf: Is there any need for the duplicate logging?

PR #440, opened 2017-11-01
Add documentation of the brokered address space

lulf:


documentation/common/address-space.adoc
```diff
@@ -127,3 +127,52 @@ The multicast address type is a scalable direct address for sending messages to
...
+semantics are not needed, have a look at the `standard` address space `queue` type.
+
+[[brokered-queue-plans]]
+===== Queue Plans
+
+* standard
+
+[[standard-queue-plan]]
+====== Standard
+The standard queue plan deploys a queue in a broked instance shared with other addresses in the
```
grs: Only issue with this is that it sounds like other plans might not share the broker.

documentation/common/address-space.adoc
```diff
@@ -127,3 +127,52 @@ The multicast address type is a scalable direct address for sending messages to
 * standard
 
 // address.multicast.plan.inmemory.shortDescription:Configures router network with multicast address
+
+[[brokered_address_space]]
+=== Brokered Address Space
```
grs: One question is whether it would be worth keeping the address spaces in distinct module files(?).

documentation/common/address-space.adoc
```diff
@@ -127,3 +127,52 @@ The multicast address type is a scalable direct address for sending messages to
...
+
+The brokered address space is designed to support broker-specific features, at the cost of limited
+scale in terms of the number of connections and addresses. This address space supports JMS
+transactions, message groups etc. The brokered address space is backed by a broker, and stores
+persisted messages on disk if required.
```
grs: The last sentence seems a little redundant to me. Its not wrong though in any way. I think whats there is good to start off with and we can then try and iterate on it as time allows.

documentation/common/address-space.adoc
```diff
@@ -127,3 +127,52 @@ The multicast address type is a scalable direct address for sending messages to
 * standard
 
 // address.multicast.plan.inmemory.shortDescription:Configures router network with multicast address
+
+[[brokered_address_space]]
+=== Brokered Address Space
+
+The brokered address space is designed to support broker-specific features, at the cost of limited
+scale in terms of the number of connections and addresses. This address space supports JMS
+transactions, message groups etc. The brokered address space is backed by a broker, and stores
```
jenmalloy: Change "etc." to "and so on" and add a comma after "message groups"

PR #443, opened 2017-11-01
Basic tests external clients

kornys:


systemtests/src/test/java/io/enmasse/systemtest/clients/BasicClientTest.java
```diff
@@ -0,0 +1,200 @@
...
+        ArgumentMap arguments = new ArgumentMap();
+        arguments.put(Argument.USERNAME, "test");
+        arguments.put(Argument.PASSWORD, "test");
+        arguments.put(Argument.BROKER, getRouteEndpoint(addressSpace).toString());
+        arguments.put(Argument.ADDRESS, dest.getAddress());
+        arguments.put(Argument.LOG_MESSAGES, "json");
+        arguments.put(Argument.COUNT, "10");
+        arguments.put(Argument.CONN_SSL, "true");
+
+        AbstractClient sender  = new RheaClientSender();
```
tkratky: Clients(with arguments) should be prepared in setUp method before each test and should be added into the List of clients. Only code of test should iterate through this list. TearDown method should clear this list. It was the original reason why we did AbstractClient, executor, etc...

systemtests/src/test/java/io/enmasse/systemtest/clients/BasicClientTest.java
```diff
@@ -0,0 +1,200 @@
...
+
+        Future<Boolean> recResult = receiver.runAsync();
+        Future<Boolean> rec2Result = receiver2.runAsync();
+
+        arguments.put(Argument.COUNT, "10");
+
+        sender.setArguments(arguments);
+
+        assertTrue(sender.run());
+        assertTrue(recResult.get());
```
tkratky: Some timeout should be there for reason that messages will never received

systemtests/src/test/java/io/enmasse/systemtest/clients/BasicClientTest.java
```diff
@@ -0,0 +1,200 @@
...
+        AbstractClient sender  = new RheaClientSender();
+        AbstractClient receiver = new RheaClientReceiver();
+
+        sender.setArguments(arguments);
+        receiver.setArguments(arguments);
+
+        Future<Boolean> recResult = receiver.runAsync();
+
+        assertTrue(sender.run());
+        assertTrue(recResult.get());
```
tkratky: add timeout as above

systemtests/src/test/java/io/enmasse/systemtest/clients/BasicClientTest.java
kornys: there is default timeout 60sec in sync and 120 in async

systemtests/src/test/java/io/enmasse/systemtest/clients/BasicClientTest.java
kornys: same as above

systemtests/src/test/java/io/enmasse/systemtest/clients/rhea/BasicClientTest.java
```diff
@@ -0,0 +1,33 @@
+package io.enmasse.systemtest.clients.rhea;
+
+import io.enmasse.systemtest.executor.client.rhea.RheaClientReceiver;
+import io.enmasse.systemtest.executor.client.rhea.RheaClientSender;
+import org.junit.Test;
+
+public class BasicClientTest extends io.enmasse.systemtest.clients.BasicClientTest {
```
tkratky: What about move those tests into the BasicClientTest in clients package?
Because we don't want to duplicate Test Classes for each client. We want do something like following
```
@Test
public void basicMessageTest() throws Exception {
       doBasicMessageTest(new RheaClientSender(), new RheaClientReceiver());
       doBasicMessageTest(new PythonClientSender(), new PythonClientReceiver());
       doBasicMessageTest(new JavaClientSender(), new JavaClientReceiver());
       ...etc
}
```

systemtests/src/test/java/io/enmasse/systemtest/clients/BasicClientTest.java
```diff
@@ -124,27 +103,18 @@ public void topicSubscribeTest() throws Exception {
         assertEquals(10, receiver.getMessages().size());
     }
 
-    @Test
-    public void messageBrowseTest() throws Exception {
+    protected void doMessageBrowseTest(AbstractClient sender, AbstractClient receiver_browse, AbstractClient receiver_receive)
+            throws Exception {
         AddressSpace addressSpace = new AddressSpace("brokered-message-browse",
```
tkratky: Maybe should be better to do only one set up of address-space and addresses for all clients? (I don't know)
advantages/disadvantages:
We don't want to have garbage in queues/topic for run of next client, but it's time consuming to create everything again. Maybe we should use the same address space, but addresses(queue/topic) should be recreated?

kornys: For now we will keep this implementation.

systemtests/src/test/java/io/enmasse/systemtest/clients/rhea/BasicClientTest.java
kornys: For now we will keep this implementation.

PR #470, opened 2017-11-06
Use RBAC for authentication and authorization

lulf:
This modifies @luksa's initial PR #454 to use the k8s rest api directly as https://github.com/fabric8io/kubernetes-model/pull/240 has not yet been merged to master.


address-controller/src/main/java/io/enmasse/controller/common/KubernetesHelper.java
```diff
@@ -344,6 +352,82 @@ public boolean isDestinationClusterReady(String clusterId) {
...
+            }
+            return new TokenReview(userName, authenticated);
+        } else {
+            return new TokenReview(null, false);
+        }
+    }
+
+    @Override
+    public SubjectAccessReview performSubjectAccessReview(String user, String path, String verb) {
+        if (client.isAdaptable(OkHttpClient.class)) {
```
luksa: How come you're not using OpenShiftClient.subjectAccessReviews() instead? Is it because you want to support Kubernetes instead of just OpenShift?

lulf: Yes, I was thinking I had already crossed a 'limit' when using the http client directly, so might as well use the same API for these things at least.

lulf:
@luksa An idea I just had:

How about instead of using nonResourceUrls for the role, using the 'configmap' resource? we don't use configmaps for anything other than storing address space and address resources, so having a Role for CRUD operations on a configmap would give the same effect as using CRDs + a Role for those.

That would remove the need for creating the 'enmasse-admin' cluster role + any 'enmasse-address-space-X' cluster roles for each address space and instead create a namespaced Role.

Wdyt?

address-controller/src/main/java/io/enmasse/controller/common/KubernetesHelper.java
```diff
@@ -344,6 +352,106 @@ public boolean isDestinationClusterReady(String clusterId) {
...
+            return new TokenReview(null, false);
+        }
+    }
+
+    @Override
+    public SubjectAccessReview performSubjectAccessReview(String user, String verb) {
+        if (client.isAdaptable(OkHttpClient.class)) {
+            JsonObject body = new JsonObject();
+
+            body.put("kind", "SubjectAccessReview");
```
luksa: Because you're now checking access to namespaced resources, you need to use LocalSubjectAccessReview here.

address-controller/src/main/java/io/enmasse/controller/common/KubernetesHelper.java
```diff
@@ -344,6 +352,106 @@ public boolean isDestinationClusterReady(String clusterId) {
...
+            JsonObject resourceAttributes = new JsonObject();
+            resourceAttributes.put("namespace", namespace);
+            resourceAttributes.put("resource", "configmaps");
+            resourceAttributes.put("verb", verb);
+
+            spec.put("resourceAttributes", resourceAttributes);
+            spec.put("user", user);
+            body.put("spec", spec);
+
+            JsonObject responseBody= doRawHttpRequest("/apis/authorization.k8s.io/v1beta1/subjectaccessreviews", "POST", body, false);
```
luksa: Must use .../localsubjectaccessreviews endpoint

luksa:
While I don't like the idea of abusing features by not using them as they were intended to be used, abusing ConfigMaps is not any worse than abusing nonResourceUrls. So, I like this change. I guess you're ok with allowing users to actually change the ConfigMap manually?

You can now use this same mechanism for per-addressspace rights by granting access to a specific addressspace ConfigMap. You can even make it more fine-grained: users which can read a specific ConfigMap, can perform read operations inside the addressspace; those who can put/patch the ConfigMap, can perform update operations inside the addressspace.

lulf:
@luksa Exactly my thoughts as well, thanks for confirming! I'm doing some more changes that I will push (along with review fixes) where the basic-auth for OSB is removed and only supports RBAC. I will also change AuthInterceptor to only do authentication, and set a SecurityContext (which performs the subject access review) that can be used in the jax-rs handlers for performing more fine-grained authorization on address spaces and addresses. Do you think that is a viable approach?

luksa:
Sounds good!

Issue #432, opened 2017-10-30
Proposal: redesign of address model schema and console tooltips


lulf:
Currently the tooltips.json are generated from the documentation, and used by the console to display tooltips. Today, the format of the tooltips is restricted to a single address space type (one key is address.queue), so the brokered address space type cannot be surfaced as is unless we generate multiple tooltips, generate a large file with both, or massage the data in the console server. Those are all short-term workarounds, as ultimately we want the same information available through the /v1/schema API and service broker as well.

After discussing this with @grs we thought this might be a good time to think through how we're generating schema, how we're dealing with tooltips, and how we use swagger.

There are 2 aspects of this:

1. What is the source of truth for the schema?
2. How does the console use the schema _and_ display additional tooltip information like links?

For 1., we need to think about if our current approach of having the documentation as source from which /v1/schema, swagger and console content is generated. Another approach might be to have swagger.json be the source of truth, and instead inject information from that into documentation, /v1/schema, console etc. The downside is that descriptions in the schema are no longer part of the documentation, which perhaps makes the documentation harder to edit. On the other hand, the current approach makes it easy to mess up the console information by simply copy pasting documentation using the same keys for i.e. queue type descriptions.

For 2. it looks like the client console should have a more dynamic way of assembling the tooltips. One alternative is to have an API in the console server that the client console uses to retrieve the needed information. When performing a GET to this API, the console server would a) retrive the schema from the address controller (which is in turn based on one source of truth). It would then b) generate the tooltips.json used by the console client based on the schema + tooltips from a different file (which could also be generated from the source of truth).

There are many other alternative ways this could be done as well, so please propose alternatives or changes to the above proposal and lets have a discussion on how we'd like this to be in the future.

lulf:
@grs We should perhaps rethink this after the last roadmap plans

PR #481, opened 2017-11-08
handle tooltips for different address-space types

grs:
workaround to allow brokered address space to have tooltips pending longer term solution for #432


rgodfrey:
LGTM

PR #483, opened 2017-11-08
display actual route hostname in online help

grs:


lulf:
@grs LGTM although you can also get the messaging hostname from the address space endpoint list, which would be agnostic to the underlying routes or ingress.

PR #491, opened 2017-11-09
Add support for RBAC auth and authz for /v1/addresses

### LINKED ISSUE #418: Authenticate /v1/addresses REST API using RBAC
The address controller REST apis should be authenticated using kubernetes/openshift RBAC so we can control who are able to create addresses.

The agent service account needs to be given permission for creating addresses when an address space is created, and the agent needs to pass the service account token when using the REST API.

lulf:
This fixes #418


rgodfrey:
LGTM

PR #497, opened 2017-11-10
Enable RBAC for multitenant deployment

lulf:


lulf:
@grs This enables RBAC only in multitenant deployments, as it turns out cluster-admin privileges is required to use authentication.k8s.io API (while authorization is ok without such privileges).

PR #500, opened 2017-11-13
Add environment parameter to controller to support multiple controllers in a cluster

### LINKED ISSUE #474: Support multiple address controllers in the same OpenShift cluster
Allows running multiple independently managed messaging services in the same cluster can be useful if you need multiple environments (test, staging, prod) in the same cluster. 

To achieve this, the address controller needs to:
   * label each namespace it creates with a 'managed-by': $myns where '$myns' is the namespace where the address controller runs
   * only touch/delete namespaces managed by itself
   * add its own namespace as part of the managed namespace to avoid address space conflicts
   * ...

lulf:
An address controller may optionally be configured with an 'ENVIRONMENT'
environment variable (available through OpenShift templates). Each
namespace created by a controller is labeled with the same environment,
ensuring that a controller will only touch namespaces with for its own
environment.

This fixes #474

### DIFF
--- a/address-controller/src/main/java/io/enmasse/controller/ControllerHelper.java
+++ b/address-controller/src/main/java/io/enmasse/controller/ControllerHelper.java
@@ -16,7 +16,6 @@
 package io.enmasse.controller;
 
 import io.enmasse.config.AnnotationKeys;
-import io.enmasse.config.LabelKeys;
 import io.enmasse.controller.common.AuthenticationServiceResolverFactory;
 import io.enmasse.controller.common.Kubernetes;
 import io.enmasse.controller.common.KubernetesHelper;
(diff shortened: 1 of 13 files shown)


lulf:
> Personally, I would use a different name for the label (e.g. perhaps control-group, controlled-by, controller-id etc or even just group-id or similar).

I initially used a naming like 'controllerNamespace', but I think using the term 'environment' for this is very accurate for how this feature will be used ('dev', 'test', 'staging', prod'), so I'd like to stick with that as the concept is well known. 

> Again, just as my personal preference, it would be nice if it were 'optional', i.e. if no specific grouping was required no extra labels would exist and the controller wouldn't try to restrict its scope. That may not be possible given the nature of openshift templates though.

It is possible if we say that an empty value means it is not used, but I would perhaps instead choose a better name for the default case than 'development'. The idea is that when you deploy EnMasse in production, you have a desire to set these options to get control of what gets deployed anyway.

@rgodfrey I'd like hear your opinion on this as well. I'll merge it for now, and we can change the names and behavior later.

PR #527, opened 2017-11-14
Issue 513 add remove addresses auth

kornys:


systemtests/src/test/java/io/enmasse/systemtest/marathon/AuthenticationTest.java
```diff
@@ -45,18 +45,24 @@ public void testCreateDeleteUsersLong() throws Exception {
         Logging.log.info("testCreateDeleteUsersLong finished");
     }
 
-    private void createUsers(AddressSpace addressSpace, String prefixName, String prefixPswd, int from, int to)
+    void createUser(AddressSpace addressSpace, String username, String password) throws Exception{
```
tkratky: Maybe methods for creating/removing user(s) should be protected instead of "without modifier". Or is here some reason why shouldn't be?

PR #535, opened 2017-11-15
Fixes for agent:

grs:
* connect to address-controller in brokered space (setup ca secret and env var, use correct port)
* handle failure to delete/create addresses on broker better
* handle deletion of topics with existing subscriptions
* don't delete temporary queues

### DIFF
--- a/agent/lib/address_ctrl.js
+++ b/agent/lib/address_ctrl.js
@@ -17,7 +17,7 @@
 
 var fs = require("fs");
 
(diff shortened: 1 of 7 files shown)


lulf:
@grs Ah, it wasn't set for the brokered address space! :+1:

lulf:
@grs It should still use 443 instead of 8081 i think though, because the PORT_HTTPS is only available when running in single-tenant mode.

PR #549, opened 2017-11-20
Limit address-controller privileges and add cluster-roles

### LINKED ISSUE #473: Limit the scope of privileges required by address-controller
Although the address-controller currently requires cluster-admin permissions, it would be good to create a role that would contain the 'minimum' amount of permissions required in order to operate EnMasse.

An address controller needs to:
* create/delete/list namespaces

within each namespace it needs to:
* create/delete/list all resources
* create service accounts in namespaces it has created
* assign policies to service accounts

The enmasse-service-account should probably be renamed to something more descriptive and be assigned this role.

lulf:
* Add cluster roles to be used by EnMasse components
* Update deployment script with descriptions on how to create cluster
roles
* Use K8S RBAC for roles and clusterroles to avoid policy grant access

This should also provide support for K8S RBAC but it has not been
tested.

This fixes #473


address-controller/src/main/java/io/enmasse/controller/common/KubernetesHelper.java
```diff
@@ -295,6 +296,7 @@ private JsonObject doRawHttpRequest(String path, String method, JsonObject body,
         OkHttpClient httpClient = client.adapt(OkHttpClient.class);
 
         HttpUrl url = HttpUrl.get(client.getOpenshiftUrl()).resolve(path);
+        //log.info("Performing {} on {} with body '{}'", method, path, body.encodePrettily());
```
rgodfrey: nitpick: probably shouldn't add commented out code, if the log is potentially useful we could keep it at debug level

agent/www/help.html
```diff
@@ -1032,4 +1032,4 @@ <h5 id="amqp_net_lite">AMQP.Net Lite</h5>
 <script src="https://cdnjs.cloudflare.com/ajax/libs/highlight.js/9.12.0/highlight.min.js"></script>
 <script>hljs.initHighlighting()</script>
 </body>
-</html>
\ No newline at end of file
+</html>
```
rgodfrey: *sigh*

PR #550, opened 2017-11-20
Update pom.xml version on release

lulf:


.travis/build.sh
```diff
@@ -10,6 +10,7 @@ SYSTEMTEST_ARGS=${SYSTEMTEST_ARGS:-SmokeTest}
 
 if [ "$VERSION" != "latest" ]; then
     TAG=$VERSION
+    mvn versions:set -DnewVersion=$VERSION
```
rgodfrey: Does the change to the .pom files actually get committed (and tagged) in git?

lulf: No, I'm not sure we want to do that

release-scripts/update-snapshot-version.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/sh
+VERSION=$1
+
+git checkout master
+
+mvn versions:set -DnewVersion=${VERSION}
+
```
rgodfrey: don't we need to actually commit to git here?

lulf: Fixed!

PR #553, opened 2017-11-21
Collect events before deleting address space

lulf:


lulf:
@tkratky I'm not sure if this catches the address spaces that are shared between tests though, as the event collection happens only when an address space is deleted. It can probably be fixed using a watch instead

tkratky:
Logs are useful when tests failed and this case should be covered by: https://github.com/EnMasseProject/enmasse/blob/master/systemtests/src/test/java/io/enmasse/systemtest/MultiTenantTestBase.java#L43
 and other cases should be covered by following commit: https://github.com/EnMasseProject/enmasse/pull/553/commits/f60ee7855d673aa5067d708a5cf2c200dee829f3 - removing default address spaces once all tests finished/stoped.

lulf:
LGTM!

PR #555, opened 2017-11-21
Issue 554 ownerreferences namespaces

lulf:
@rgodfrey The second commit is just picked from master to get a jenkins build.


lulf:
@rgodfrey Looks like the test failure cause is that this feature only works on OpenShift >= 3.6

lulf:
Even though its supported in kubernetes 1.5, its not enabled in OpenShift 1.5: https://github.com/openshift/origin/issues/11951

PR #560, opened 2017-11-22
Replace use of Ingress with LoadBalancer for kubernetes

lulf:
This attempts to fix the broken Ingress support, which has become more broken after moving to HTTPS for
our endpoints (Ingress doesnt support TLS passthrough).

Instead, on kubernetes, expose services using LoadBalancer, which works nicely with minikube as it will expose these using NodePort. For
production deployments, type=LoadBalancer is what you'd typically want so I think this makes the kubernetes use case much better.

### DIFF
--- a/templates/install/kubernetes/addons/external-lb.yaml
+++ b/templates/install/kubernetes/addons/external-lb.yaml
@@ -1,73 +1,16 @@
 apiVersion: v1
-items:
-- apiVersion: v1
-  kind: Service
-  metadata:
-    annotations: {}
-    labels:
-      app: enmasse
-    name: address-controller-external
-  spec:
-    ports:
-    - name: https
-      port: 443
-      protocol: TCP
-      targetPort: https
-    selector:
-      name: address-controller
-    type: LoadBalancer
-- apiVersion: v1
-  kind: Service
-  metadata:
-    labels:
-      app: enmasse
-    name: messaging-external
-  spec:
-    ports:
-    - name: amqp
-      port: 5672
-      protocol: TCP
-      targetPort: 5672
-    - name: amqps
-      port: 5671
-      protocol: TCP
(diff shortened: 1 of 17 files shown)


templates/install/kubernetes/addons/external-lb.yaml
```diff
@@ -1,73 +1,16 @@
...
+  name: restapi-external
+spec:
+  ports:
+  - name: https
+    port: 443
+    protocol: TCP
+    targetPort: https
+  selector:
+    name: address-controller
+  type: LoadBalancer
```
ppatierno: Why here we have the service only for the address controller and not for all the other services (messaging, mqtt, ... ) ?

lulf: @ppatierno They are created based on the address space endpoints. By default, it creates the services for messaging, mqtt and console unless otherwise restricted. This allows different address spaces to get their own loadbalancer endpoints.

From the user side of things, the 'default' singletenant deployment will look like before, but with the endpoints as loadbalancer services instead of ingress. With minikube, it actually exposes loadbalancer services as nodeports, so you are able to more easily the services as well.

ppatierno: ok got it, thanks !

lulf: @ppatierno Relevant code is here: https://github.com/EnMasseProject/enmasse/pull/560/files#diff-f843557b1ea4676aeecc78f15d6db723R148

ppatierno: Yes I miss the point that the address controller is in charge to deploy all the other stuff so it has to create the services in the address space and then exposing them with loadbalancer. Thanks @lulf !

PR #565, opened 2017-11-22
simple authorization for console 

grs:
Based on keycloak groups returned by auth service. Enabled through KEYCLOAK_GROUP_PERMISSIONS env var. Being a member of address-space-admin group provides full access. Being a member of address-space-monitor provides read only access to all data. Being a member of view_<address-name> provides read only access to the address in question (no wildcards yet). User not in either address-space-admin or address-space-monitor can only see their own connections.

### DIFF
--- a/agent/lib/authz.js
+++ b/agent/lib/authz.js
@@ -1,0 +1,110 @@
+/*
+ * Copyright 2016 Red Hat Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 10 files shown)


agent/lib/authz.js
```diff
@@ -0,0 +1,110 @@
...
+    }
+};
+
+AuthorizationPolicy.prototype.can_publish = function (sender, message) {
+    var groups = sender.connection.options.groups;
+    if (this.has_permission('address-space-admin', groups) || this.has_permission('address-space-monitor', groups)) {
+        return true;
+    } else if (message.subject === 'address' || message.subject === 'address_deleted') {
+        //TODO: handle wildcards
+        return this.has_permission('view_' + encodeURIComponent(message.body.address), groups);
```
rgodfrey: is this actually checking if the user has permission to send a message to an address? if so, you could use "send_" here rather than "view_"... "send_" is what the broker uses for that case

grs: No, this is checking that the user is allowed to view the address. If it is the console will send it a message containing the details and stats of that address.

rgodfrey: I thought that's what it probably was, I didn't think we had anything as sophisticated in the console as sending messages :-)

PR #570, opened 2017-11-23
Refactor systemtest base classes

lulf:
This is an initial refactoring of the base classes so that they share
more functionality.

* Reuse the same client factories for tests to avoid leaking vertx instances.
* The address api client is now shared globally.


systemtests/src/test/java/io/enmasse/systemtest/TestBase.java
```diff
@@ -48,73 +48,57 @@
...
-            mqttClientFactory.close();
-        }
-        if (amqpClientFactory != null) {
-            amqpClientFactory.close();
-        }
-        if (addressApiClient != null) {
-            if (createDefaultAddressSpace()) {
-                setAddresses();
-            }
-            addressApiClient.close();
```
tkratky: addressApiClient is never closed, right?

systemtests/src/test/java/io/enmasse/systemtest/BrokeredTestBase.java
```diff
@@ -0,0 +1,28 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.enmasse.systemtest;
+
+public class BrokeredTestBase extends TestBaseWithDefault {
+
+    private static final String defaultAddress= "brokered-default-";
+    protected AddressSpace defaultBrokeredAddressSpace = new AddressSpace(defaultAddress + "0", AddressSpaceType.BROKERED);
```
tkratky: This shouldn't be there, right? and all inherited classes should use defaultAddressSpace instead of defaultBrokeredAddressSpace

systemtests/src/test/java/io/enmasse/systemtest/TestBaseWithDefault.java
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+
+    public AddressSpace getSharedAddressSpace() {
+        return defaultAddressSpace;
+    }
+
+    @Before
+    public void setupDefault() throws Exception {
+        if (environment.isMultitenant()) {
```
tkratky: this method doesn't exist since your last commit

lulf: Fixed :)

systemtests/src/test/java/io/enmasse/systemtest/BrokeredTestBase.java
lulf: You are right, removed.

systemtests/src/test/java/io/enmasse/systemtest/TestBase.java
lulf: No, but it is static now so it is only instantiated once throughout the entire testrun.

PR #571, opened 2017-11-23
Marathon fix close connection

kornys:


lulf:
It looks good, but another fix would be to close the connection factory. The fixes in #570 will make these changes void i think.

Issue #465, opened 2017-11-06
Report error events in controllers using k8s events


lulf:
The events should be labeled based on component and address space. This will be valuable for monitoring the service.

grs:
agent support in https://github.com/EnMasseProject/enmasse/pull/583

lulf:
I think we have events where we want them now.

PR #572, opened 2017-11-23
Add java interface for reporting events and a k8s implementation

lulf:
* Add use of events in ConfigMapAddressSpaceApi as an example

Issue #465

### DIFF
--- a/address-controller/src/main/java/io/enmasse/controller/standard/AddressController.java
+++ b/address-controller/src/main/java/io/enmasse/controller/standard/AddressController.java
@@ -4,6 +4,7 @@ import io.enmasse.address.model.AddressType;
 import io.enmasse.amqp.SyncRequestClient;
 import io.enmasse.controller.common.*;
 import io.enmasse.address.model.Address;
+import io.enmasse.controller.event.ControllerKind;
 import io.enmasse.k8s.api.*;
 import io.fabric8.kubernetes.api.model.Container;
 import io.fabric8.kubernetes.api.model.ContainerPort;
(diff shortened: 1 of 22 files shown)


address-controller/src/main/java/io/enmasse/controller/standard/AddressController.java
```diff
@@ -116,21 +116,26 @@ public synchronized void resourcesUpdated(Set<Address> newAddressSet) throws Exc
...
+            createBrokers(clusterList, addressByGroup);
+
+            // Perform status check
+            checkStatuses(newAddressSet);
+            for (Address address : newAddressSet) {
+                try {
+                    addressApi.replaceAddress(address);
+                } catch (KubernetesClientException ex) {
+                    log.warn("Error syncing address {}", address, ex);
+                    eventLogger.log("AddressSyncFailed", "Error syncing address: " + ex.getMessage(), "Error", "Address", address.getName());
```
rgodfrey: https://kubernetes.io/docs/api-reference/v1.8/#event-v1-core only talks of event types "Normal" and "Warning", is "Error" also valid (it seems perfectly sensible for it to be so, but the link only talks of "new types could be added in the future"

k8s-api/src/main/java/io/enmasse/k8s/api/EventLogger.java
```diff
@@ -0,0 +1,32 @@
...
+    /**
+     * Log an event.
+     *
+     * @param reason Created, Deleted etc.
+     * @param message Descriptive message
+     * @param type Severity: Normal, Error, Warning etc.
+     * @param objectKind Kind of object involved in event
+     * @param objectName Name of object involved in event
+     */
+    void log(String reason, String message, String type, String objectKind, String objectName);
```
rgodfrey: Is it worth creating constants for the even kind "Normal", "Warning" rather than having them subject to typos where this method is called?  The objectKind values should also probably come from constants though not defined here.

lulf: I'll add for Reason as well

PR #574, opened 2017-11-24
UPD AuthenticationTest - added 40s sleep after addresses are deployed

tkratky:


tkratky:
this branch was only experimental

PR #575, opened 2017-11-24
Dk add cli proton python

kornys:
#407

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -60,5 +60,6 @@ systemtests:
 
 client_install:
 	npm install -g cli-rhea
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -60,5 +60,6 @@ systemtests:
 
 client_install:
 	npm install -g cli-rhea
+	pip install cli-proton-python
```
tkratky: make sure that pip is installed in system and add it into the list of prerequisites: https://github.com/EnMasseProject/enmasse/blob/master/HACKING.md

PR #581, opened 2017-11-28
Consumers count tkratky

tkratky:


systemtests/src/main/java/io/enmasse/systemtest/BrokerManagement.java
```diff
@@ -0,0 +1,17 @@
+package io.enmasse.systemtest;
```
lulf: I think you can just merge this with ArtemisManagement as it doesn't really add anything by being in a separate  class.

systemtests/src/test/java/io/enmasse/systemtest/TestBase.java
```diff
@@ -71,12 +77,18 @@ public void setup() throws Exception {
         addressSpaceList = new ArrayList<>();
         amqpClientFactory = new AmqpClientFactory(openShift, environment, null, username, password);
         mqttClientFactory = new MqttClientFactory(openShift, environment, null, username, password);
+        managementAmqpClientFactory = new AmqpClientFactory(openShift,
+                environment,
+                null,
+                managementCredentials.getUsername(),
+                managementCredentials.getPassword());
```
lulf: Not sure why you'd create another factory here. The username/password is just about defaults. You can create clients where you set the username/password explicitly to something different instead.

systemtests/src/test/java/io/enmasse/systemtest/TestBaseWithDefault.java
```diff
@@ -74,8 +76,21 @@ public void setupDefault() throws Exception {
...
+        getKeycloakClient().createUser(defaultAddressSpace.getName(),
+                managementCredentials.getUsername(),
+                managementCredentials.getPassword());
+
+        createGroup(defaultAddressSpace, "admin");
+        joinGroup(defaultAddressSpace, "admin", managementCredentials.getUsername());
+
         amqpClientFactory = new AmqpClientFactory(openShift, environment, defaultAddressSpace, username, password);
         mqttClientFactory = new MqttClientFactory(openShift, environment, defaultAddressSpace, username, password);
+        managementAmqpClientFactory = new AmqpClientFactory(openShift,
```
lulf: Same as above, shouldn't be needed.

systemtests/src/main/java/io/enmasse/systemtest/BrokerManagement.java
tkratky: I would like to implement logic for apache activemq broker in the future (not sure if this broker will be used in enmasse)

systemtests/src/test/java/io/enmasse/systemtest/TestBase.java
tkratky: that's true, thank you

systemtests/src/main/java/io/enmasse/systemtest/BrokerManagement.java
lulf: It's not a big deal, but me personal preference is that such a refactoring should be done when its needed. I'm fine with the current approach.

PR #598, opened 2017-11-30
Initial draft of proposal on shared infrastructure for address spaces

lulf:
@grs @rgodfrey Just had to write down some of my ideas around the topic of shared infrastructure for address spaces, so that I don't forget it when we discuss it later and makes it easier for me to explain.


lulf:
Will reopen with new proposal based on latest f2f discussions

PR #616, opened 2017-12-07
UPD TestBaseWithDefault - remove shared address spaces after all tests

tkratky:


tkratky:
Those changes should save ~30 minutes in Jenkins

systemtests/src/test/java/io/enmasse/systemtest/SystemTestRunListener.java
```diff
@@ -20,6 +21,14 @@ public void testFinished(Description description) throws Exception {
         Logging.log.info(description + "FINISHED");
     }
 
+    @Override
+    public void testRunFinished(Result result) throws Exception {
+        for (AddressSpace defaultAddressSpace : TestBaseWithDefault.defaultAddressSpaces) {
```
lulf: IMHO its not ideal to make this global and reference it from another class, but I can't think of any better approach either :)

tkratky: yes, I wasn't able to find better solution :/

PR #617, opened 2017-12-07
Create realm admin user with random credentials

lulf:


lulf:
@rgodfrey The idea is that we automatically create a user and a password. Since this is not stored anywhere, it will have to be reset manually (or by a script). It just simplifies the operations needed to be done outside of enmasse.

lulf:
@rgodfrey actually Even better is to disable the user on creation and activate + set passord externally

lulf:
@rgodfrey actually Even better is to disable the user on creation and activate + set passord externally

PR #618, opened 2017-12-07
allow the messaging cert to be downloaded from the console

grs:


templates/include/admin.jsonnet
```diff
@@ -220,6 +229,12 @@ local common = import "common.jsonnet";
                 "secret": {
                     "secretName": "admin-internal-cert"
                 }
+            },
+            {
+                "name": "messaging-cert",
+                "secret": {
+                    "secretName": "external-certs-messaging"
```
lulf: This should be "${MESSAGING_SECRET}" as the secret name is derived from the address space definition and passed as a template parameter.

agent/lib/console_server.js
```diff
@@ -197,11 +214,15 @@ ConsoleServer.prototype.listen = function (env, callback) {
         } else if (request.method === 'GET') {
             try {
                 var user = myutils.basic_auth(request);
-                var transform;
+                var next;
                 if (url.parse(request.url).pathname === '/help.html' && env.MESSAGING_ROUTE_HOSTNAME !== undefined) {
```
lulf: This is not set anywhere is it?

templates/include/roles.jsonnet
```diff
@@ -8,6 +8,19 @@
         "name": "enmasse-address-admin"
       },
       "rules": [
+        {
+          "apiGroups": [
+            ""
+          ],
+          "resources": [
+            "routes"
```
lulf: why is this needed? how does it know the route name to look up?  its not necessarily named 'messaging' ,because it can be configured when creating address space.

agent/lib/console_server.js
grs: Yes, it is. It is set on startup based on the hostname in the route.

templates/include/roles.jsonnet
grs: The route is used to adjust the online help doc to have the actual hostname in it. Not having permission is a regression for that feature. If the route is named something else, this adjustment won't happen (just as it doesn't happen without the correct permission). I've added an env var to pass through the route name to lookup. However at the moment, I think fixing the permissions to allow the existing feature to work again is valuable as for those users who don't override the route name, it make it easy to find (especially if they don't have access to the actual route object in openshift). Expanding to cover the more advanced case or indeed revising the mechanism entirely can be done as a separate PR.

lulf: Agreed this is sufficient for now.

PR #622, opened 2017-12-11
Cleanup templates and remove kafka-bridge templates

lulf:
* _REPO -> _IMAGE
* Remove with_kafka parameter
* Stop generated enmasse-with-kafka template
* Remove old role file


lulf:
@ppatierno I think both the fact that that it doesn't work (doesnt have authentication + authorization enabled, image version released does not match template versions etc...),  and because we eventually want to support it in an official way (though that is not an immediate plan).

ppatierno:
@lulf in order to have this demo (https://github.com/ppatierno/amqp-kafka-demo) still working (for people who are curious about this kind of support) can you point me to the latest EnMasse version that should work so that I can update that repository ? Do you think that it should be the 0.9.0 ?

ppatierno:
@lulf for that demo I have to update the references to "barnabas" as well but I'd like to have it working even if we are going to remove the Kafka template from EnMasse (which makes sense to me).

lulf:
@ppatierno I think maybe 0.11.2 should work

PR #629, opened 2017-12-12
Split standard-specific parts of address-controller into a new component 'standard-controller'

lulf:
The standard-controller is responsible for managing brokers in the
standard address space and update status of addresses based on broker
and router status.


lulf:
@grs This is the first phase of moving the controller loop for standard address space into the address space.

lulf:
@grs WDYT about merging this? This only affects the standard address space

templates/include/standard-space-infra.jsonnet
```diff
@@ -32,19 +35,34 @@ local prometheus = import "prometheus.jsonnet";
...
+    ],
+
+    local storage_template_configmap = [
+      templateConfig.storage("enmasse-storage-templates")
+    ],
+
+    local template_config = (if use_template_configmap then "enmasse-storage-templates" else ""),
+
+    "objects": [
+      standardController.deployment(template_config, "authservice-ca"),
```
grs: Is there a reason you prefer this outside the admin pod?

lulf: No, I can move it in there instead. It was a bit easier to experiment with it separately.

grs: I have no objection to merging

PR #647, opened 2017-12-14
Add support for prometheus metrics for Artemis

lulf:


templates/install/openshift/enmasse.yaml
```diff
@@ -1524,6 +1486,8 @@ objects:
               name: amqps-normal
             - containerPort: 8161
               name: jolokia
+            - containerPort: 8080
+              name: metrics
```
ppatierno: From my experience with Prometheus in Barnabas ...
If you are going to use a Prometheus server for scraping the metrics, you can use the kubernetes discovery feature on pods. In order to avoid having Prometheus scraping metrics on all container ports which don't expose metrics at all (i.e. 8161 here), you can configure a relabel config matching a specific port name. As you, I started to use "metrics" as port name and then it turned out that a container in the K8S dns pod exposes metrics with a port having this name. It means that the Prometheus server will get them, which could be something you don't want. For this reason I switched to "kafkametrics" :-)

lulf: Ah, good to know, thanks!

PR #662, opened 2017-12-18
Update partitionPV from upstream

lulf:


lulf:
@grs There is a potential issue here when upgrading with the data directory rename from artemis-N to split-N. Also the addition of the .data_initialized might cause some trouble when upgrading.

I'm not sure if we should worry about that at this point though.

PR #673, opened 2017-12-20
Js refactor

grs:


grs:
Yes, the title of this PR was too general! I did not intend to imply it is the only refactoring needed. I agree that there is plenty more to do. This is really just the first step, to eliminate duplicated (and diverging) code.

lulf:
Ok, thats fine. Was just wondering how one would go about doing that, and if its something that goes against JS convention :)

Merge at will!

grs:
I think what is needed is more clarity on the abstractions relied on by different components, and then ensure translation as necessary on input/output from those. i.e. better defined (and better designed) interfaces between different pieces.

agent/lib/pod_watcher.js
```diff
@@ -0,0 +1,73 @@
+/*
+ * Copyright 2016 Red Hat Inc.
+ *
```
rgodfrey: (c) 2017 perhaps :)

agent/lib/topic.js
```diff
@@ -0,0 +1,40 @@
+/*
+ * Copyright 2016 Red Hat Inc.
```
rgodfrey: 2017 :)

lulf:
Merged manually via cli

PR #691, opened 2018-01-05
Fix unstable colocated queue test

lulf:


lulf:
@grs @tkratky @rgodfrey The last jenkins failed due to standard-controller crashing due to its restrictive memory limits (64MB heap). After bumping the limit, the test is green.

PR #700, opened 2018-01-09
UPD Jenkinsfile - archive artifacts when build timeout was reached

tkratky:


Jenkinsfile
```diff
@@ -41,16 +41,14 @@ node('enmasse') {
                         sh 'PATH=$PATH:$(pwd)/systemtests/web_driver DISPLAY=:10 ARTIFACTS_DIR=artifacts OPENSHIFT_PROJECT=${JOB_NAME::16}${BUILD_NUMBER} ./systemtests/scripts/run_test_component.sh templates/install /var/lib/origin/openshift.local.config/master/admin.kubeconfig systemtests'
                     } finally {
                         junit '**/TEST-*.xml'
+                        archive '**/TEST-*.xml'
```
lulf: This will cause artifacts to be stored on sucessful runs as well I think + it would not store artifacts if test fails in unit test or other build steps.

PR #717, opened 2018-01-12
use a router with address-authz support

grs:


lulf:
@grs Merged this by a mistake. Reverted on master, but will create another PR with this change and systemtest fixes.

PR #729, opened 2018-01-16
fixed docs content issues

jenmalloy:


lulf:
@jenmalloy @nengard This broke the CI build, so I'll revert this change on master. The reason is that the attributes used to generate the console tooltips were missing.

PR #738, opened 2018-01-18
Generate templates with correct docker registry

lulf:


tkratky:
@lulf thank you for the fix! I presumed that deploy_openshift script execute "make -C templates", so it was my fault!

PR #743, opened 2018-01-19
modify package target for agent to include the runtime dependencies

grs:
in the tarball


agent/install_into
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+
```
lulf: Can this just be moved into the makefile target? I.e

```
cd build; npm install --production
```

The cd will only affect the shell invocation on that line.

grs: Yes, that is much nicer! Done now.

PR #769, opened 2018-01-22
Issue 742 scheduler improvements

lulf:


lulf:
@grs useEventLoop is perhaps a bad name. The intention is that it will not invoke the addressesChanged() on every change to the address resources, but scan them periodically. This to hopefully reduce the potential backlog of blocked threads.

PR #785, opened 2018-01-24
Enable authz by default in brokered address space

rgodfrey:


kornys:
I think the issue in branch issue_784 is not with SmokeTest. Because following scenario does not work: create brokered address space -> create user test a and add them to groups send_*/recv_* -> create queue -> run sender and receiver under user -> receiver retunrs Authorization error: user does not have permissions to 'CONSUME', same bahavior is when I create queue and then create user and add them to groups.

kornys:
@rgodfrey there is another issue see #811

PR #787, opened 2018-01-24
Issue 736 fixes

lulf:


amqp-utils/src/main/java/io/enmasse/amqp/Artemis.java
```diff
@@ -146,7 +148,7 @@ private Message doAttribute(String resource, String attribute, Object ... parame
     }
 
     private Message doRequestResponse(Message message, Object ... parameters) throws TimeoutException {
-        return doRequestResponse(5, TimeUnit.SECONDS, message, parameters);
+        return doRequestResponse(10, TimeUnit.SECONDS, message, parameters);
```
rgodfrey: I can't help feeling that timeouts should come from config or env variables rather than only being changable by recompiling the code... however the issue is not new in this change :)

amqp-utils/src/main/java/io/enmasse/amqp/Artemis.java
```diff
@@ -46,15 +46,17 @@
     private final Vertx vertx;
     private final ProtonConnection connection;
     private final ProtonSender sender;
+    private final ProtonReceiver receiver;
```
rgodfrey: I see that we now pass this in to the constructor, and initialize there... but is it actually being used somewhere, or is this just something to stop things being collected by GC or something...?

artemis/amqp-connector/src/main/java/org/apache/activemq/artemis/integration/amqp/AMQPConnectorServiceFactory.java
```diff
@@ -72,6 +74,7 @@ public ConnectorService createConnectorService(String connectorName, Map<String,
       Map<String, Object> connectorConfig = new HashMap<>();
       connectorConfig.put(TransportConstants.HOST_PROP_NAME, configuration.get(TransportConstants.HOST_PROP_NAME));
       connectorConfig.put(TransportConstants.PORT_PROP_NAME, configuration.get(TransportConstants.PORT_PROP_NAME));
+      connectorConfig.put(TransportConstants.NETTY_CONNECT_TIMEOUT, "10000");
```
rgodfrey: Similar comment to before on the hardcoding of timeouts

artemis/amqp-connector/src/main/java/org/apache/activemq/artemis/integration/amqp/AMQPConnectorServiceFactory.java
```diff
@@ -50,6 +50,8 @@
       properties.add(TransportConstants.KEYSTORE_PASSWORD_PROP_NAME);
       properties.add(TransportConstants.TRUSTSTORE_PATH_PROP_NAME);
       properties.add(TransportConstants.TRUSTSTORE_PASSWORD_PROP_NAME);
+      properties.add(TransportConstants.USE_GLOBAL_WORKER_POOL_PROP_NAME);
+      properties.add(TransportConstants.REMOTING_THREADS_PROPNAME);
```
rgodfrey: What does this change do?

amqp-utils/src/main/java/io/enmasse/amqp/Artemis.java
lulf: It was to be able to issue credits manually, which I removed. But on the other hand it felt natural to leave it in just in case :)

amqp-utils/src/main/java/io/enmasse/amqp/Artemis.java
lulf: Agreed, I will fix that because it might be quite useful.

artemis/amqp-connector/src/main/java/org/apache/activemq/artemis/integration/amqp/AMQPConnectorServiceFactory.java
lulf: This allows those settings to be set in the broker XML and override defaults.

artemis/amqp-connector/src/main/java/org/apache/activemq/artemis/integration/amqp/AMQPConnectorServiceFactory.java
lulf: I'll add it to the list of allowed properties as well so that it can be overridden.

PR #794, opened 2018-01-25
Issue 792 evaluation installing docs

lulf:
@jenmalloy @grs  Please ignore the first commit. The second commit is what needs reviewing


documentation/service_admin/installing-openshift.adoc
```diff
@@ -153,3 +159,55 @@ The deployments required for running {ProductName} are now created.
 ----
 oc get pods -n enmasse
 ----
+
+===== Deploying {ProductName} for evaluation
```
grs: I think of it more as deploying for single tenant use. The terms production/evaluation suggest that the latter is somehow less reliable, which I don't think is actually the issue.

lulf: @grs I wasn't sure what to call it really. At present I don't think we should set the expectation that it is supported in the same way, because we don't have any CI tests for this mode ( so it kind of is less reliable in that sense).

documentation/service_admin/installing-openshift.adoc
```diff
@@ -96,6 +96,12 @@ oc process -f ./openshift/addons/standard-authservice.yaml | oc create -f -
 
 ==== Deploying {ProductName}
 
+To deploy {ProductName} in production, it is recommended that you have cluster-admin access in order
+to setup the required roles for creating namespaces and managing resources in those namespaces. For
```
jenmalloy: set up

documentation/service_admin/installing-openshift.adoc
```diff
@@ -153,3 +159,55 @@ The deployments required for running {ProductName} are now created.
...
+
+. Create a certificate to use for the address controller. For testing purposes, you can create a self-signed certificate:
++
+[options="nowrap"]
+----
+mkdir -p address-controller-cert/
+openssl req -new -x509 -batch -nodes -days 11000 -subj "/O=io.enmasse/CN=address-controller.enmasse.svc.cluster.local" -out address-controller-cert/tls.crt -keyout address-controller-cert/tls.key
+----
+
+. Create secret with address-controller certificate:
```
jenmalloy: Create a secret for(?) the address-controller certificate:

documentation/service_admin/installing-openshift.adoc
```diff
@@ -153,3 +159,55 @@ The deployments required for running {ProductName} are now created.
...
++
+[options="nowrap"]
+----
+oc create secret tls address-controller-cert --cert=address-controller-cert/tls.crt --key=address-controller-cert/tls.key
+----
+
+. Deploy address-controller:
```
jenmalloy: Deploy the address controller(?):

documentation/service_admin/installing-openshift.adoc
```diff
@@ -96,6 +96,12 @@ oc process -f ./openshift/addons/standard-authservice.yaml | oc create -f -
...
 
+To deploy {ProductName} in production, it is recommended that you have cluster-admin access in order
+to setup the required roles for creating namespaces and managing resources in those namespaces. For
+evaluation purposes, {ProductName} can also be deployed in a single namespace.
+
+===== Deploying {ProductName} for production
+
 .Procedure
 
 . Create service account for the {ProductName} address controller:
```
jenmalloy: Create a service account...

PR #796, opened 2018-01-25
initial description of suggested plans mechanism

grs:


documentation/design_docs/design/plans.adoc
```diff
@@ -0,0 +1,174 @@
...
+            "name": "router",
+            "min": 1,
+            "max": 5
+        },
+        {
+            "name": "aggregate",
+            "max": 6
+        }
+    },
+    "addressPlans":["queue-myplan", "topic-small", "topic-large"]
```
lulf: Did we agree on making templates optional or separate in the first pass? If optional, we could add a
```
"template": {
  "templateName": "standard-space-infra",
  "templateParameters": [
    {
       "name": "MY_TEMPLATE_PARAMETER",
       "value": "42"
    }
}
```

documentation/design_docs/design/plans.adoc
```diff
@@ -0,0 +1,174 @@
...
+    "requiredResources": [
+        {
+            "name": "broker",
+            "credit": 0.05
+        },
+        {
+            "name": "router",
+            "credit": 0.01
+        }
+    ]
```
lulf: Should we also add an optional field 'clusterName' here for the colocated plans?

documentation/design_docs/design/plans.adoc
```diff
@@ -0,0 +1,174 @@
...
+deployed.
+
+An AddressSpacePlan has the following fields:
+
+* name (this would be the name of the config map)
+
+* displayName (a possibly different name more suitable for displaying
+  to users, with less constraints on allowed chars)
+
+* description
```
lulf: split into shortDescription and longDescription

lulf:
@grs We should probably specify fields that are optional (if any). Should all of name, displayName, shortDescription and longDescription be required?

documentation/design_docs/design/plans.adoc
grs: What would be the purpose of that field? Do we actually need it if the brokers in the pool have a stable identity and the standard-controller is marking addresses with the identity of the broker it has been allocated to?

documentation/design_docs/design/plans.adoc
grs: If we envisage perhaps moving it out at a later stage (and overhauling it), my instinct would be to keep it separate.

However as long as it is optional and there are the appropriate warnings about using it, I don't have a problem including it in the plans.

grs:
I would say name and addressType and requiredResources are the only required fields. The displayName and descriptions are optional and are just about providing a better experience for anyone using a gui to provision addresses.

documentation/design_docs/design/plans.adoc
lulf: My main motivation for including them in the plan would be that it would make converting the current static plan interface to the configmap-backed ones a bit easier.

documentation/design_docs/design/plans.adoc
lulf: It would allow us to have multiple pooled broker deployments. Without it, I think we have to hardcode the name like we do at present. I just think its a small addition which gives a lot more flexibility if you want isolated broker clusters for some pooled addresses.

grs: Would that not just be a parameter to the template?

lulf: If that will be the mechanism, it still needs to know if it should instantiate that template or not. For sharded queues, the cluster name is the name of the address. For pooled queues, it is currently the name of the plan. But it wouldn't have any way to know which of them to choose. So an alternative would be to have an attribute 'pooled' but it would still need to know the name it should use.

grs: An address can be pooled if the broker resource required as specified in the associated plan is < 1.  If an address plan has a template and parameters associated with it, I would assume that the template is invoked when the first address referencing that template and  parameter combination is created. Subsequent use of the same template+params would share or scale up as needed. Perhaps there should be something that ties a named template and specific parameters to some identifier?

E.g. resource_constructors = [
    {
        name: simple-pooled-queue,
        template: pooled-queue
        parameters: [
            {
               "name": "MY_TEMPLATE_PARAMETER",
               "value": "42"
            }
        ]
    },
   ....
]

This would be address-space level, and the the address-plans would be mapped to one of these. (Makes it easier to detect where two plans reference the same set of params and where they do not).

How would the cluster name be used? Is it to identify the deployment/stateful-set to scale up? Is it just about avoid naming clashes.

lulf: Ah, yes, I forgot about that the requiredResources would dictate what to choose. Then i just want some way to identify the deployment.

The cluster name is one way to identify the deployment (right now each deployment has a clustername which is also put into the cluster_id annotation of the pod, to make service discovery easier), but I'm open to alternatives.

Would the resource_constructors be copied into each address space then and referenced by the address plan? So in this case, the plan would have some field 'resourceConstructor' that would reference this?

In the non-pooled case it would always invoke the constructor, passing the address as the 'clusterName' template parameter always. In the pooled case, the 'clusterName' template parameter would be hardcoded in the resource constructor?

grs: I think perhaps standardising a template parameter that is the cluster_id (not crazy about that name, but thats what it is now). That is anyway required for any template used for a non-pooled resource. Making the existence and correct operation of such a parameter a condition on all templates makes sense to me. That way the standard controller can choose the name used in all cases. It does already for non-pooled, where it uses the address name. I do also think we need to associate a name with template+parameters if we are to actually support different parameters being specified for the same template. That name could then be used as the value for the cluster_id parameter when creating the resource for pooled addresses.

PR #799, opened 2018-01-25
allow dynamic source and anonymous sender

### LINKED ISSUE #789: Using qdstat/qdmanage with plain SASL auth doesn't work
When using the qdstat tool to query the dispatch router like this
`qdstat -a --sasl-username=user --sasl-password=pw --sasl-mechanisms=PLAIN -b amqps://192.168.99.100:30071`
I get the error
`LinkDetached: sender 720d2bf2-e2f1-44f5-aa63-19aaaefffa13-$management to $management closed due to: Condition('amqp:unauthorized-access', 'link disallowed by local policy')`
So far that is expected as the user isn't in any "send/recv_[*]" Keycloak groups.

But after having assigned Keycloak groups `send_%24management` and `recv_%24management` (ie. send/recv_$management) to the user, there still is the error
`LinkDetached: receiver ab1c4941-1d8e-4b6a-901d-7f2dcb3c1955-a9ee7336-0a32-4877-90b8-2cc0aba75067 from None closed due to: Condition('amqp:unauthorized-access', 'link disallowed by local policy')`

Also assigning the groups `send_%2A` and `recv_%2A` (for send/recv_*) doesn't help.

grs:
Fixes #789

### DIFF
--- a/router/Dockerfile
+++ b/router/Dockerfile
@@ -1,4 +1,4 @@
-FROM enmasseproject/qdrouterd-base:1.0.0-fix4
+FROM gordons/qdrouter-base:DISPATCH-920
 ARG version=latest
 
 COPY ./run_qdr.sh ./qdrouterd.conf.template colocated-topic.snippet subscriptions.snippet amqp-kafka-bridge.snippet policy.json /etc/qpid-dispatch/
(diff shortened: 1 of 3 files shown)


lulf:
@grs Looks like there is some issues with topics, as repeated runs fail on the smoketest when using topics.

PR #803, opened 2018-01-26
Initial proposal on resource definitions

lulf:


documentation/design_docs/design/resource-definitions.adoc
```diff
@@ -0,0 +1,99 @@
...
+and address plans need to exist.
+
+A resource definition would be specified in a configmap containing the actual definition in one of
+the fields. The fields of a resource definition is:
+
+name [required], name of the resource
+
+template [optional], name of the template to instantiate
+
+template [optional], parameters to pass to the template
```
grs: typo? should this be parameters?

documentation/design_docs/design/resource-definitions.adoc
```diff
@@ -0,0 +1,99 @@
...
+
+An example resource definition:
+
+```
+{
+    "apiVersion": "enmasse.io/v1",
+    "kind": "ResourceDefinition",
+    "metadata": {
+        "name":
```
grs: Minor, but would be clearer to show a name here since that is required.

documentation/design_docs/design/resource-definitions.adoc
```diff
@@ -0,0 +1,99 @@
...
+
+The resource definition configmap would look like this (in YAML for readability)
+
+```
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: resource-definition-foo
+  labels:
+    type: resource-definition-config
```
grs: Minor, why resource-definition-config instead of just resource-definition?

documentation/design_docs/design/resource-definitions.adoc
```diff
@@ -0,0 +1,99 @@
...
+                "value": "val1"
+            }
+        ]
+    }
+```
+
+All resource definitions are stored and validated globally by the address controller. When an
+address space is created, the resource definitions are copied into the address space.
+
+The ResourceDefinition can be referenced by both AddressPlans and AddressSpacePlans. The plans are
```
grs: Perhaps better to say 'A', rather than 'The' here? The point being that it sis unlikely you would ever refer to the same resource definition from AddressPlan and AddressSpacePlan.

PR #804, opened 2018-01-26
UPD Jenkinsfiles - added commit sha parameter for running systests

tkratky:


tkratky:
I'll wait with merge until Monday morning

Jenkinsfile
```diff
@@ -28,6 +28,7 @@ pipeline {
         stage('checkout') {
             steps {
                 checkout scm
+                sh 'echo $(git log --format=format:%H -n1) > actual-commit.file'
```
lulf: Actually, jenkins expose a GIT_COMMIT variable which you can use instead perhaps?

tkratky:
@lulf updated and looks like it works correctly, thank you for the tip! 👍

PR #821, opened 2018-01-30
Ensure that only address spaces active for a given testclass are kept

lulf:
This should prevent the default address spaces to exist when tests that do not require them are running. This reduces the resource consumption at the cost of having to recreate the default space in the next test.

Ideally, we should find a way to order tests so that those using the default address spaces execute before any of the other tests.


tkratky:
I'm little bit worried about the time required for executing all tests without controlling order of the tests (as you said)
otherwise lgtm

tkratky:
lgtm! thank you for that 👍

tkratky:
@lulf I renamed TestBaseWithDefault to TestBaseWithShared in last commit, it does more make sense to me, but feel free to revert this commit if you disagree

PR #826, opened 2018-02-01
Proposal of plan-based quota check, scaling and scheduling

lulf:
@grs @rgodfrey 

This is a proposal with a pseudo-algorithm that can be implemented in standard-controller to:

* check quota
* scale/create deployments
* schedule pooled addresses

### DIFF
--- a/documentation/design_docs/design/resource-scheduler.adoc
+++ b/documentation/design_docs/design/resource-scheduler.adoc
@@ -1,0 +1,161 @@
+== why
+
+The new plans introduced requires a new way of deploying the infrastructure and scheduling pooled
+queues based on the credits.
+
+== what/how
+
+The algorithm assumes that address phases are implemented.
+
+=== Decomposed algorithm overview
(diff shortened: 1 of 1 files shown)


documentation/design_docs/design/resource-scheduler.adoc
```diff
@@ -0,0 +1,225 @@
...
+
+A variant where the usage and need calculations are the same and there is a clear distinction
+between sharded and pooled behavior throughout.
+
+. Calculate actual usage of created addresses:
++
+```
+usage = {}
+detailed_usage = {}
+for a in all addresses with phase != Pending
```
grs: is it all addresses in Pending, or just Active addresses?

grs: I see there is now a Configuring state, so it would certainly be Running + Configuring. Question I guess is whether Terminating and Failed addresses are counted.

PR #831, opened 2018-02-01
UPD enable core dumps in Jenkins

tkratky:


lulf:
I also propose that the build should fail if a core dump is found

PR #833, opened 2018-02-01
Change agent to create and delete address configmaps directly, rather…

grs:
… than requesting address-controller to do so.

Use address-plan configmaps to determine what plans and types console should offer.
The agent now no longer has any dependency on the address-controller at runtime.


lulf:
@grs FYI something is wrong with the Travis CI tests( they don’t run, my mistake)

agent/lib/internal_address_source.js
```diff
@@ -93,4 +95,98 @@ AddressSource.prototype.check_status = function (address_stats) {
...
+            name: definition.address,
+            addressSpace: this.address_space
+        },
+        spec: {
+            address: definition.address,
+            type: definition.type,
+            plan: definition.plan
+        },
+        status: {
+            isReady: 'False',
```
lulf: I think this is a boolean and not a string field.

lulf: Also, the status field is not required when creating an address through the rest api.

agent/lib/internal_address_source.js
```diff
@@ -93,4 +95,98 @@ AddressSource.prototype.check_status = function (address_stats) {
...
+        apiVersion: 'v1',
+        kind: 'ConfigMap',
+        metadata: {
+            name: configmap_name,
+            labels: {
+                type: 'address-config'
+            },
+            annotations: {
+                addressSpace: this.address_space,
+                cluster_id: configmap_name
```
lulf: This should not be the name of the configmap but the name of the address (i.e. definition.address).

Issue #834, opened 2018-02-02
Transaction related exceptions in Keycloak-plugin


calohmn:
Using 0.16.0:
In the keycloak container, we've encountered the following exceptions (independent of each other):

1. In PlainSaslServerMechanism:
````
13:13:41,235 SEVERE [io.vertx.core.impl.ContextImpl] (vert.x-eventloop-thread-0) Unhandled exception: java.lang.IllegalStateException: Cannot call evict() without a transaction
	at org.keycloak.models.cache.infinispan.UserCacheSession.evict(UserCacheSession.java:112)
	at org.keycloak.credential.PasswordCredentialProvider.isValid(PasswordCredentialProvider.java:218)
	at org.keycloak.credential.UserCredentialStoreManager.validate(UserCredentialStoreManager.java:143)
	at org.keycloak.credential.UserCredentialStoreManager.isValid(UserCredentialStoreManager.java:133)
	at org.keycloak.credential.UserCredentialStoreManager.isValid(UserCredentialStoreManager.java:97)
	at io.enmasse.keycloak.spi.PlainSaslServerMechanism$1.processResponse(PlainSaslServerMechanism.java:85)
	at io.enmasse.keycloak.spi.SaslAuthenticator.process(SaslAuthenticator.java:117)
	at io.vertx.proton.impl.ProtonServerImpl$1.process(ProtonServerImpl.java:178)
	at io.vertx.proton.impl.ProtonTransport.processSaslAuthentication(ProtonTransport.java:205)
	at io.vertx.proton.impl.ProtonTransport.handleSocketBuffer(ProtonTransport.java:188)
	at io.vertx.core.net.impl.NetSocketImpl.handleDataReceived(NetSocketImpl.java:325)
	at io.vertx.core.net.impl.NetServerImpl.handleMsgReceived(NetServerImpl.java:87)
````

2. In AmqpServer:
````
07:38:11,132 SEVERE [io.vertx.core.impl.ContextImpl] (vert.x-eventloop-thread-0) Unhandled exception: java.lang.IllegalStateException: Cannot access delegate without a transaction
	at org.keycloak.models.cache.infinispan.RealmCacheSession.getDelegate(RealmCacheSession.java:142)
	at org.keycloak.models.cache.infinispan.RealmCacheSession.getGroupById(RealmCacheSession.java:780)
	at org.keycloak.models.cache.infinispan.RealmAdapter.getGroupById(RealmAdapter.java:1239)
[...]

PR #837, opened 2018-02-02
Use txn when accessing model objects

rgodfrey:


tkratky:
https://github.com/EnMasseProject/enmasse/issues/834

PR #852, opened 2018-02-07
New test for console auth

kornys:


systemtests/src/test/java/io/enmasse/systemtest/bases/web/WebConsoleTest.java
```diff
@@ -544,12 +550,26 @@ public void doTestViewAddressesWildcards() throws Exception {
         }
     }
 
+    protected void doTestCanOpenConsolePage(String username, String password, boolean canOpen) throws Exception {
+        String blankPage = "<html><head></head><body></body></html>";
```
tkratky: check if body exists and body is empty could be better solution

PR #854, opened 2018-02-07
NEW TestBase - method for check count of subscribers via console

tkratky:
- 'get web browser drivers' moved into TestBase
- standard clients test(testMessageSelectorTopic, testTopicSubscribe) enabled for rhea, proton(java, python)


tkratky:
thanks @kornys  for help

PR #859, opened 2018-02-08
Reduce initial delay for keycloak deployment

### LINKED ISSUE #853: Reduce readinessProbe-initialDelaySeconds for Keycloak deployment
In the deployment yml for Keycloak, the readinessProbe-initialDelaySeconds is set to 120s.
I think this is quite high - we have seen startup times for Keycloak between 10s and 40s - and can be reduced.

lulf:
This fixes #853


kornys:
@lulf do you think that this can be merged?

lulf:
@kornys I'll rebase it against master and see what the test results will be. I wasn't sure why the PR tests were failing at the time.

Issue #660, opened 2017-12-18
Brokered: openwire client raises JMSSecurityException when try to send messages on topic


kornys:
All systemtests with openwire client are passed, but topic tests fail on JMSSecurityException when sender try to send messages on topic. Receivers work fine, all receiveres are connected and queues on topic are created.

io.enmasse.systemtest.brokered.clients.openwire.MsgPatternsTest.testTopicSubscribe

```
512617 [main] ERROR test - javax.jms.JMSSecurityException: AMQ119032: User: test does not have permission='CREATE_ADDRESS' on address topic-subscribe
	at org.apache.activemq.artemis.core.protocol.openwire.OpenWireConnection.convertException(OpenWireConnection.java:359)
	at org.apache.activemq.artemis.core.protocol.openwire.OpenWireConnection.bufferReceived(OpenWireConnection.java:276)
	at org.apache.activemq.artemis.core.remoting.server.impl.RemotingServiceImpl$DelegatingBufferHandler.bufferReceived(RemotingServiceImpl.java:642)
	at org.apache.activemq.artemis.core.remoting.impl.netty.ActiveMQChannelHandler.channelRead(ActiveMQChannelHandler.java:68)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:362)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:348)
	at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:340)
	at io.netty.handler.codec.ByteToMessageDecoder.fireChannelRead(ByteToMessageDecoder.java:310)
	at io.netty.handler.codec.ByteToMessageDecoder.channelRead(ByteToMessageDecoder.java:284)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:362)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:348)
	at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:340)
	at io.netty.handler.ssl.SslHandler.unwrap(SslHandler.java:1336)
	at io.netty.handler.ssl.SslHandler.decodeJdkCompatible(SslHandler.java:1127)
[...]

PR #871, opened 2018-02-12
Permit senders to open non-durable queues

lulf:
This allows OpenWire clients to work out of the box


kornys:
fixes https://github.com/EnMasseProject/enmasse/issues/660

PR #873, opened 2018-02-12
Issue 746 plan based scheduling

lulf:


grs:
I get a compilation error locally when trying to build this branch. Looks like travis hits it also:

[ERROR] COMPILATION ERROR : 
[ERROR] /home/gordon/projects/enmasse/address-controller/src/main/java/io/enmasse/controller/api/v1/AddressApiHelper.java:[66,63] incompatible types: io.enmasse.address.model.Schema cannot be converted to io.enmasse.address.model.v1.SchemaProvider

lulf:
@grs Compilation should be fixed now. There are a few assumptions on the agent being able to do the following:

* Only configure addresses with phase != Terminating && phase != Pending
* Create queues (and connector) on brokers based on the annotation

I would also be nice to move the status check from the standard-controller to agent.

Issue #843, opened 2018-02-05
system-tests: plan resource requests for addresses/address-spaces


tkratky:
create new tests for newly defined plans for addresses and address-spaces

PR #874, opened 2018-02-12
Issue 843 new plans

tkratky:


tkratky:
related issue: https://github.com/EnMasseProject/enmasse/issues/843

tkratky:
Yes that's true, I should move this test into standard and rewrite it (little bit). I added this test into brokered because of time required for creating new address space. And original purpose of this test was verify that is possible to use dynamically created plan. I'll fix that.

PR #875, opened 2018-02-13
Use ansible to deploy enmasse in systemtests

lulf:


kornys:
So before merge please update jenkins manage files for openstack machines, there is no ansible installed. ansible-playbook: command not found

systemtests/scripts/test_func.sh
```diff
@@ -20,21 +20,14 @@ function setup_test() {
     export CURDIR=`readlink -f \`dirname $0\``
```
kornys: @lulf why there is still export env var?

kornys:
@lulf - tests in CI fails, but deploy is ok -> all pods are up.

there are 2 kinds of exception:
1.
Caused by: io.fabric8.kubernetes.client.KubernetesClientException: Failure executing: GET at: ****/apis/route.openshift.io/v1/namespaces/enmasse-pr-enma4/routes/restapi. Message: Forbidden! User ****/127-0-0-1:8443 doesn't have permission. User "system:anonymous" cannot get routes.route.openshift.io in project "enmasse-pr-enma4".

2. and this weird exception:
java.lang.NoClassDefFoundError: Could not initialize class io.enmasse.systemtest.standard.QueueTest

kornys:
@lulf your last commit fixed issues, I think that you can merge with master.

PR #881, opened 2018-02-13
Broker id annotation

grs:


lulf:
@grs Just for making things easier when merging (and me updating the issue-746-... branch: how about changing this PR to merge it into the branch issue-746-plan-based-scheduling instead of master? History will be the same in the end, but it makes it easier for us to update the branches in parallel.

grs:
@lulf I don't know how to change an existing PR, but created a new one doing as you suggest: https://github.com/EnMasseProject/enmasse/pull/885

lulf:
@grs I rebases #885 against issue-746... (it seems to be created from master) to make sure it has both changes in it for the tests to pass. Closing this.

PR #893, opened 2018-02-16
Ansible systemtests scripts

kornys:


templates/install/ansible/playbooks/openshift/systemtests-dependecies.yml
```diff
@@ -0,0 +1,3 @@
+- hosts: localhost
+  roles:
+    - systemtests_dependencies
```
lulf: Can you just push this into the systemtests.yml playbook?

kornys: Sure that is not a problem.

systemtests/scripts/test_func.sh
```diff
@@ -28,6 +28,10 @@ function setup_test() {
     oc adm --config ${KUBEADM} policy add-cluster-role-to-user cluster-admin $OPENSHIFT_USER
     export OPENSHIFT_TOKEN=`oc whoami -t`
     ansible-playbook ${ENMASSE_DIR}/ansible/playbooks/openshift/systemtests.yml --extra-vars "namespace=${OPENSHIFT_PROJECT}"
+    if [ $? -ne 0 ]; then
+        echo "DEPLOYMENT FAILED - tests won't be executed."
+        exit 1
```
lulf: Maybe out this check into run_test_component? I think other abort conditions are defined there as well

kornys:
@lulf I think I can merge this with master. I removed unused stages from jenkins files (install clients, webdrivers, run Xvfb etc...) and now these steps are executed with your deploy openshift ansible. PR job in Jenkins looks works. Do you agree?

kornys:
@lulf I'm not sure if I understand correctly. You mean remove maven installation steps and trust that on system some maven is already installed? Or try to install maven by yum from rhscl or epel?

lulf:
@kornys To install it from yum

kornys:
@lulf I have to use apache maven > 3.5.0 due to artemis core client, when I try to compile cli-java by maven from yum, If fails.

Issue #897, opened 2018-02-19
Enforce valid name for addresses


lulf:
Address resources have a mandatory 'name' field and an optional 'address' field. If the address field is not specified, its value will be taken from the 'name' field.  The reason for having the 'name' and 'address' field separate is so that 'name' should be a valid k8s identifier. This is not enforced currently. 

From the user POV, only the 'address' field is really relevant. As a fix, we can

* Make 'address' mandatory and required 'name' to be empty in the REST API. This will make the API simpler, and avoid any conflict between names and addresses.
* Enforce that if both 'name' and 'address' is specified, the 'name' must match our sanitized version of 'address'.

grs:
Why move address from spec to metadata?

lulf:
@grs I guess the only reason is that it is a bit strange we have 'name' and 'addressSpace' in 'metadata' but 'address' in 'spec'. Also I think it just feels a bit strange to have 'empty' metadata :)

grs:
I see the spec as being the description of the address we want, and the address field is very much part of that.

The metadata would not be empty once the resource is created.

lulf:
@grs should we then move uuid and addressSpace to the spec?

grs:
For addressSpace, I could see either way. Fo uuid, I think that should *not* be in spec as it has nothing to do with how the address is realised. It is internal metadata, similar to the name (as distinct from address), so fits better in the metadata.

lulf:
@grs ok, I guess 'addressSpace' is mostly used as 'metadata' so it should belong there then. I've updated the initial bullet point to keep 'address' in spec.

lulf:
@grs How about also this change to the sanitizing logic in both console and restapi:  Instead of creating a qualifier based on a hash of the address, make the last part of the identifier a uuid and truncate the 'name' part to have the uuid string fit.

We would then use this identifier as the name of the configmap as well as the 'name' in the address. This would avoid the 'address-config-' prefix and would ensure that the configmap would not conflict with other addresses with a similar name. A configmap would therefore look like this:

```
apiVersion: v1
kind: ConfigMap
metadata:
  name: myqueue-264d1392-1619-11e8-afb8-507b9def37d9
data:
  config.json: |-
    {
      "apiVersion": "enmasse.io/v1",
      "kind": "Address",
      "metadata": {
        "name": "myqueue-264d1392-1619-11e8-afb8-507b9def37d9",
[...]

PR #905, opened 2018-02-19
Issue 897 address definition change

lulf:
This implements the changes proposed in #897

### DIFF
--- a/address-model-lib/src/main/java/io/enmasse/address/model/KubeUtil.java
+++ b/address-model-lib/src/main/java/io/enmasse/address/model/KubeUtil.java
@@ -9,7 +9,7 @@ package io.enmasse.address.model;
  */
 public class KubeUtil {
     public static String sanitizeName(String name) {
-        String clean = name.toLowerCase().replaceAll("[^a-z0-9]-", "");
+        String clean = name.toLowerCase().replaceAll("[^a-z0-9\\-]", "");
         if (clean.startsWith("-")) {
             clean = clean.replaceFirst("-", "1");
         }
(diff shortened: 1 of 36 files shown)


lulf:
@grs One thing that is missing here (because I'm not 100% sure how to change that) is to have the agent use UUIDs for the last part of the configmap name. By doing that, the delete_address must also be changed to pass the 'name' and no the 'address' so that it deletes the correct configmap.

lulf:
@grs Update: I've changed the agent to retrieve all address configmaps and delete the one matching the address for now.

agent/testlib/mock_resource_server.js
```diff
@@ -336,7 +336,7 @@ function get_config_map(name, type, key, content) {
 }
 
 ConfigMapServer.prototype.add_address_definition = function (def, name, annotations, status) {
-    var address = {kind: 'Address', metadata: {name: def.address}, spec:def};
+    var address = {kind: 'Address', metadata: {name: name}, spec:def};
```
grs: I don't think we need this change. If we want it I would do name: name || def.address as the name is not always specified.

address-model-lib/src/main/java/io/enmasse/address/model/KubeUtil.java
```diff
@@ -8,22 +8,33 @@
...
         if (clean.endsWith("-")) {
-            clean = clean.substring(0, clean.length() - 2) + "1";
+            clean = clean.substring(0, clean.length() - 1) + "1";
         }
         return clean;
     }
 
+    public static String sanitizeWithUuid(String name, String uuid) {
+        name = sanitizeName(name);
+        if (name.length() + uuid.length() > MAX_KUBE_NAME) {
```
grs: Is this missing '+1' to account for the separator '-'? (the chars reserved for pod id will mean it doesn't exceed an invalid name but it would restrict to 9 pods - probably still plenty for most cases of course).

lulf: @grs Good catch, thanks!

PR #911, opened 2018-02-20
Agent address naming

grs:


lulf:
@grs This build probably won't pass as it requires changes in #905 . I've rebased that against this so I can either change that PR to merge into this, or just merge #905 containing these fixes.

grs:
Either way is fine by me

PR #912, opened 2018-02-20
Systests plan tests

tkratky:


systemtests/src/test/java/io/enmasse/systemtest/PlansTest.java
```diff
@@ -0,0 +1,105 @@
...
+import org.slf4j.Logger;
+
+import java.util.*;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+
+import static org.junit.Assert.assertEquals;
+
+@Category(IsolatedAddressSpace.class)
+public class PlansTest extends TestBase {
```
kornys: Why this file is not in some package?

tkratky: this class is in io.enmasse.systemtest; package (it doesn't match any of currently created packages and it doesn't make sense for me to create new package only for this class. Maybe once we'll have more Plans related tests then we could create new package for them) do you agree?

lulf: Maybe call it PlansTestBase

lulf: In general, I think we should try to not rely too much on inheritance for this kind of things. It would be better if we moved to junit5 IMO to get better tools to create these kind of 'test fixtures'.

tkratky: but it's not 'Base' because each class with new plans should add @setup and @teardown because java doesn't support multi-inheritance and we would like to use it in WebConsole tests for example. Basically is it the same approach like is already used with SeleniumProvider in WebConsole tests.

PR #914, opened 2018-02-21
NEW test for colocatedTopics

tkratky:


tkratky:
It could be problem due to reporting test results and there is no class for TopicQueueTest :/ 
or we could move both tests into standard.SmokeTests potentially (SmokeTests will be never reported because there are tests with mixed Polarion components)

lulf:
Ok, lets leave it like this for now then, but we should start thinking about how to make tests 'share' addresses/brokers to speed things up

PR #930, opened 2018-02-22
UPD AddressApiClient - deleteAddresses use AddressResource name

tkratky:
- use Address resource name if was specified otherwise request this name

### DIFF
--- a/systemtests/src/main/java/io/enmasse/systemtest/Address.java
+++ b/systemtests/src/main/java/io/enmasse/systemtest/Address.java
@@ -6,18 +6,36 @@ package io.enmasse.systemtest;
 
 public class Address {
 
+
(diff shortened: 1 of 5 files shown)


tkratky:
failed in CI, I'm working on fix

tkratky:
First: thanks for the PR, I really like your solution. :+1: 

I didn't know that 'sanitize' is so easy. I disagree with the option for deploy, I would prefer set some boolean parameter on Destination (autoGenerate=false/true) because when we set Name within Destination and we don't use this name during deploy, then is impossible to determine which addresses were deployed with 'dest.Name' and which were deployed with autogenerated Name.

So if autoGenerate is true then we should request Name of resource via rest-api and put it into the delte/replace request. What do you think about that?

PR #937, opened 2018-02-23
NEW WebConsole test for dynamically created plans

tkratky:
- openWebConsolePage method modified to throw Exception when page is not loaded
- PlansTest moved into new plan package


tkratky:
I really don't like the solution with created IsolatedWebConsole tests but I wasn't able to found better way how to do that. Again as in "basic PlanTest" (via rest api) plansProvider is required, so setUp and tearDown method must implement plansProvider.setup/teardown. I hope we will able to do that better  during rewriting systemtests to junit5.

PR #966, opened 2018-02-28
UPD AddressApiClient - check response in all methods on 200

tkratky:
- refactored logging


tkratky:
I wasn't able to find out why request for delete address space wasn't resolved. Only thing what I found out was that in delete method wasn't check on response. So I added this check into the all methods. 
Maybe it was client-side issue. We'll see if it happen again in the future. :crossed_fingers:

PR #976, opened 2018-03-02
Variable substition for generating the templates for keycloak-plugin and subserv

vbusch:
When generating the templates:
- allow to specify the node command to use
- allow to specify the keycloak home dir
- allow to specify if you are adding extra script to keycloak-plugin

### DIFF
--- a/templates/include/auth-service.jsonnet
+++ b/templates/include/auth-service.jsonnet
@@ -178,9 +178,7 @@ local images = import "images.jsonnet";
       }
     },
 
-
-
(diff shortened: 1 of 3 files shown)


templates/include/auth-service.jsonnet
```diff
@@ -251,11 +253,16 @@ local images = import "images.jsonnet";
                     }
                   }
                 ],
+                [if add_keycloak_user=="true" then "command"]: [
+                  "/opt/eap/standalone/bin/start-keycloak.sh"
```
lulf: This should be keycloak_home

PR #978, opened 2018-03-02
Cleanup broker config after moving to stateful sets

lulf:


lulf:
@grs @rgodfrey  Pay attention to the runServer() function. I changed it to copy new configuration and certs based on the intuition that if the broker image is upgraded, it may need that configuration to work.

 The other part that I think should always be there is that the certificates are re-imported at each startup, in case they have been updated.

Issue #917, opened 2018-02-21
system-tests: change plans of topic addresses in systemtests to "pooled-topic"


tkratky:

tkratky:
system-tests were modified to use pooled-topics and pooled-queues

PR #994, opened 2018-03-05
Use pooled brokers by default in webconsole tests

lulf:
This prevents CI from creating many brokers that the test strictly doesn't need


tkratky:
btw related issue: https://github.com/EnMasseProject/enmasse/issues/917

PR #1003, opened 2018-03-05
ensure results count is updated when addresses change

grs:


lulf:
@tkratky @grs I was a bit eager on the merge button here, the change looked so innocent so I didn't check the CI :)

Issue #949, opened 2018-02-27
agent: Results count is not set correctly after refresh the page


tkratky:
Count of 'Results' in 'Addresses' section is not set correctly after refresh the page

automated reproducer:
```
Xvfb :10 -ac &
make webdriver_install
PATH=$PATH:$(pwd)/systemtests/web_driver DISPLAY=:10 make MULTITENANT=true SYSTEMTEST_ARGS='standard.web.FirefoxWebConsoleTest#testViewAddresses' systemtests
```
manual steps to reproduce:
1. create address space
2. create 1 address (via restapi or webconsole)
3. check count of Results on top of the page - should be 1 Results
4. refresh web page in browser and check the Results value again (now is 0)

before refresh:
![image](https://user-images.githubusercontent.com/17743200/36723782-83d3bcee-1bb1-11e8-880a-0f95570d254e.png)

after refresh:
![image](https://user-images.githubusercontent.com/17743200/36723798-8d06c2e8-1bb1-11e8-82ab-70fc9bead38a.png)

tkratky:
fixed as part of PR above

Issue #824, opened 2018-01-31
System-tests: February list of disabled tests


kornys:
1.  io.enmasse.systemtest.standard.mqtt.PublishTest#testPublishQoS2
 - bug: (QoS2 isn't implemented)
 - code: https://github.com/EnMasseProject/enmasse/blob/master/systemtests/src/test/java/io/enmasse/systemtest/standard/mqtt/PublishTest.java#L55

5. io.enmasse.systemtest.standard.QueueTest#testScaledown
 - bug: #851 
 - code: https://github.com/EnMasseProject/enmasse/blob/master/systemtests/src/test/java/io/enmasse/systemtest/standard/QueueTest.java#L175

9. io.enmasse.systemtest.standard.TopicTest#testTopicWildcards
 - bug: #105 (probably)
 - code: https://github.com/EnMasseProject/enmasse/blob/master/systemtests/src/test/java/io/enmasse/systemtest/standard/TopicTest.java#L209

10. io.enmasse.systemtest.standard.TopicTest#testDurableLinkRoutedSubscription
 - bug: (if I remember correctly TopicTests were disabled due to some issue with sending mqtt messages)
 - code: https://github.com/EnMasseProject/enmasse/blob/master/systemtests/src/test/java/io/enmasse/systemtest/standard/TopicTest.java#L230

11. io.enmasse.systemtest.standard.TopicTest#testDurableMessageRoutedSubscription
 - bug: (if I remember correctly TopicTests were disabled due to some issue with sending mqtt messages)
 - code: https://github.com/EnMasseProject/enmasse/blob/master/systemtests/src/test/java/io/enmasse/systemtest/standard/TopicTest.java#L273

13. io.enmasse.systemtest.brokered.SmokeTest#testCreateAlreadyExistingAddress
 - bug: #346 
 - code: https://github.com/EnMasseProject/enmasse/blob/master/systemtests/src/test/java/io/enmasse/systemtest/brokered/SmokeTest.java#L107

14. io.enmasse.systemtest.brokered.jms.TopicTest
 - bug: ENTMQBR-912
 - code: https://github.com/EnMasseProject/enmasse/blob/master/systemtests/src/test/java/io/enmasse/systemtest/brokered/jms/TopicTest.java#L118
- fix for that was merged: https://github.com/apache/activemq-artemis/pull/1678

15. io.enmasse.systemtest.brokered.TopicTest.test#TopicPubSubWildcards
[...]

PR #1006, opened 2018-03-06
ensure results count is updated when addresses change

grs:
fix for #949 (updated to check that filterconfig is initialised)


tkratky:
@grs could we enable web console tests that verify this fix within this PR? (No. 26. https://github.com/EnMasseProject/enmasse/issues/824)

grs:
@tkratky yes, can you confirm the tests? That issue has three, but two seem to be the same one (line 106 is not disabled). The second test is now marked as also failing due to another unopened issue so I have left that disabled but revised the comment.

tkratky:
@grs : enabled tests for brokered as well, so now everything looks fine, thank you!

grs:
@tkratky doh! sorry, got confused by the similar filename and tests!

PR #1008, opened 2018-03-06
Pooled status

lulf:


kornys:
Check if sharded addresses are ready works fine! Check if addresses are removed works fine as well!

Issue #429, opened 2017-10-30
'Message not accepted' after sending message to address which is created/removed in loop


tkratky:
### description: 
- when 2 addresses are deployed into default address space (1 message is send to first of them) and one by one are removed, then time to time is connection rejected with following message: _**Message not accepted (remote state: Rejected{error=Error{condition=failed, description='null', info=null}}) after 0 messages sent**_

### prerequisites:
- enmasse running in multitenant mode: `sh ./templates/install/deploy-openshift.sh -a "none standard" -o "multitenant"`
### steps to reproduce:
1. execute already existing test: (is required to uncomment `runQueueTest(client, address, 1);` line in this test)
`make MULTITENANT=true SYSTEMTEST_ARGS='standard.QueueTest#testCreateDeleteQueue' systemtests`
https://github.com/EnMasseProject/enmasse/blob/master/systemtests/src/test/java/io/enmasse/systemtest/standard/QueueTest.java#L129
2. output from test:
```
2108148 [main] INFO test - Test is running in multitenant mode
2108148 [main] INFO test - Following HTTP request will be used for getting address space:/v1/addressspaces/ with host: restapi-jenkins-enmasse-1308.172.16.84.22.xip.io with Port: 80
2108154 [vert.x-eventloop-thread-0] INFO test - Following list of address spaces received{"apiVersion":"enmasse.io/v1","kind":"AddressSpaceList","items":[{"metadata":{"name":"testspace","namespace":"testspace"},"spec":{"type":"standard","plan":"unlimited","endpoints":[{"name":"console","service":"console","host":"console-testspace.172.16.84.22.xip.io"},{"name":"messaging","service":"messaging","host":"messaging-testspace.172.16.84.22.xip.io","certProvider":{"name":"secret","secretName":"external-certs-messaging"}},{"name":"mqtt","service":"mqtt","host":"mqtt-testspace.172.16.84.22.xip.io","certProvider":{"name":"secret","secretName":"external-certs-mqtt"}}],"authenticationService":{"type":"standard","details":{}}},"status":{"isReady":true}}]}
2108155 [main] INFO test - Address space '{name=testspace,namespace=testspace}' already exists.
2108155 [main] INFO test - Waiting for 2 minutes before starting tests
2108159 [main] INFO test - Testing endpoint : keycloak-jenkins-enmasse-1308.172.16.84.22.xip.io:80
[...]

PR #1023, opened 2018-03-08
UPD QueueTest - uncommented simple send/receive from testCreateDelete

tkratky:


lulf:
See my comment in https://github.com/EnMasseProject/enmasse/issues/429

PR #1055, opened 2018-03-13
FIX setup-openshift - restart openshift cluster if didn't start properly

tkratky:


tkratky:
I saw this issue yesterday first time and it was on:
```
FAIL
   Error: cannot retrieve nodes
   Caused By:
     Error: couldn't get version/kind; json parse error: unexpected end of JSON input
error: A server URL must be specified
Now: 2018-03-13:13:07:45
```

and today 'oc cluster up' finished correctly but waiting until up failed on:
```
Notice: Docker registry is not yet ready
Notice: Router is not yet ready
Notice: Persistent volumes are not yet ready

Notice: 3 OpenShift component(s) are not yet ready (see above)
```

PR #1057, opened 2018-03-14
Upgrade keycloak to 3.4.3.Final

lulf:


calohmn:
@lulf we had a look into this Keycloak version and had to update the 
`keycloak-plugin/src/main/configuration/standalone.xml`
file for Keycloak to start - the original 
`standalone/configuration/standalone.xml`
file from the Keycloak 3.4.3 release zip has some changes compared to the 3.3.0 version.

lulf:
@calohmn Thanks, that saved us some time!

Issue #701, opened 2018-01-09
Keep AddressSpace and Address resources around until deletion


lulf:
When deleting address spaces and addresses, there is no way of knowing when the operation is completed.

To fix this, provide a new field in the Address and AddressSpace resource that indicates the lifecycle state of the resource in the same way this is done for Pods.

For pods, the possible values are explained here: https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/

The proposal is that we add a similar 'phase' field to the status of address and address space resources which can have the following values:

* Pending      - Resource has been created but it is not yet ready to be configured
* Configuring      - Resource has been created but component is not yet configured
* Active          - The infrastructure has been configured with the resource
* Failed           - The infrastructure has failed to configuration
* Terminating - The resource is marked for deletion and is awaiting infrastructure to be reconfigured. The resource will be deleted by the controller responsible for handling the Terminating state.

Using the same names as for Pods makes the intention of this field familiar to k8s users, but Running is replaced by Active.

The different operations have the following behavior when a resource already exists in one of the states:

* Create
   * Pending, Configuring, Active, Failed, Terminating: Already exists error
* Delete
   * Pending, Configuring, Active, Failed, Terminating: Initiates termination and sets phase to Terminating
* Replace
   * Pending, Configuring, Active, Failed, Terminating: Update resource and set state to Pending

lulf:
@grs Now that the console 'bypasses' the restapi, it also needs to set phase=Terminating and writeback the configmap instead of deleting them as in #892 .

Another thought is that for the brokered address space, there is no GC process like for the standard address space, so when the agent runs in the brokered address space, it needs to look at addresses with phase == Terminating, delete them from the broker (as well as those it cant find a configmap for), and then delete the configmap.

lulf:
We decided to leave this for 1.0, where all resources have finalizers to ensure deletion is complete.

PR #1066, opened 2018-03-16
Terminate old addresses through garbage collection

lulf:
* Reuse status checks when terminating
* Ensure pooled addresses can handle status check success on termination

Issue #701


lulf:
@grs Would it be much work to have the console set address phase to Terminating instead of deleting the configmap when address is deleted?

grs:
No, I don't think so. Which component deletes them?

lulf:
@grs Currently the standard controller, since it has the status checks to verify that router has been de-configured, broker deleted etc. It could be moved to agent if we move the status checks there.

lulf:
@grs If deleting the configmap and removing the address from the broker is done at the same time, theres not much point. If they are not, the configmap could be kept until the address is deleted from the broker.

PR #1067, opened 2018-03-16
Check keycloak realm existence every 30 seconds by default

lulf:
This fixes an issue where the controller would receive all address
space creation events before keycloak was fully running and accepting
connections. The update to keycloak would then get lost. Instead, introduce a more frequent interval in which keycloak is checked against the latest known set of address spaces.

@grs @kornys I believe this will the 'unable to find realm within timeout' failures we've seen in the systemtests.


kornys:
@lulf I looked into jenkins for results, Timeout exception 'unable to find realm within timeout' is still present.

lulf:
@kornys Fixed in https://github.com/EnMasseProject/enmasse/pull/1071 so I'm merging this anyway even though it is not a direct cause of the issue, I think it is needed anywa.

PR #1071, opened 2018-03-19
Pass the authentication service in the request so that the expected r…

lulf:
…ealm is created

I'm 99% sure this is the cause of the issue. All of the tests that fail (the plans test and in my PR) are using this constructor, which ended up creating address spaces with the "None" authentication service. The keycloak controller filters out any address spaces that does not use the standard authentication servicve type so it never creates the realm.


kornys:
@lulf ok I going to check it.

PR #1076, opened 2018-03-20
FIX restart docker service when not running

kornys:


systemtests/scripts/setup-openshift.sh
```diff
@@ -13,6 +13,13 @@ sudo rm -rf /var/lib/origin/openshift.local.pv
 sudo rm -rf /var/log/containers/*
 sudo rm -rf /var/log/pods/*
 
+DOCKER_STATUS=$(sudo systemctl show --property ActiveState docker-latest | sed -n -e 's/^ActiveState=//p')
+if [ $DOCKER_STATUS != "active" ]; then
+    echo "Docker service is not running"
+    echo "Starting docker service"
+    sudo systemctl restart docker-latest
```
lulf: 'docker-latest' should be taken from the env $DOCKER  so that it can work with just 'docker'.

kornys: Thanks!

PR #1085, opened 2018-03-22
Use reduced set of privileges for service account

### LINKED ISSUE #1072: Reduce privileges needed on OpenShift
With the new kubernetes-client version 3.1.8, we can use impersonation when creating projects. This means we can probably get rid of some of the current privileges required around impersonation and rolebindingrestrictions,  as long as the address-controller is allowed to impersonate the user who creates the address space.

lulf:
This reduces the scope of cluster admin needed for the enmasse-admin SA
to:

* Create projectrequests
* Create tokenreviews

The use of impersonation is removed.

This fixes #1072


lulf:
Another misunderstanding on my part, #1080  is actually the solution we want.

PR #1090, opened 2018-03-22
Issue 421 hostnames

tkratky:


tkratky:
@lulf do you have any idea why host in endpoints for address spaces is not set in travis ? It works fine locally and in Jenkins. Maybe some different configuration for Travis machines?
```
21:53:58,925 [main] INFO io.enmasse.systemtest.bases.TestBase - Address-space successfully created: {name=brokered-shared-1,namespace=brokered-shared-1,type=brokered,plan=unlimited-brokered,endpoint-console={host=null,port=32655},endpoint-messaging={host=null,port=31998}}
```

lulf:
@tkratky This is because on kubernetes, routes does not exist. And instead, loadbalancer service types are used instead. However, those do not generate a hostname to use, so hostname will be null. I think that if hostname is null, the rule should be to lookup the service ip of the service field in the endpoint entry and use that.

systemtests/src/test/java/io/enmasse/systemtest/bases/TestBase.java
```diff
@@ -164,14 +174,24 @@ protected void createAddressSpace(AddressSpace addressSpace, boolean extraWait)
                 Thread.sleep(120_000);
             }
         } else {
+            addrSpaceResponse = TestUtils.getAddressSpaceObject(addressApiClient, addressSpace.getName());
             log.info("Address space '" + addressSpace + "' already exists.");
         }
+        if (addressSpace.getEndpoints().isEmpty()) {
+            addressSpace.setEndpoints(addrSpaceResponse.getEndpoints());
+            log.info(String.format("Address-space '%s' endpoints successfully set.", addressSpace.getName()));
```
kornys: you can use log.info("foo {} bar", variable) instead of String.format

systemtests/src/test/java/io/enmasse/systemtest/bases/TestBase.java
```diff
@@ -164,14 +174,24 @@ protected void createAddressSpace(AddressSpace addressSpace, boolean extraWait)
...
+        }
+        log.info(String.format("Address-space successfully created: %s", addressSpace));
```
kornys: Same as above

PR #1093, opened 2018-03-23
Upgrade to Artemis 2.5.0

lulf:


lulf:
@grs Please review the latest commit. It appears that the 'to' property of the message is now changed in Artemis. A new message annotation was added containing the original address. I've modified the mqtt-gateway to look for that annotation. I'm not sure if the change in Artemis is intended or a bug though.

grs:
@lulf it looks like there is already similar logic that was looking at an application property. So perhaps the change is that artemis now uses a message-annotation instead.

PR #1109, opened 2018-03-27
Console oauth support

grs:
This will not yet pass as tests need to be updated. I have kept the chnages as three separate commits for clarity. These can be separated into their own PRs if we wish, I kept them together for ease of evaluation for now.

It also relies on fixes in https://github.com/EnMasseProject/enmasse/pull/1099 plus one small addition I will push as a PR to that branch.


kornys:
@grs may I update webconsole tests in this branch?

grs:
@kornys sure, that would be much appreciated!

grs:
Committed through https://github.com/EnMasseProject/enmasse/pull/1119

PR #1110, opened 2018-03-27
derive base uri from service host if set

grs:
@rgodfrey, though a little ugly this seems to me the simplest solution to getting the url set correctly


lulf:
@grs @rgodfrey will this work with an external auth service that uses the plugin or an auth service that has to be accessed through an external route? In the latter case, the host in this env will not be accessible by the console

grs:
No it won't. But an external auth service would be configured outside of enmasse I think anyway, and there is an env var with which the full url can be specified to the plugin.

Issue #973, opened 2018-03-01
Configure per-address settings for each address created on broker


lulf:
Change agent to configure address-settings for each queue it creates and set max-size-bytes as (global-max-size * credit) - C (where C is the constant giving us headroom for management)

See https://github.com/EnMasseProject/enmasse/blob/master/documentation/design_docs/design/broker-limits.adoc

PR #1111, opened 2018-03-28
Set per-address address settings to limit max size

grs:
(This requires using a connector per address, created on the broker after the queue is created,
in order to ensure the autolinks do not fire before address settings and queue have been created
explicitly, as triggering the links will cause implicit creation of queue)

### DIFF
--- a/agent/lib/utils.js
+++ b/agent/lib/utils.js
@@ -173,7 +173,7 @@ module.exports.changes = function (last, current, compare, unchanged, stringify)
             added: current,
             removed: [],
             modified: [],
-            decription: util.format('initial %s', description(current))
+            description: util.format('initial %s', description(current))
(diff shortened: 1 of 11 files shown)


grs:
Closes #973

agent/lib/utils.js
```diff
@@ -223,3 +223,26 @@ module.exports.changes = function (last, current, compare, unchanged, stringify)
         }
     }
 };
+
+const bytes_pattern = /^(\d*)\s*(KB|MB|GB)?$/i;
+module.exports.parse_bytes = function (s) {
```
lulf: I can't see this being used anywhere, is it still needed?

grs: It is not, no.

Issue #1113, opened 2018-03-28
Unable to login to webconsole using OAuth


kornys:
When I try to open webconsle from different server on same subnet as server running the openshift webbrowser cannot reach the keycloak because it tries to redirect me to service IP.

Env:
- Server running openshift with enmasse
- WebBrowser running on different system in same subnet

Steps to reproduce:
1. Deploy enmasse and create address space
2. Try open and login webconsole using webbrowser

kornys:
https://github.com/EnMasseProject/enmasse/pull/1109

grs:
oauth is not yet supported in the address-controller for an external auth service either. In that case I guess it would be a new field in the address-space definition. That same field could perhaps be optionally used to provide the value even for the 'standard' auth service type, with the current cobbling together merely a fallback (to avoid having to specify it). Possibly it would even be cleaner to derive the default url if not specified somewhere else (since its the address-controller that sets up the route anyway). I did not have the big picture for this component, so made a 'pragmatic'(!) hack to get it working.

PR #1121, opened 2018-04-03
Allow oauthURL to be overridden for standard authentication service

lulf:


grs:
So with this, solution for https://github.com/EnMasseProject/enmasse/issues/1113 is to explicitly specify the oauth url as a field of the address space when it is created through the rest api, right?

PR #1122, opened 2018-04-03
Configure standard authentication service using configmap

lulf:


documentation/service_admin/installing-openshift.adoc
```diff
@@ -83,7 +83,14 @@ oc create secret generic keycloak-credentials --from-literal=admin.username=admi
...
 ----
-oc adm policy add-role-to-user view system:serviceaccount:$(oc project -q):default
+oc adm policy add-role-to-user view system:serviceaccount:enmasse:default
+----
+
+. Create keycloak configuration used by controller and service
++
+[options="nowrap"]
+----
+oc create configmap keycloak-config --from-literal=hostname=standard-authservice --from-literal=amqpPort=5671 --from-literal=httpPort=8443 --from-literal=caSecretName=standard-authservice-cert --from-literal=oauthURL=https://keycloak-enmasse.127.0.0.1.nip.io:443/auth
```
grs: I don't think this will work, as 'keycloak-enmasse.127.0.0.1.nip.io' will resolve to 127.0.0.1 in the admin pod.

documentation/service_admin/installing-openshift.adoc
```diff
@@ -139,7 +147,7 @@ openssl req -new -x509 -batch -nodes -days 11000 -subj "/O=io.enmasse/CN=address
 +
 [options="nowrap"]
 ----
-oc create secret tls address-controller-cert --cert=address-controller-cert/tls.crt --key=address-controller-cert/tls.key
+oc create secret tls address-controller-cert jn enmasse --cert=address-controller-cert/tls.crt --key=address-controller-cert/tls.key
```
grs: Is the 'jn' intentional?

templates/install/ansible/roles/standard_authservice/defaults/main.yml
```diff
@@ -0,0 +1,2 @@
+---
+route_suffix: 127.0.0.1.nip.io
```
grs: As above, I don't think the 127.0.0.1 based host will work correctly.

templates/install/deploy-openshift.sh
```diff
@@ -178,6 +178,7 @@ do
         KEYCLOAK_PASSWORD=`random_string`
         create_self_signed_cert "oc" "standard-authservice.${NAMESPACE}.svc.cluster.local" "standard-authservice-cert"
         runcmd "oc create secret generic keycloak-credentials --from-literal=admin.username=admin --from-literal=admin.password=$KEYCLOAK_PASSWORD" "Create secret with keycloak admin credentials"
+        runcmd "oc create configmap keycloak-config --from-literal=hostname=standard-authservice --from-literal=amqpPort=5671 --from-literal=httpPort=8443 --from-literal=caSecretName=standard-authservice-cert --from-literal=oauthURL=https://keycloak-${NAMESPACE}.127.0.0.1.nip.io:443/auth" "Create standard authentication service configuration"
```
grs: If we create the configmap after running the template, we can get the cluster ip for the service and use that to set the hostname. Do you see any issue with that?

documentation/service_admin/installing-kubernetes.adoc
```diff
@@ -88,6 +88,22 @@ kubectl create secret generic keycloak-credentials --from-literal=admin.username
 kubectl apply -f ./kubernetes/addons/standard-authservice.yaml
 ----
 
+. Create keycloak configuration used by controller and service. For the messaging console and
+keycloak controller to work, you need to specify the `httpUrl` setting. To get the service ip:
```
grs: Not a blocker, but to me it would be clearer to describe this as being for oauth (if it is not set, the console can still use basic auth). That also makes it clearer why the auth path is there.

keycloak-controller/src/main/java/io/enmasse/keycloak/controller/KeycloakParams.java
```diff
@@ -8,41 +8,33 @@
...
 import java.util.Map;
-import java.util.Optional;
 
 public class KeycloakParams {
 
     private static final Logger log = LoggerFactory.getLogger(KeycloakParams.class);
 
-    private final String host;
-    private final int httpPort;
+    private final String keycloakUri;
```
grs: Again, not a blocker, but it seems like the same piece of data is called different things in various places, httpUrl, keycloakUri, oauthUrl... might be clearer to just use the same throughout?

documentation/service_admin/installing-kubernetes.adoc
lulf: @grs It is also used by keycloak-controller to connect to keycloak, so not only oauth (which is why the parameter name was changed).

PR #1137, opened 2018-04-11
Websocket tests

kornys:


systemtests/src/main/resources/rhea.html
```diff
@@ -0,0 +1,47 @@
...
+                append('received: ' + context.message.body);
+                if (current++ < requests) {
+                    next_request();
+                } else {
+                    context.connection.close();
+                }
+            });
+            var ws = client.websocket_connect(WebSocket);
+            var connection = client.connect({"connection_details":ws(server, ["binary", "AMQPWSB10", "amqp"]), transport: 'tls',
+                username: username, password: password, "reconnect":false, rejectUnauthorized: false});
```
grs: Minor point: the rejectUnauthorized won't do anything here.

kornys: Ah, good point, thank you!

PR #1139, opened 2018-04-11
Lots of keycloak improvements

lulf:


lulf:
@rgodfrey So this PR evolved a bit :)

Keycloak now trusts the OpenShift CA in the keycloak config, so identity brokering works. I've also modified the keycloak-controller to automatically add the identity provider if it is configured, and also create the user, groups and have the admin user join the 'admin' and 'manage' groups automatically. Still need to do more testing, but the main things are there now.

What remains is to modify the ansible scripts to provide the 'oauth client' credentials from a service account instead (where you can use the token of the service account instead of having to generate a password).

PR #1150, opened 2018-04-18
NEW TestTag class with predefined test tags

tkratky:
This functionality could help to avoid typos in @ Tags


systemtests/src/main/java/io/enmasse/systemtest/TestTag.java
```diff
@@ -0,0 +1,7 @@
+package io.enmasse.systemtest;
```
kornys: you have to add license string here

PR #1151, opened 2018-04-18
Refactor without mqtt tests

kornys:


systemtests/src/main/java/io/enmasse/systemtest/ability/ITestBaseWithoutMqqt.java
```diff
@@ -0,0 +1,19 @@
+/*
+ * Copyright 2018, EnMasse authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+package io.enmasse.systemtest.ability;
+
+import io.enmasse.systemtest.AddressType;
+
+public interface ITestBaseWithoutMqqt extends ITestBase {
```
lulf: Mqqt -> Mqtt

kornys: @lulf good catch, thank you!

PR #1152, opened 2018-04-18
Extend selenium interface

kornys:


systemtests/src/test/java/io/enmasse/systemtest/bases/web/WebConsoleTest.java
```diff
@@ -54,9 +53,6 @@ public void setUpWebConsoleTests() throws Exception {
 
     @AfterEach
     public void tearDownWebConsoleTests(ExtensionContext context) throws Exception {
```
tkratky: you can probably remove DI of ExtensionContext

PR #1162, opened 2018-04-20
Unify OpenShift and Kubernetes resources

lulf:
* Remove separation and duplication of resources
* Remove use of jsonnet for address space definitions

@grs You had a concern previously about ability to edit the configmaps. If you look at the templates/install/resources/address-controller/address-space-definitions.yaml  configmap, everything is now in the yaml format so it is a bit better. Do you think it is ok?

This is a gradual step away from using jsonnet in the magic way we've done so far. It also makes it a bit easier to do upgrades for the global components.


grs:
Yes, I think its fine and makes sense.

templates/install/deploy.sh
```diff
@@ -0,0 +1,238 @@
...
+            echo "optional arguments:"
+            echo "  -h                   show this help message"
+            echo "  -a \"none standard\" Deploy given authentication services (default: \"none\")"
+            echo "  -d                   create an all-in-one cluster on localhost"
+            echo "  -e                   Environment label for this EnMasse deployment"
+            echo "  -n NAMESPACE         Project name to install EnMasse into (default: $DEFAULT_NAMESPACE)"
+            echo "  -m MASTER            Master URI to login against (default: https://localhost:8443)"
+            echo "  -o mode              Deploy in given mode, 'singletenant' or 'multitenant'.  (default: \"singletenant\")"
+            echo "  -u USER              User to run commands as (default: $DEFAULT_USER)"
+            echo "  -t CLUSTER_TYPE      Type of cluster, 'openshift' or 'kubernetes' (default: openshift) User to run commands as (default: $DEFAULT_USER)"
```
calohmn: ` User to run commands as (default: $DEFAULT_USER)` is to be removed here.

calohmn:
Testing with `./deploy.sh -t kubernetes` in minikube, the AddressController liveness/readiness probes fail. Error in log:
````
2018-04-24 09:37:04 INFO  DefaultExceptionMapper:43 - Returning client error HTTP status 404: RESTEASY003210: Could not find resource for full path: https://172.17.0.4:8081/healthz
````
Another error:
````
2018-04-24 09:37:04 WARN  ControllerChain:123 - Error syncing address space default
io.fabric8.kubernetes.client.KubernetesClientException: An error has occurred.
	at io.fabric8.kubernetes.client.KubernetesClientException.launderThrowable(KubernetesClientException.java:62)
	at io.fabric8.kubernetes.client.KubernetesClientException.launderThrowable(KubernetesClientException.java:53)
	at io.fabric8.kubernetes.client.dsl.base.BaseOperation.load(BaseOperation.java:330)
	at io.fabric8.kubernetes.client.dsl.base.BaseOperation.load(BaseOperation.java:70)
	at io.enmasse.controller.common.KubernetesHelper.processTemplate(KubernetesHelper.java:127)
	at io.enmasse.controller.TemplateInfraResourceFactory.createResourceList(TemplateInfraResourceFactory.java:126)
	at io.enmasse.controller.CreateController.handle(CreateController.java:66)
	at io.enmasse.controller.ControllerChain.onUpdate(ControllerChain.java:118)
	at io.enmasse.k8s.api.ResourceChecker.doWork(ResourceChecker.java:48)
	at io.enmasse.k8s.api.ResourceChecker.run(ResourceChecker.java:40)
	at java.lang.Thread.run(Thread.java:748)
Caused by: java.io.FileNotFoundException: /address-space-definitions/standard-space-infra.json (No such file or directory)
	at java.io.FileInputStream.open0(Native Method)
	at java.io.FileInputStream.open(FileInputStream.java:195)
	at java.io.FileInputStream.<init>(FileInputStream.java:138)
	at io.fabric8.kubernetes.client.dsl.base.BaseOperation.load(BaseOperation.java:327)
	... 8 more
````

**Comment by lulf** [comment]
@calohmn From those logs it looks like its running an outdated address-controller image. The code is changed to look for '.yaml' files (so should at least fail finding standar-space-infra-.yaml), and the health check URL was moved to /healthz.

**Comment by calohmn** [comment]
@lulf oh, yes. I've built images locally using the "latest" tag and forgot, that with such a tag, Kubernetes will pull the image instead of using the one that is present locally.

**Comment by calohmn** [comment]
@lulf A further thought about the "templates/include/address-space-definitions.yaml" file:
````
apiVersion: v1
kind: ConfigMap
metadata:
  labels:
    app: enmasse
  name: address-space-definitions
data:
  standard-space-infra-without-mqtt.yaml: ...
  standard-space-infra.yaml: ...
  brokered-space-infra.yaml: ...
````
I think it could make sense to have the contained yaml files as separate files in the "include" folder. That would make it easier to compare these files with each other, keeping them in sync, and would also allow using yaml-editor features like collapsing/hiding certain parts to make navigation inside the files easier.

The release "address-space-definitions.yaml" file could then be created via the Makefile.
(_Or_ the "address-space-definitions" ConfigMap could be created via
````
kubectl create configmap address-space-definitions --namespace $NS --from-file=standard-space-infra-without-mqtt.yaml --from-file=standard-space-infra.yaml --from-file=brokered-space-infra.yaml
````
from the deploy.sh, omitting the rather large "address-space-definitions.yaml".)

**Comment by lulf** [comment]
@calohmn Good suggestion, I like it!

I was also thinking that eventually we can decompose the contents of those '*space-infra*yaml" files so that we would have resources/qdrouterd/deployment.yaml etc. at the "top level". The idea being the same that it would simplify upgrades of an address space (and also reduce duplication). One way to configure those components could be to create an 'address-space-config' containing the parameters now mostly used by the templates.

File: templates/include/address-controller.jsonnet [hunk 2]
```diff
@@ -23,84 +43,208 @@ local common = import "common.jsonnet";
...
+                      "optional": true
                     }
+                  }
                 },
-                "ports": [
-                  common.container_port("https", 8081),
-                  common.container_port("http", 8080)
-                ],
-               "livenessProbe": common.http_probe("https", "/apis/enmasse.io/v1/health", "HTTPS", 30),
-                "readinessProbe": common.http_probe("https", "/apis/enmasse.io/v1/health", "HTTPS", 30),
```
calohmn: @lulf The last param value (initialDelaySeconds: 30) is now not contained in the livenessProbe and readinessProbe definitions any more. Was that change intentional?

lulf: Sort off. The plan was to see how it behaved, and I've not seen any issues with using the defaults yet, so I was thinking maybe to keep it.

calohmn: ok

PR #1179, opened 2018-04-27
Travis docker logs

tkratky:


tkratky:
Now there should be all required logs in tar: https://dl.bintray.com/enmasse/snapshots/6876/ 
Thank you for help @lulf !

PR #1180, opened 2018-04-27
Fix address controller deployment order

calohmn:
Ensure that `address-space-definitions.yaml` is applied before deploying the address controller.

This prevents errors in address controller about the ConfigMap `address-space-definitions` not being found.


lulf:
Thanks!

PR #1183, opened 2018-04-30
Make deployment script work if only 'oc' is available

lulf:


kornys:
@lulf I updated wait until up also, WDYT?

lulf:
@kornys Looks good, thanks!

kornys:
@lulf do you think that we can merge this?

PR #1210, opened 2018-05-07
Remove configuration of address settings

grs:
This does not scale well with the number of queues on a broker. Alternative approaches seem overly complex in comparison to the derived benefit.


grs:
@tkratky @kornys I have disabled the test for this, we can re-enable if/when we add back the functionality: https://github.com/EnMasseProject/enmasse/pull/1210/commits/5a8638dc4420d7839ac280bf6ec32574414e641f#diff-a2c66b163ce4927345876d307bf95765

PR #1218, opened 2018-05-12
Remove per-address space type service and replace with a single EnMas…

lulf:
@rgodfrey Wdyt, having a single service entry, and instead have plans for both brokered and standard. After talking to OSB experts, it seems that nesting multiple levels will not be possible.

### DIFF
--- a/service-broker/src/main/java/io/enmasse/osb/api/ServiceMapping.java
+++ b/service-broker/src/main/java/io/enmasse/osb/api/ServiceMapping.java
@@ -11,22 +11,26 @@ import io.enmasse.address.model.AddressSpacePlan;
 import io.enmasse.address.model.AddressSpaceType;
 import io.enmasse.address.model.Schema;
 import io.enmasse.osb.api.catalog.*;
+import org.apache.commons.codec.Charsets;
 
 import java.nio.charset.StandardCharsets;
 import java.text.MessageFormat;
 import java.util.*;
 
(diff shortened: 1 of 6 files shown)


rgodfrey:
So I see a couple of problems with the single service idea

1. As a user I'm going to see a lot of plans for things which are
essentially quite different
2. We want to enable switching between plans (e.g. from XS to S to M and
back), which should be possible within a single address space type; but
it's never going to be possible to switch between a brokered and a standard
(or a Kafka address space or whatever the future holds).

If OSB were richer in terms of nesting, or more configurable in terms of
saying which plans can be switched between then maybe it would make
sense... but I'm not really sold on this idea - what do you see as the main
benefits?

lulf:
@rgodfrey All good points, and I hadn't really thought about the aspect of switching plans, which I agree will be problematic with this approach. 

I think the main benefit is that the catalog UI would expose only a single service, but reading your comment I think it is a very minor benefit compared to the switching plans.

rgodfrey:
That's my view ... if we are tagging all the services with a common name
"such as EnMasse" the services should be fairly easy to find anyway.  l can
see the utility of a single service with different plans for the different
address space types where there is actually only one plan per address space
type (and so switching wouldn't be possible)... I'm wary of trying to make
the service broker instance sufficiently configurable to allow this to be
an option though (at that point I think it's probably a different service
broker implementation)

PR #1240, opened 2018-05-21
NEW retry mechanism for docker push

tkratky:


tkratky:
@lulf looks like retry mechanism works fine in our CI, in last two runs (# 10, # 11) it used repeat mechanism for standard-controller and topic-forwarder image. 
I'm not sure with location of the docker_push script. Any suggestions where could be it placed?
Thanks!

lulf:
@tkratky I don't think it belong in 'templates' as it is really just stuff related to templates that should be there.

Perhaps we should rename the toplevel folder 'release-scripts' to 'scripts' and put it there?

kornys:
good job, I would like to merge it!

PR #1241, opened 2018-05-21
Fix API version and docs misrpints

dmvolod:
It woluld be good to generate swagger.json file automatically and have SwaggerUI option for this.
I will try to take care about this in the future.


lulf:
Thanks!

Issue #1197, opened 2018-05-02
deploy.sh: Use alternative to clusterIP for Keycloak httpUrl


calohmn:
When I run the `deploy.sh` to deploy EnMasse on my Minikube cluster (with authentication "standard"), the Keycloak `httpUrl` gets set to the ClusterIP from the Keycloak service.
This IP however isn't reachable from the parent host, which causes the EnMasse console login to not work.
I wonder in which case usage of the ClusterIP here does indeed work?

I guess, finding a mechanism to return the correct `httpUrl` for all possible scenarios is quite difficult.
It could be the Minikube/Minishift IP+NodePort or the Loadbalancer hostname (`.status.loadBalancer.ingress.hostname`) or even something like `.metadata.annotations."dns.alpha.kubernetes.io/external"` if the service was annotated in such a way.

To make it work in the "let's get started with EnMasse and use Minikube/Minishift" case, I would suggest adding a detection for Minikube/Minishift in the `deploy.sh` und use Minikube/Minishift IP+NodePort in that case. I could create a PR for this.

lulf:
The approach used in the ansible playbook is that you can configure the keycloak_http_url setting. I think for more advanced installations we should have ansible work for kubernetes. However, having the deploy.sh is useful for the local deploy/development use case. 

Adding minikube/minishift autodetection is a good approach IMO.

lulf:
@calohmn Ok to close? Or would you like to keep this around in search for another solution?

calohmn:
@lulf yes, closing it. There can be a new ticket for another solution.

PR #1244, opened 2018-05-22
Add optional oauthUrl entry in keycloak-config (#1197)

calohmn:
This config entry is mapped to AUTHENTICATION_SERVICE_OAUTH_URL of the agent config. By making it optional, OAuth can be disabled for the agent console login. This is used for the deploy.sh Kubernetes deployment where the external keycloak URL cannot be determined in all cases.

### DIFF
--- a/address-space-controller/src/main/java/io/enmasse/controller/Main.java
+++ b/address-space-controller/src/main/java/io/enmasse/controller/Main.java
@@ -104,7 +104,7 @@ public class Main extends AbstractVerticle {
                 resolverMap.put(AuthenticationServiceType.STANDARD, new StandardAuthenticationServiceResolver(
                         config.getData().get("hostname"),
                         Integer.parseInt(config.getData().get("port")),
-                        config.getData().get("httpUrl"),
+                        config.getData().get("oauthUrl"),
                         config.getData().get("caSecretName")));
             } else {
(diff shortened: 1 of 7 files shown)


lulf:
One alternative would be to explicitly enable/disable oauth in the keycloak config, which I think might be better than configuring identical urls, wdyt?

Note: This PR has quite a few conflicts with #1246 on the docs part, where the install procedure has been somewhat merged in the source. I'd like to have both merged, i think maybe the easiest is to merge this one last.

calohmn:
I first thought, being able to use a separate `oauthUrl` could have benefits in some cases (being able to use the cluster-internal URL for the "keycloak-controller to keycloak" communication without having to use an external URL that could take a while to become available upon deployment (concerning DNS entries, AWS resources, etc.)).
But then there's the disadvantages of added config complexity and having to create certs with both host names.

So, yes, using a flag to enable/disable oauth looks like the better alternative (I would use an optional "oauthDisabled" flag, as I guess usage of OAuth should be the default).
I'll create a new PR with that - once PR #1246 is merged.

lulf:
Good points, and I agree with you on the added complexity. Maybe there is a more elegant way we can support internal + external urls in the future. For now, I think adding a oauthDisabled flag is ok.

PR #1245, opened 2018-05-22
Add autogenerated files and modules to .gitignore

dmvolod:
The git tree is not clear after the build


.gitignore
```diff
@@ -14,4 +14,9 @@
...
 
 # VS Code
-.vscode
\ No newline at end of file
+.vscode
+
+#Autogenerated build files
+dependency-reduced-pom.xml
+velocity.log
+router-metrics/*.tgz
```
lulf: Maybe this should be in router-metrics/.gitignore ?

dmvolod: Thanks for feedback.
Look like need to fix Makefile and Dockerfile as router-metrics-$version.tgz created in root folder instead of /build. I will do this.

dmvolod:
@lulf , please look at fix. Changed some build and docker files.

lulf:
@dmvolod Looks good, but would like to see CI pass before merging.

PR #1246, opened 2018-05-22
Update swagger spec and docs

lulf:
@rgodfrey This adds docs for service broker, api server, custom resources etc. It also updates the swagger spec and adds a REST API reference in the appendix of the docs.

Finally it also converts the ansible playbooks to use inventory files as they should be.


lulf:
A rendered version: http://enmasse.io/documentation/update-swagger-spec-and-docs/

PR #1256, opened 2018-05-23
Improve usage output of deployment script

ctron:
This change shows an explicit list of options on cases where options are available and also always falls back to showing the usage help when an argument is unknown.

### DIFF
--- a/templates/install/deploy.sh
+++ b/templates/install/deploy.sh
@@ -66,26 +66,25 @@ while getopts a:de:gm:n:o:u:t:yvh opt; do
         v)
             set -x
             ;;
-        h)
+        *)
             echo "usage: deploy.sh [options]"
             echo
             echo "deploy the EnMasse suite into a running Kubernetes cluster"
             echo
             echo "optional arguments:"
             echo "  -h                   show this help message"
-            echo "  -a \"none standard\"   Deploy given authentication services (default: \"none\")"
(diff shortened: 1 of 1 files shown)


templates/install/deploy.sh
```diff
@@ -66,28 +66,27 @@ while getopts a:de:gm:n:o:u:t:yvh opt; do
...
+        *)
             echo "usage: deploy.sh [options]"
             echo
             echo "deploy the EnMasse suite into a running Kubernetes cluster"
             echo
             echo "optional arguments:"
             echo "  -h                   show this help message"
-            echo "  -a \"none standard\"   Deploy given authentication services (default: \"none\")"
+            echo "  -a TYPE              Deploy given authentication services (default: \"none\")"
+            echo "                         TYPE := { none | standard }"
```
lulf: It is possible to deploy both, should there by a different syntax for that?

ctron: Oh I see .. yes. You are right I will change that.

ctron:
So this is how e.g. `ss` does it:
~~~
   -A, --query=QUERY, --socket=QUERY
       QUERY := {all|inet|tcp|udp|raw|unix|unix_dgram|unix_stream|unix_seqpacket|packet|netlink|vsock_stream|vsock_dgram}[,QUERY]
~~~

However that requires a comma separated list. That is why I needed to change the for loop in addition.

lulf:
LGTM!

PR #1258, opened 2018-05-23
UPD tests for new bindings via OSBAPI

tkratky:


tkratky:
@lulf thanks for review, fixed!

PR #1264, opened 2018-05-24
Docs: Add info about setting default k8s ns; fix typos/formatting

calohmn:
Kubernetes documentation: Add info on how to set the enmasse namespace as the default namespace after having created the namespace. This is needed as the following kubectl commands don't have the `--namespace` param.


lulf:
Thanks!

PR #1284, opened 2018-05-29
the definition of equals (and hashCode) on AddressSpace should include namespace

### LINKED ISSUE #1283: Errors when attempting to create two addresses with the same name in different namespaces
When attempting to create two addressesspaces with the same "name" but in different namesspaces it appears to suceed; but attempting to create an address in the second addressspace fails.  This is because the second address space is never actually created.

rgodfrey:
this closes #1283

### DIFF
--- a/address-model-lib/src/main/java/io/enmasse/address/model/AddressSpace.java
+++ b/address-model-lib/src/main/java/io/enmasse/address/model/AddressSpace.java
@@ -105,12 +105,12 @@ public class AddressSpace {
 
         AddressSpace that = (AddressSpace) o;
 
-        return name.equals(that.name);
+        return Objects.equals(name, that.name) && Objects.equals(namespace, that.namespace);
     }
 
     @Override
     public int hashCode() {
-        return name.hashCode();
+        return Objects.hash(name, namespace);
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


rgodfrey:
@lulf it seemed a little over the top to create a unit test just for equals and hashCode when, if those methods are ever changed then the test would likely need to be changed anyway ... however I can do if you like.  Obviously if we were doing TDD and someone had written a test *before* the original commit, that might have made sense :-)

lulf:
@rgodfrey I agree the change itself may be trivial. But there was a bug there, so I was thinking a unit test would at least make people think twice before changing hashcode/equals. In general I was thinking it would be a good start for other tests for the AddressSpace class. I'm fine with merging this as is though.

rgodfrey:
@lulf I'll probably get round to writing a unit test tonight.  My observation was really that the bug here was not caused by the lack of a test, the bug was that we didn't think through the implication of adding namespacing to AddressSpaces, i.e. that the notion of equality should have changed.  (So if there had previously been a test, it would have checked that equality was determined on name... which would have continued to pass after adding namespace as an attribute).  If a future change changes the notion of equality on namespaces then we'd have to rewrite the test anyway.

lulf:
+1 Don't do it if you don't think its right. I'm just projecting my bad conscience for introducing the bug in the first place! :)

PR #1286, opened 2018-05-29
change order of steps in pr builds

tkratky:


tkratky:
order of 'start openshift' and 'build images' steps changed because: setup openshift does remove of all docker images :)

PR #1289, opened 2018-05-29
Add label app=enmasse to all entities

dmvolod:
The activities with labels help to delete with selector like
oc delete all,pvc,configmap,rolebindings,secrets,sa -l 'app=enmasse'
I will update ansible configs and add remove action inside deploy.sh in the future if necessary

### DIFF
--- a/templates/include/api-server.jsonnet
+++ b/templates/include/api-server.jsonnet
@@ -145,7 +145,10 @@ local images = import "images.jsonnet";
         "apiVersion": "apiregistration.k8s.io/v1beta1",
         "kind": "APIService",
         "metadata": {
-          "name": "v1alpha1.enmasse.io"
+          "name": "v1alpha1.enmasse.io",
+          "labels": {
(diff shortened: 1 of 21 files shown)


dmvolod:
@lulf , added to .jsonnet files when was missing

templates/include/api-server.jsonnet
```diff
@@ -145,7 +145,8 @@ local images = import "images.jsonnet";
         "apiVersion": "apiregistration.k8s.io/v1beta1",
         "kind": "APIService",
         "metadata": {
-          "name": "v1alpha1.enmasse.io"
+          "name": "v1alpha1.enmasse.io",
+          "app": "enmasse"
```
lulf: Needs to be nested within a label

dmvolod: sorry, fixed

lulf:
@dmvolod Thanks!

PR #1290, opened 2018-05-29
added sleep

tkratky:


systemtests/scripts/test_func.sh
```diff
@@ -206,12 +206,14 @@ function categorize_dockerlogs {
 
 function stop_and_check_openshift() {
     oc cluster down #for the case that cluster is already running
+    sleep 10
```
lulf: Is the processes still running after cluster down has been run?

lulf: Sounds like it should be reported to openshift as well

PR #1293, opened 2018-05-30
Enable client TLS authentication for API server

### LINKED ISSUE #1265: Add support for mTLS configuration of aggregated api server
At present, the api server cannot be used as a kubernetes aggregated api server while RBAC is enabled. This limits its use outside of evaluation and experimentation purposes.

lulf:
* Support auth delegation when client certificates are valid
* Allow anonymous access but leave auth decisions up to kubernetes
* Remove enable_user_lookup setting
* Add additional cert to api server trust store
* Report proper v1.Status error responses

Fixes #1265

### DIFF
--- a/api-common/src/main/java/io/enmasse/api/auth/KubeAuthApi.java
+++ b/api-common/src/main/java/io/enmasse/api/auth/KubeAuthApi.java
@@ -5,7 +5,6 @@
 package io.enmasse.api.auth;
 
 import io.fabric8.kubernetes.api.model.Secret;
-import io.fabric8.openshift.api.model.User;
 import io.fabric8.openshift.client.NamespacedOpenShiftClient;
 import io.vertx.core.json.JsonObject;
 import okhttp3.*;
@@ -182,14 +181,4 @@ public class KubeAuthApi implements AuthApi {
     public String getNamespace() {
         return client.getNamespace();
     }
-
-    @Override
-    public String getUserId(String userName) {
-        User user = client.users().withName(userName).get();
-        if (user != null) {
-            return user.getMetadata().getUid();
-        }
-        return null;
-    }
-
 }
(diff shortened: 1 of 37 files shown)


api-common/src/main/java/io/enmasse/api/auth/KubeAuthApi.java
```diff
@@ -76,7 +76,13 @@ public TokenReview performTokenReview(String token) {
             spec.put("token", token);
             body.put("spec", spec);
 
+            if (log.isDebugEnabled()) {
```
rgodfrey: is it really worth doing the if(log.isDebugEnabled()) here (and the other places you've used this idiom)?

api-common/src/main/java/io/enmasse/api/common/DefaultExceptionMapper.java
```diff
@@ -28,22 +28,25 @@ public Response toResponse(Exception exception) {
         if (exception instanceof WebApplicationException) {
             WebApplicationException webApplicationException = (WebApplicationException) exception;
             status = Response.Status.fromStatusCode(webApplicationException.getResponse().getStatus());
-            response = webApplicationException.getResponse();
+            response = Response.status(status)
+                    .entity(new ErrorResponse(status.getStatusCode(), status.getReasonPhrase(), exception.getMessage()))
+                    .build();
+            webApplicationException.getResponse();
```
rgodfrey: Is this doing anything here, or just left behind after the change by mistake?

api-common/src/main/java/io/enmasse/api/auth/KubeAuthApi.java
lulf: Its done for every api interaction and the responses can be a medium json object, so I thought it might be good to guard it.

rgodfrey: sure but the JSON object won't get serialized unless isDebugEnabled is true anyway (since that'll be the first check in log.debug), the performance difference is just that between making a method call with no arguments and making a method call with two arguments which I would hope would be negligible

lulf: Ah you’re right, this is java! :)

PR #1294, opened 2018-05-30
FIX check oc 'cluster' status and sign in into correct cluster

tkratky:


tkratky:
@lulf this fix have to work, manually verified on my server

PR #1298, opened 2018-05-30
initial move to configmap configured routers

grs:


templates/include/address-space-definitions.yaml
```diff
@@ -1249,6 +1546,22 @@ data:
...
+          addressSpace: ${ADDRESS_SPACE}
+        labels:
+          app: enmasse
+        name: qdrouterd-headless
+      spec:
+        clusterIP: None
+        ports:
+        - name: inter-router
+          port: 55672
+          targetPort: 55672
```
lulf: Can/should these ports be removed from the «messaging» service now?

templates/include/address-space-definitions.yaml
```diff
@@ -63,8 +63,125 @@ data:
...
+      kind: ConfigMap
+      metadata:
+        name: qdrouterd-config
+      data:
+        qdrouterd.conf: |-
+          router {
+            mode: interior
+            id: ${HOSTNAME}
+            defaultDistribution: unavailable
+            allowResumableLinkRoute: ${ALLOW_DURABLE_SUBSCRIPTIONS}
```
lulf: Would be nice if worker threads were still configurable. Also, ideally the image should auto tune it based on cores in the system (or does it already do that?)

templates/install/resources/address-space-controller/address-space-definitions.yaml
```diff
@@ -977,8 +1159,125 @@ data:
...
+
+          sslProfile {
+            name: auth_service_ssl
+            certDb: /etc/qpid-dispatch/authservice-ca/tls.crt
+          }
+
+          authServicePlugin {
+            name: auth_service
+            authService: ${AUTHENTICATION_SERVICE_HOST}:${AUTHENTICATION_SERVICE_PORT}
+            #saslInitHostname: ${AUTHENTICATION_SERVICE_SASL_INIT_HOST}
```
lulf: Should this be un-commented?

templates/include/address-space-definitions.yaml
grs: They could be. That seems an orthogonal question to this change though in that I don't believe they are used before or after this patch.

templates/include/address-space-definitions.yaml
grs: They are still configurable through the configmap. There was no template parameter for them previously, so they would only have been configurable then be directly editing the deployment for the router.

address-space-controller/src/main/java/io/enmasse/controller/common/NoneAuthenticationServiceResolver.java
```diff
@@ -44,7 +45,8 @@ public int getPort(AuthenticationService authService) {
 
     @Override
     public Optional<String> getSaslInitHost(AddressSpace addressSpace, AuthenticationService authService) {
```
lulf: I guess we're saying that this field is really required now, does it make sense to change it to return a String instead of the optional?

grs: Yes, that probably does make sense.

PR #1300, opened 2018-05-31
Minor fixes

lulf:


templates/install/deploy.sh
```diff
@@ -212,7 +212,7 @@ do
         if [ -n "$USE_OPENSHIFT" ]; then
             runcmd "oc create -n ${NAMESPACE} -f ${RESOURCE_DIR}/standard-authservice/route.yaml" "Create standard authservice route"
             OAUTH_DISABLED="false"
-            HTTP_URL="https://$(oc get route keycloak -o yaml -o jsonpath={.spec.host})/auth"
+            HTTP_URL="https://$($CMD get service standard-authservice -n ${NAMESPACE} -o jsonpath={.spec.clusterIP}):8443/auth"
```
calohmn: @lulf doesn't this make the Console OAuth redirect fail in the (not unlikely) case that the cluster-ip is not reachable from where the Console is requested? I thought with using the route host, one would be more on the safe side here (although for the internal keycloak connections, the above looks like the better solution).

lulf: @calohmn If using oc cluster up (a local openshift in docker running), using the route won't work, because keycloak-enmasse.127.0.0.1.nip.io will resolve to localhost, so the keycloak-controller won't work in that case.

If you need to deploy a public cluster, we should either add an option to override it with the route, or use ansible which supports setting this explicitly.

calohmn: @lulf How about detecting the local cluster case with route URL `*.127.0.0.1.nip.io` (where OpenShift public-hostname is 127.0.0.1) and using the `oc get service standard-authservice` clusterIP in that case - and using the `oc get route keycloak` host in the other cases?
That would also make the Minishift case work again.

lulf: Yes, I think that would work.

PR #1303, opened 2018-06-01
Test debug

tkratky:


tkratky:
io.enmasse.systemtest.common.plans.PlansTest.testQuotaLimitsPooled test fixed in this branch

systemtests/src/main/java/io/enmasse/systemtest/selenium/page/ConsoleWebPage.java
```diff
@@ -275,10 +275,14 @@ public Integer getResultsCount() throws Exception {
      */
     public AddressWebItem getAddressItem(Destination destination) throws Exception {
         AddressWebItem returnedElement = null;
-        List<AddressWebItem> addressWebItems = getAddressItems();
-        for (AddressWebItem item : addressWebItems) {
-            if (item.getName().equals(destination.getAddress()))
-                returnedElement = item;
+        TimeoutBudget budget = new TimeoutBudget(15, TimeUnit.SECONDS);
+
+        while(!budget.timeoutExpired() && returnedElement == null) {
```
kornys: This is good approach 👍 , but anyway in situations or in tests where we want check if count of address elements is zero and we expected it, it will wait unnecessarily 15s.

PR #1316, opened 2018-06-04
FIX check if ansible ready

kornys:


systemtests/scripts/test_func.sh
```diff
@@ -255,3 +255,11 @@ function clean_oc_location() {
     sudo rm -rf /var/log/containers/*
     sudo rm -rf /var/log/pods/*
 }
+
+function check_if_ansible_ready() {
+    info "Running: check if ansible is already installed"
+    if ! sudo rpm -qa | grep -qw ansible; then
```
tkratky: "if ! ansible; then" should work as well

kornys:
@tkratky we have to use jenkins manage file as well, because we need to set up internal repos etc..., anyway for now I would leave it as it is. and as you said move some dependencies to ansible 👍

PR #1326, opened 2018-06-06
Register api server in systemtests deployment

kornys:


systemtests/src/main/java/io/enmasse/systemtest/Endpoint.java
```diff
@@ -14,6 +17,17 @@ public Endpoint(String host, int port) {
         this.port = port;
     }
 
+    public Endpoint(String url) {
+        URL myurl = null;
+        try {
+            myurl = new URL(url);
+        } catch (MalformedURLException e) {
+            e.printStackTrace();
```
lulf: Perhaps this should be fatal? I.e. rethrow exception

PR #1333, opened 2018-06-07
Durable subscriptions impl

grs:
Some initial implementation. This allows subscriptions to be created via console. Only subscriptions on pooled topics currently get correctly allocated by standard-controller (subscriptions on sharded topics need to be explicitly allocated via editing the configmap at present, pending further updates to standard-controller).

Subscriptions are accessed by fully qualified name e.g. mytopic::mysubscription.

I've tried to separate into discrete commits in case anyone wants to cherry-pick parts of it.


address-model-lib/src/main/java/io/enmasse/address/model/Address.java
```diff
@@ -23,11 +23,12 @@
     private final String addressSpace;
     private final String type;
     private final String plan;
+    private final Optional<String> topic;
```
lulf: This is just a java 'convention' nitpick. There seems to be an agreement that using Optional as a field in a data structure is a misuse of the API and should be avoided. Returning an Optional<> is however fine. Most of the code in this library is already following this convention.

The 'getTopic()' however should instead use `return Optional.ofNullable(topic)`

standard-controller/src/main/java/io/enmasse/controller/standard/AddressProvisioner.java
```diff
@@ -90,7 +90,30 @@ private void addToUsage(Map<String, Map<String, UsageInfo>> usageMap, Address ad
...
+        String cluster = topic.getAnnotations().get(AnnotationKeys.CLUSTER_ID);
+        String broker = topic.getAnnotations().get(AnnotationKeys.BROKER_ID);
+        subscription.getAnnotations().put(AnnotationKeys.CLUSTER_ID, cluster);
+        if (broker != null) {
+            brokerUsage.get(broker).addUsed(requested.getAmount());
+            subscription.getAnnotations().put(AnnotationKeys.BROKER_ID, broker);
+        } else {
+            brokerUsage.get(cluster).addUsed(requested.getAmount());
+            log.info("need to allocate subscription {} on {} to broker from {}", subscription.getAddress(), topic.getAddress(), cluster);
+            //TODO: find least used broker on topic cluster
```
lulf: I think in this place, we should do something similar to 'scheduleAddress' which sorts a list of brokers by usage and allocates to the least used broker.

However, I guess the problem here is that the UsageInfo for sharded address plans only tracks the usage of the cluster in sum, and not per broker (see https://github.com/EnMasseProject/enmasse/blob/ce82d1bcaa042e8718c5706b533c9f250f9847ab/standard-controller/src/main/java/io/enmasse/controller/standard/AddressProvisioner.java#L50). 

The usage of a sharded broker would be the `creditsUsedByAddress + sumCreditsForSubscriptionsOnBroker`.

agent/www/tooltips.json
```diff
@@ -32,6 +32,10 @@
                 "longDescription": "The queue address type is a store-and-forward queue. This address type is appropriate for implementing a distributed work queue, handling traffic bursts, and other use cases where you want to decouple the producer and consumer. A queue can be sharded across multiple storage units; however, message order is no longer guaranteed.",
                 "shortDescription": "A store-and-forward queue"
             },
+            "subscription": {
+                "longDescription": "The subscription address type allows a subscription to be created for a topic which will hold messages published to the topic even if the subscriber is not attached. The subscription is accessed by the consumer via <topic-address>::<subscription-address>. E,g, for a subscription 'mysub' on a topic 'mytopic', the consumer should consume from address 'mytopic::mysub'.",
+                "shortDescription": "A subscription on a chosen topic"
```
jenmalloy: Change "chosen" to "specified" ?

documentation/common/address-space.adoc
```diff
@@ -44,6 +45,13 @@ The anycast address type is a scalable direct address for sending messages to on
 The multicast address type is a scalable direct address for sending messages to multiple consumers. Messages sent to a multicast address are forwarded to all consumers receiving messages on that address. It is important to note that only pre-settled messages can be sent to multicast addresses, as message acknowledgements from consumers are not propagated to producers.
 // standard.address.multicast.longDescription:stop
 
+[[subscription]]
+==== Subscription
+// standard.address.subscription.shortDescription:A subscription on a chosen topic
+// standard.address.subscription.longDescription:start
+The subscription address type allows a subscription to be created for a topic which will hold messages published to the topic even if the subscriber is not attached. The subscription is accessed by the consumer via <topic-address>::<subscription-address>. E,g, for a subscription 'mysub' on a topic 'mytopic', the consumer should consume from address 'mytopic::mysub'.
```
jenmalloy: Change as indicated above

agent/www/tooltips.json
```diff
@@ -32,6 +32,10 @@
                 "longDescription": "The queue address type is a store-and-forward queue. This address type is appropriate for implementing a distributed work queue, handling traffic bursts, and other use cases where you want to decouple the producer and consumer. A queue can be sharded across multiple storage units; however, message order is no longer guaranteed.",
                 "shortDescription": "A store-and-forward queue"
             },
+            "subscription": {
+                "longDescription": "The subscription address type allows a subscription to be created for a topic which will hold messages published to the topic even if the subscriber is not attached. The subscription is accessed by the consumer via <topic-address>::<subscription-address>. E,g, for a subscription 'mysub' on a topic 'mytopic', the consumer should consume from address 'mytopic::mysub'.",
```
jenmalloy: Change as indicated above

grs:
@jenmalloy changes made as suggested (just fyi, the file under documentation is the source for the other changes which are generated)

jenmalloy:
@grs, ah, right! thanks for the reminder, will pay close attention to that going forward

grs:
@jenmalloy actually the generated files have been removed from git now so it will be less confusing

systemtests/src/test/java/io/enmasse/systemtest/standard/TopicTest.java
```diff
@@ -237,6 +237,34 @@ void testMessageSelectorsProperty() throws Exception {
                 received.get(1, TimeUnit.MINUTES).get(0).getGroupId(), is(groupID));
     }
 
+    @Test
+    void testDurableSubscriptionOnPooledTopic() throws Exception {
+        Destination topic = Destination.topic("mytopic", "pooled-topic");
+        Destination subscription = Destination.topic("mysub", "standard-subscription");
```
lulf: Shouldn't this be Destination.subscription?

grs: @lulf, yes! Thanks! Been chasing that down for ages. :(

lulf:
@grs Not much to do about it now, but I think it should have been squashed a bit :)

PR #1336, opened 2018-06-08
wait for address space ready before reload endpoints

tkratky:


lulf:
@tkratky You can close this i think

kornys:
@tkratky I think you can remove waiting from first commit and merge only second commit which add some useful log :)

PR #1337, opened 2018-06-08
Get endpoint object from correct json object

lulf:
@tkratky @kornys I think this is what broke systemtests :)


lulf:
Test passed locally

PR #1340, opened 2018-06-11
Pr tests reduction

kornys:


tkratky:
From long time point of view is not possible to run all of our tests against each PR, right now it takes more then 6 hours. So we made a decision to remove all SmokeTests (because these are running in Travis) and few other tests that should be at least partly covered by another tests. So it doesn't mean that some of our tests are redundant. These tests just verify SUT functionality deeper. Anyway full tests suite will be executed against master, so there are no tests that shouldn't be used. (except list of disabled tests).

After our changes full PR pipeline take 4h 30m. I would like to keep this time under 5 hours if possible.

PR #1347, opened 2018-06-12
Build agent using maven

lulf:
* Use asciidoctor maven plugin to generate docs
* Use exec maven plugin to run python to generate tooltips
* Use frontend-maven-plugin to download fixed version of node/npm and
use for building agent
* Update make targets to build agent artifacts from maven


grs:
The log suggests the node tests aren't being run?

```
[INFO] --- maven-surefire-plugin:2.20.1:test (default-test) @ agent ---
[INFO] Tests are skipped.
[INFO] 
[INFO] --- frontend-maven-plugin:1.6:npm (npm test) @ agent ---
[INFO] Skipping execution.
```

lulf:
@grs Ah, i somehow accidentally deleted the 'test' target when removing the -q to get some more info from the build.

I've not tested the resulting agent image yet, so probably more issues with this.

 I'd like like to know your thoughts on the general approach though

lulf:
@grs There is an issue at present where the npm install is run without --production, in which case you get a lot more stuff in the node_modules. I'll try to fix it so it doesnt do that.

lulf:
@grs So I think I've sorted out the issues. Like with the make files, I run npm install separately for the 'production' build.

Also the tests seems to be invoked now. The travis build is a bit deceiving in that it first does a mvn install -DskipTests before invoking our travis script, so that may show up as skipping tests.

The actual build is running the agent tests now: https://travis-ci.org/EnMasseProject/enmasse/builds/391924903#L5454

 Please have another look.

lulf:
Merging, as CI failures are from other changes and being fixed.

PR #1348, opened 2018-06-12
Provisioner unit tests

lulf:
@grs @vbusch Here are some basic unit tests for the durable subscriptions.

The pooled test passes, but the sharded test fails with a NPE in the current code.

As we agreed, I made this PR against the 'development work' branch.


lulf:
@grs @vbusch There is also another potential issue with pooled subscriptions in the current implementation, where you have 1 topic on a pooled broker, and then add many subscribers causing another broker to be allocated. However, the subscribers allocated to the second broker will not get any messages since the topic does not exist there!

I think it is similar to what we discussed in the sharded case, that we should track subscriptions independently from the topics themselves, and ensure subscriptions are allocated on the same broker as the topic.

Issue #1342, opened 2018-06-11
Move systemtest playbooks into systemtests


lulf:
The new playbook structure involves having a split between inventory files and playbooks.

Playbooks located in `templates/install/ansible/playbooks/openshift` and example inventory files in `templates/install/ansible/inventory`. 

Right now the systemtest playbooks and roles are part of the release bundle, which I think they should not be.

I propose that instead the systemtest roles and playbooks are moved into the 'systemtests` folder, and that the playbooks for installing 'enmasse' and 'systemtests' are run independently. This way we can verify that the 'install' playbook is working correctly.

There should also be a 'systemtest.inventory' file used when deploying the install playbook. I.e. I'm using the following locally:

```
[enmasse]
localhost ansible_connection=local

[enmasse:vars]
namespace=enmasseci
multitenant=True
enable_rbac=True
api_server=True
register_api_server=False
service_catalog=True
secure_api_server=False
keycloak_admin_password=admin
authentication_services=["none", "standard"]

```

kornys:
@lulf  I agree

PR #1349, opened 2018-06-13
Move systemtests roles

kornys:


.travis/setup-kubernetes.sh
```diff
@@ -10,7 +10,7 @@ ENMASSE_DIR=${1}
 OPENSHIFT_CLIENT_URL=${2:-"https://github.com/openshift/origin/releases/download/v3.7.0/openshift-origin-client-tools-v3.7.0-7ed6862-linux-64bit.tar.gz"}
 MINIKUBE_RELEASE_URL=${3:-"https://storage.googleapis.com/minikube/releases/v0.25.2/minikube-linux-amd64"}
 KUBECTL_RELEASE_URL=${4:-"https://storage.googleapis.com/kubernetes-release/release/v1.8.7/bin/linux/amd64/kubectl"}
-ansible-playbook ${ENMASSE_DIR}/ansible/playbooks/openshift/environment.yml \
+ansible-playbook ${ENMASSE_DIR}/ansible/playbooks/environment.yml \
```
lulf: It's not really ENMASSE_DIR anymore, so it should be renamed.

systemtests/scripts/test_func.sh
```diff
@@ -29,7 +29,8 @@ function setup_test() {
     oc login -u ${OPENSHIFT_USER} -p ${OPENSHIFT_PASSWD} --insecure-skip-tls-verify=true ${OPENSHIFT_URL}
     oc adm --config ${KUBEADM} policy add-cluster-role-to-user cluster-admin $OPENSHIFT_USER
     export OPENSHIFT_TOKEN=`oc whoami -t`
-    ansible-playbook ${ENMASSE_DIR}/ansible/playbooks/openshift/systemtests.yml --extra-vars "namespace=${OPENSHIFT_PROJECT} admin_user=${OPENSHIFT_USER}"
+    ansible-playbook ${CURDIR}/../ansible/playbooks/systemtests-dependencies.yml
+    ansible-playbook ${ENMASSE_DIR}/ansible/playbooks/openshift/install.yml -i ${CURDIR}/../ansible/inventory/systemtests.inventory --extra-vars "namespace=${OPENSHIFT_PROJECT} admin_user=${OPENSHIFT_USER}"
```
lulf: Maybe I'm just confused, but I think ENMASSE_DIR will point to 'systemtests' in this case, and not 'templates/install', so the install playbook does not exist there.

kornys: @lulf I will rename ENMASSE_DIR you're right name is confusing, but install playbook is still there it is your install playbook but I run this playbook with systemtests.inventory

kornys:
https://github.com/EnMasseProject/enmasse/issues/1342

lulf:
@kornys I missed this in the review, but the HACKING.md should be updated with the correct commands and paths.

kornys:
@lulf you're right, I will do that. Thanks!

PR #1350, opened 2018-06-13
Support longer address names

lulf:
* Change address/configmap name to follow addressSpace.sanitizedAddress.uuidFromAddress
* Ensure sharded brokers get a short uuid based on address + uuid of name
* Simplify handling of cluster id in standard controller
* Add systemtest deploying a sharded queue with a long name


calohmn:
I think it would be good to have extra validation in the `POST` and `PUT` methods of the `HttpNestedAddressService` - checking whether the `metadata.name` value in the request JSON contains the "{addressSpaceName}." prefix with the address space name matching the corresponding part of the request URI.
I guess this could be done in `AddressApiHelper.validateAddress`.

lulf:
@calohmn At present though, the format is not mandated. It is the format we will use if 'name' is not specified in the payload. If it is specified, we just take care to validate it not containing any illegal characters. The name should not be interpreted by any piece of code at this point I think.

calohmn:
@lulf Ah, ok. 
The name is interpreted as beginning with the addressSpace in connection with the `/apis/enmasse.io/v1alpha1/namespaces/{namespace}/addresses` requests, though (`HttpAddressService.parseAddressSpace()`).
- POST/PUT requests must have `metadata.name` in the request JSON already set, beginning with the address space name. Using `metadata.addressSpace` instead (and omitting `metadata.name`) would seem more natural here, IMHO. And that way the client doesn't have to set a sanitized name.
- GET/DELETE requests are restricted to addresses having been created with the standard naming convention - to be noted in the API docs.

lulf:
@calohmn I added some validation to the Address.validate() which is called before anything is serialized and will return an error through the API.

PR #1353, opened 2018-06-13
Fix unit test failure

lulf:


lulf:
@grs Yes. When I run it in intellij it fails. But running it from maven seemed to work. Could be there is some race in the test here.

lulf:
Actually, the test is badly written. I will fix it

lulf:
@grs As suspected, the test is flawed. The test could complete too early before having checked all requests. If the awaitSuccess() was called before the second request was sent, the test would pass.

Issue #403, opened 2017-10-25
Add the ability to use AMQP over WebSockets to a route on port 443 


rgodfrey:
For clients sitting behind corporate firewalls which block outgoing traffic other than HTTP/HTTPS, we should provide messaging over websockets for both the standard and brokered address spaces.

kornys:
@lulf we have tests for ws with brokered. In standard is not ws allowed yet

lulf:
@kornys Ok thanks! Lets leave this open then

lulf:
@rgodfrey Maybe the route, service etc. should not be called http but websocket ? In my mind, this is more accurate and would not prevent us from having a 'http' service in the future that does http - amqp bridging.

lulf:
Or maybe 'amqp-ws', 'amqp-websocket' or 'messaging-websocket'

lulf:
@rgodfrey @grs I guess http -> amqp-ws and https -> amqp-wss ?

Another point, there is a 1-1 mapping between service and route. We can either improve the model to support multiple routes for a service (i.e. messaging, amqp-ws, amqp-wss all pointing to the 'messaging' service having multiple ports). Or we can create one service for each (so we'd end up with messaging, amqp-ws, amqp-wss as services as well. I think the former is the cleanest in that we don't end up with many service resources, but it requires some more modelling.

grs:
At the moment we have messaging which does for TLS and non-TLS based amqp, right? I would do the same for websockets (with only the TLS based access available outside the cluster, as for messaging).

grs:
Misunderstood your point. I'm fine either way, a new service or using messaging.

lulf:
Reopening as there is an issue with authenticatePeer: yes and http listeners causing router to authenticate using client certs.

PR #1355, opened 2018-06-13
Add example rhea client using websockets

lulf:
Sorry I missed the documentation on my previous PR. This one has a client example for using websockets.

Potentially there are more documentation needed for clients, but I don't know the best place to put that.

Issue #403

### DIFF
--- a/documentation/messaging_app_developer/connecting.adoc
+++ b/documentation/messaging_app_developer/connecting.adoc
@@ -152,7 +152,31 @@ container.on('sendable', function (context) {
     context.sender.send({body:'Hello World!'});
     context.sender.detach();
 });
-container.connect({port:443, host:'<messaging-route-hostname>', transport:'tls', rejectUnauthorized:false});
+container.connect({username: '<username>', password: '<password>', port:443, host:'<messaging-route-hostname>', transport:'tls', rejectUnauthorized:false});
+----
+
+==== {ClientAmqpJavaScriptName} using WebSockets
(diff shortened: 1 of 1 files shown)


documentation/messaging_app_developer/connecting.adoc
```diff
@@ -155,6 +155,29 @@ container.on('sendable', function (context) {
...
+});
+container.on('message', function (context) {
+    console.log(context.message.body);
+    context.connection.close();
+});
+container.on('sendable', function (context) {
+    context.sender.send({body:'Hello World!'});
+    context.sender.detach();
+});
+container.connect({username: 'user1', password: 'password1', transport:'tls', rejectUnauthorized:false, connection_details: ws("wss://<messaging route hostname>:443", ["binary"], {rejectUnauthorized: false})});
```
grs: The 'transport:'tls', rejectUnauthorized:false,' part will be ignored as connection_details is specified (and wss implies tls and rejectUnauthorized is specified as an option to the websocket).

lulf: @grs Should be fixed now. Also I removed the username/password for now, as there isn't any provided for the other client examples.

grs: For the javascript examples we should leave in the username (e.g. as anonymous), as otherwise it will not use a sasl layer, which is required. Other languages should use sasl layer by default.

PR #1366, opened 2018-06-14
Disable auto create of addresses in brokered address space

rgodfrey:
should stop openwire clients failing on sending to a topic

### DIFF
--- a/agent/lib/broker_controller.js
+++ b/agent/lib/broker_controller.js
@@ -359,11 +359,14 @@ BrokerController.prototype.create_address = function (a) {
     var self = this;
     if (a.type === 'queue') {
         log.info('[%s] Creating queue "%s"...', self.id, a.address);
-        return self.broker.createQueue(a.address).then(function () {
+        return self.broker.createAddress(a.address, {anycast:true}).then(self.broker.createQueue(a.address).then(function () {
             log.info('[%s] Created queue "%s"', self.id, a.address);
             self.post_event(myevents.address_create(a));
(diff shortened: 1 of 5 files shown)


agent/testlib/mock_broker.js
```diff
@@ -117,10 +117,8 @@ function MockBroker (name) {
             return self.list_queues().map(function (a) { return a.name; });
         },
         createQueue : function (address, routingTypes, name, filter, durable, maxConsumers, purgeOnNoConsumers, autoCreateAddress) {
-            assert.equal(routingTypes, 'ANYCAST');
```
grs: I don't think this needs to be removed, does it?

agent/lib/broker_controller.js
```diff
@@ -359,12 +359,15 @@ BrokerController.prototype.create_address = function (a) {
     var self = this;
     if (a.type === 'queue') {
         log.info('[%s] Creating queue "%s"...', self.id, a.address);
-        return self.broker.createQueue(a.address).then(function () {
+        return self.broker.createAddress(a.address, {anycast:false}).then(function () { return self.broker.createQueue(a.address).then(function () {
```
lulf: Is anycast:false correct here? A iirc, a queue is anycast: true in the broker

rgodfrey:
Unfortunately this approach will not work as we JMS temporary queues rely on auto-create-address being true.  However the issue in Artemis which causes this problem is resolved in version 2.6.1

Issue #1271, opened 2018-05-25
Create rolebindings for Kubernetes deployment if RBAC enabled


calohmn:
On clusters with RBAC support enabled, the Kubernetes deployment with `deploy.sh` will not work because of missing rolebindings.

As also minikube seems to have RBAC enabled by default now (tested with 0.27.0), it would be good to create the necessary rolebindings in `deploy.sh` if RBAC is enabled.

Related issue: #201.

calohmn:
It would be good to use more fine-grained permissions than the ones in #201 / PR #203.

Adding a rolebinding to the admin clusterrole
````
kubectl create rolebinding enmasse-admin-binding --clusterrole=admin --serviceaccount=${NAMESPACE}:enmasse-admin --namespace=${NAMESPACE}
````
leads to one remaining issue in the address-space-controller:
````
2018-05-24 16:52:06 WARN  ControllerChain:124 - Error syncing address space default
io.fabric8.kubernetes.client.KubernetesClientException: Failure executing: GET at: https://10.96.0.1/. Message: Forbidden!Configured service account doesn't have access. Service account may have been revoked. forbidden: User "system:serviceaccount:enmasse:enmasse-admin" cannot get path "/".
	at io.fabric8.kubernetes.client.dsl.base.OperationSupport.requestFailure(OperationSupport.java:470)
	at io.fabric8.kubernetes.client.dsl.base.OperationSupport.assertResponseCode(OperationSupport.java:407)
	at io.fabric8.kubernetes.client.dsl.base.OperationSupport.handleResponse(OperationSupport.java:379)
	at io.fabric8.kubernetes.client.dsl.base.OperationSupport.handleResponse(OperationSupport.java:343)
	at io.fabric8.kubernetes.client.dsl.base.OperationSupport.handleResponse(OperationSupport.java:327)
	at io.fabric8.kubernetes.client.dsl.base.BaseOperation.getRootPaths(BaseOperation.java:236)
	at io.fabric8.kubernetes.client.BaseClient.rootPaths(BaseClient.java:140)
	at io.fabric8.openshift.client.OpenshiftAdapterSupport.isOpenShift(OpenshiftAdapterSupport.java:53)
	at io.fabric8.openshift.client.OpenshiftAdapterSupport.isOpenShiftAPIGroups(OpenshiftAdapterSupport.java:92)
[...]

PR #1372, opened 2018-06-18
Create AddressSpace RoleBindings for Kubernetes (#1271)

calohmn:
Part of the fix for #1271.
This makes `addAddressSpaceAdminRoleBinding()` and `addAddressSpaceRoleBindings()` in `KubernetesHelper` work in case of a Kubernetes cluster.

### DIFF
--- a/address-space-controller/src/main/java/io/enmasse/controller/common/KubernetesHelper.java
+++ b/address-space-controller/src/main/java/io/enmasse/controller/common/KubernetesHelper.java
@@ -363,8 +363,9 @@ public class KubernetesHelper implements Kubernetes {
 
     private void createRoleBinding(String name, String namespace, String refKind, String refName, List<Subject> subjectList) {
 
-        String apiVersion = client.isAdaptable(OpenShiftClient.class) ? "v1" : "rbac.authorization.k8s.io/v1beta1";
-        String apiPath = client.isAdaptable(OpenShiftClient.class) ? "/oapi/v1" : "/apis/rbac.authorization.k8s.io/v1beta1";
+        Boolean isOpenShift = client.isAdaptable(OpenShiftClient.class);
+        String apiVersion = isOpenShift ? "v1" : "rbac.authorization.k8s.io/v1";
+        String apiPath = isOpenShift ? "/oapi/v1" : "/apis/rbac.authorization.k8s.io/v1";
 
         JsonObject body = new JsonObject();
 
@@ -386,7 +387,9 @@ public class KubernetesHelper implements Kubernetes {
 
         for (Subject subjectEntry : subjectList) {
             JsonObject subject = new JsonObject();
-            subject.put("apiGroup", "rbac.authorization.k8s.io");
+            if (isOpenShift) {
+                subject.put("apiGroup", "rbac.authorization.k8s.io");
+            }
             subject.put("kind", subjectEntry.getKind());
             subject.put("name", subjectEntry.getName());
             if (subjectEntry.getNamespace() != null) {
@@ -403,7 +406,7 @@ public class KubernetesHelper implements Kubernetes {
 
     @Override
     public void addAddressSpaceAdminRoleBinding(AddressSpace addressSpace) {
-        if (isRBACSupported() && client.isAdaptable(OpenShiftClient.class)) {
+        if (isRBACSupported()) {
             createRoleBinding("addressspace-admins", addressSpace.getAnnotation(AnnotationKeys.NAMESPACE), "ClusterRole", "admin", Arrays.asList(
                     new Subject("ServiceAccount", addressControllerSa, namespace)));
         }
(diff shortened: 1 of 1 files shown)


address-space-controller/src/main/java/io/enmasse/controller/common/KubernetesHelper.java
```diff
@@ -434,12 +437,12 @@ public void createServiceAccount(String namespace, String saName) {
...
 
-        if (isRBACSupported() && client.isAdaptable(OpenShiftClient.class)) {
+        if (isRBACSupported()) {
             createRoleBinding("address-space-viewers", namespace, "ClusterRole", "view", Arrays.asList(
-                    new Subject("ServiceAccount", "default", namespace)));;
+                    new Subject("ServiceAccount", "default", namespace)));
             createRoleBinding("address-admins", namespace, "ClusterRole", "edit", Arrays.asList(
                     new Subject("ServiceAccount", addressSpaceAdminSa, namespace)));
-            if (hasClusterRole("event-reporter")) {
+            if (client.isAdaptable(OpenShiftClient.class) && hasClusterRole("event-reporter")) {
```
calohmn: @lulf I've assumed here that the "event-reporter" role is OpenShift specific. Is that correct?

lulf: @calohmn I think we can actually report the whole if sentence

address-space-controller/src/main/java/io/enmasse/controller/common/KubernetesHelper.java
```diff
@@ -386,7 +387,9 @@ private void createRoleBinding(String name, String namespace, String refKind, St
 
         for (Subject subjectEntry : subjectList) {
             JsonObject subject = new JsonObject();
-            subject.put("apiGroup", "rbac.authorization.k8s.io");
+            if (isOpenShift) {
+                subject.put("apiGroup", "rbac.authorization.k8s.io");
```
lulf: Is this not the correct api group for kubernetes as well?

address-space-controller/src/main/java/io/enmasse/controller/common/KubernetesHelper.java
```diff
@@ -363,8 +363,9 @@ private boolean isRBACSupported() {
 
     private void createRoleBinding(String name, String namespace, String refKind, String refName, List<Subject> subjectList) {
 
-        String apiVersion = client.isAdaptable(OpenShiftClient.class) ? "v1" : "rbac.authorization.k8s.io/v1beta1";
-        String apiPath = client.isAdaptable(OpenShiftClient.class) ? "/oapi/v1" : "/apis/rbac.authorization.k8s.io/v1beta1";
+        Boolean isOpenShift = client.isAdaptable(OpenShiftClient.class);
+        String apiVersion = isOpenShift ? "v1" : "rbac.authorization.k8s.io/v1beta1";
+        String apiPath = isOpenShift ? "/oapi/v1" : "/apis/rbac.authorization.k8s.io/v1beta1";
```
lulf: Why v1beta1 instead of v1? I think we should use a stable api available in 1.9.

address-space-controller/src/main/java/io/enmasse/controller/common/KubernetesHelper.java
calohmn: @lulf setting the apigroup here caused this exceptions:
````
2018-06-18 11:46:17.405 WARN  KubernetesHelper:345 - Error performing POST on /apis/rbac.authorization.k8s.io/v1beta1/namespaces/bosch/rolebindings: 422, 
{"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":
"RoleBinding.rbac.authorization.k8s.io \"addressspace-admins\" is invalid: 
subjects[0].apiGroup: 
Unsupported value: \"rbac.authorization.k8s.io\": 
supported values: \"\"","reason":"Invalid","details":{"name":"addressspace-admins","group":"rbac.authorization.k8s.io","kind":"RoleBinding","causes":[{"reason":"FieldValueNotSupported","message":"
Unsupported value: \"rbac.authorization.k8s.io\": supported values: \"\"","field":"subjects[0].apiGroup"}]},"code":422}
````

address-space-controller/src/main/java/io/enmasse/controller/common/KubernetesHelper.java
calohmn: Ah, right. Will change to v1.

address-space-controller/src/main/java/io/enmasse/controller/common/KubernetesHelper.java
lulf: @calohmn Ok, interesting. We can remove this once OpenShift 3.7 support goes away.

address-space-controller/src/main/java/io/enmasse/controller/common/KubernetesHelper.java
calohmn: @lulf you mean remove the "if" line? On my minikube cluster, there is no "event-reporter" ClusterRole, though. Or do you mean remove the whole "if" block?

calohmn:
I've amended the commit based on the comments.

PR #1375, opened 2018-06-19
Enabled authz tests

kornys:


systemtests/src/test/java/io/enmasse/systemtest/bases/authz/AuthorizationTestBase.java
```diff
@@ -89,6 +89,7 @@ protected void doTestUserPermissionAfterRemoveAuthz() throws Exception {
         createUser(sharedAddressSpace, user, Group.RECV_ALL_BROKERED.toString(), Group.RECV_ALL_STANDARD.toString());
         assertReceive(user);
         removeUser(sharedAddressSpace, user.getUsername());
+        Thread.sleep(5000);
```
lulf: Why is this needed?

lulf: @kornys Hmm, that seems like a keycloak bug to me. I agree we need the workaround for now, but we should ask about the behavior.

PR #1382, opened 2018-06-19
Create ClusterRoleBindings for Kubernetes (#1271)

calohmn:
Fix for #1271:
Ensures that EnMasse is working on Kubernetes clusters with RBAC enabled by adding needed ClusterRoleBindings. Tested with singleTenant and multiTenant mode.

I have left the OpenShift specific "v1" RBAC namespace in the new "cluster-roles" file (and only adapt it on the fly) so that `oc create -f ${RESOURCE_DIR}/cluster-roles/` still works.

I haven't tested this on OpenShift, but the only change in this regard here is the added segment in `cluster-roles/address-space-controller.yaml` - all other places are Kubernetes-specific. 
I guess it could make sense to create separate `cluster-roles/address-space-controller.yaml` files for OpenShift and Kubernetes. Wdyt?


lulf:
I think splitting the address-space-controller.yaml files are probably better, but remember to update the docs in documentation/service_admin as well.

I don't think we should add the template-operations cluster-role-binding in the address space controller. This is something that is likely to forever be a kubernetes-specific thing, and I'd like to move against not having to do kubernetes or openshift specific stuff post 3.9.

Is one acceptable short-term fix to configure kubernetes to allow unauthenticated access to '/' which seems to be the case for openshift?

I think longer term, we should:

1) get rid of pre-3.9 rbac stuff
2) fix the kubernetes-client so it doesn't connect to the api server when doing local template processing (causing the exception in the standard-controller).

With the above fixed, I think there would not be a need for any isOpenShift in both address-space-controller and standard-controller.

lulf:
@calohmn I.e short term one could do this:

`kubectl create clusterrolebinding stuff --clusterrole=enmasse.io:template-operations --group=system:unauthenticated`

Btw, it seems strange that users can get /apis  but not / by default.

calohmn:
@lulf good points. I've amended the commit and assigned the "template-operations" clusterrole (renamed to `enmasse.io:k8s-api-get-root-url`) to `--group=system:serviceaccounts`. That reduces the changes in `deploy.sh` and keeps `KubernetesHelper` unchanged.

Concerning kubernetes/openshift specific stuff, I've put that in corresponding sub-directories of "cluster-roles":
````
kubernetes/address-space-controller.yaml
kubernetes/k8s-api-get-root-url.yaml
openshift/address-space-controller.yaml
openshift/keycloak-controller.yaml
````

File: templates/include/cluster-roles/kubernetes/address-space-controller.yaml [hunk 1]
```diff
@@ -0,0 +1,16 @@
...
+  name: enmasse.io:address-space-controller
+rules:
+  - apiGroups: [ "" ]
+    resources: [ "namespaces" ]
+    verbs: [ "list", "get", "create" ]
+  - apiGroups: [ "rbac.authorization.k8s.io" ]
+    resources: [ "rolebindings", "clusterrolebindings" ]
+    verbs: [ "list", "get", "create" ]
+  - apiGroups: [ "rbac.authorization.k8s.io" ]
+    resources: [ "clusterroles" ]
```
lulf: I guess this one is no longer needed? Or is the bind required for other things?

calohmn:
Oh, yes (had the changed main.ymls hidden in different IDEA changelist).

templates/include/cluster-roles/kubernetes/address-space-controller.yaml
calohmn: Right, the only thing still needed now is
````
  - apiGroups: [ "rbac.authorization.k8s.io" ]
    resources: [ "rolebindings" ]
    verbs: [ "list", "get", "create" ]
````

PR #1389, opened 2018-06-20
Address Space Quota Support

lulf:


lulf:
@rgodfrey I'm tempted to say we should not touch the users instances if the quota is changed but state that it only affects creating new address spaces.

I'm imagining a scenario where a quota change was applied by mistake, causing deletion of address spaces. And it would have to choose a random (or using some heuristic) candidate to delete. 

We could also consider a hard vs. soft limit, where lowering the hard limit could case address spaces to be deleted, whereas the soft limit would allow the count to go above for existing address spaces. 

I'm not sure if it matters in practice, i.e. one could periodically check users going above the quota, and then 'manually' interact with users to resolve the issue.

rgodfrey:
@lulf so I think it would be useful if there were an easy way to extract a "users over quota" list, possibly including all the addressspaces associated with the user... where do you think the best place to implement that would be?

lulf:
@rgodfrey How about a REST endpoint like:

POST /apis/enmasse.io/v1alpha1/addressspacequotareviews

Payload
```
kind: AddressSpaceQuotaReview
spec:
  user: developer
```

Response
```
kind: AddressSpaceQuotaReview
spec:
  user: developer
status:
  exceeded: false
```

Modelled after token reviews and subject access reviews in kubernetes:

https://kubernetes.io/docs/reference/access-authn-authz/authorization/

rgodfrey:
@lulf works for me

lulf:
@rgodfrey Need real-time edits :) I  had not considered the quota delete case for the second alternative, good point.

I think the first alternative is fine then.

lulf:
@rgodfrey Updated the design doc with the changes

lulf:
Closing PR due to inactivity. Consider implementing this using admission hooks instead.

PR #1391, opened 2018-06-20
Allow role deploying keycloak resources to be parameterized

lulf:


kornys:
@lulf looks like smoke tests fails with this PR, I will look at it

lulf:
@kornys its because of a missing external-service.yaml file i removed, which is needed after all

kornys:
@lulf aha ok

lulf:
@kornys Please have another look

PR #1401, opened 2018-06-22
added better logging when waitForNReplicas fail

tkratky:


tkratky:
should help with debugging of AnycastTest testScaleRouterAutomatically failures
could provide more info about qdrouterd pods in address-space

PR #1407, opened 2018-06-27
Print more info about user credentials when binding is created

lulf:
@kornys FYI

### DIFF
--- a/service-broker/src/main/java/io/enmasse/osb/api/bind/OSBBindingService.java
+++ b/service-broker/src/main/java/io/enmasse/osb/api/bind/OSBBindingService.java
@@ -78,6 +78,7 @@ public class OSBBindingService extends OSBServiceBase {
             this.random.nextBytes(passwordBytes);
             String password = Base64.getEncoder().encodeToString(passwordBytes);
 
+            log.info("User: {}, password; {}", username, password);
             UserRepresentation userRep = keycloakClient.createUser(addressSpace.getAnnotation(AnnotationKeys.REALM_NAME), username, password);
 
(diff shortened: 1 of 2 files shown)


kornys:
@lulf /home/travis/build/EnMasseProject/enmasse/service-broker/src/main/java/io/enmasse/osb/keycloak/KeycloakInstance.java:[21,63] non-static method getClass()

kornys:
@lulf I restarted job again, I hope that in this build cert test will fail and we will see what was wrong.

service-broker/src/main/java/io/enmasse/osb/api/bind/OSBBindingService.java
```diff
@@ -78,6 +78,7 @@ public Response bindServiceInstance(@Context SecurityContext securityContext, @P
             this.random.nextBytes(passwordBytes);
             String password = Base64.getEncoder().encodeToString(passwordBytes);
 
+            log.info("User: {}, password; {}", username, password);
```
rgodfrey: I know that OpenShift secrets are not exactly secure, but I still think we should avoid logging passwords

kornys: @rgodfrey I think @lulf log this info only in for debugging this issue :) It will be removed after debugging.

lulf: @rgodfrey Yes, this is not intended to be merged.

kornys:
@lulf I run several times all tests in this branch and test still not fail :D

PR #1408, opened 2018-06-27
Enable fixed cli-rhea test and testRouterScale

kornys:
@lulf this issue was fixed in cli-rhea and published in 2.1.4, so I have enabled test and it passed.


kornys:
@lulf I will add enabling testScaleRouterAutomatically in this branch and try to debug.

kornys:
@lulf could you please approve it again? Problem with testRouterScale was with timeout for waiting and querying for replicas. Anyway sometimes when qdrouter pod is terminating by scale down it write error "The container metrics did not stop cleanly when terminated (exit code 137)."

PR #1415, opened 2018-07-02
UPD address-api-client to check expected status code

tkratky:


tkratky:
@lulf shouldn't status code on delete operation(s) be 204 instead of 200? because response on delete operation is empty body
` DELETE-address: path /apis/enmasse.io/v1alpha1/namespaces/enmasseci/addressspaces/brokered-shared-0/addresses/brokered-shared-0.brokeredqueuea
11:14:49,604 [vert.x-eventloop-thread-0] ERROR io.enmasse.systemtest.apiclients.ApiClient - expected-code: 204, response-code: 200, body: null`

PR #1419, opened 2018-07-03
Add scheduling affinity towards nodes labeled enmasse and support storage class parameter

lulf:


lulf:
Closing this until an actual use case is presented

PR #1438, opened 2018-07-18
separate in and out traffic between router and broker 

grs:
… over different connections (related to issue-1431)


grs:
@lulf or @rgodfrey could someone re-review this? I believe the previous issue is now fixed.

Issue #349, opened 2017-10-11
Jolokia agent for artemis (maybe some others) should be configured


gordillo-ramon-redhat:
See for example, https://github.com/jboss-openshift/cct_module/blob/master/os-java-jolokia/added/jolokia-opts.

Otherwise, java console in openshift will not work. I am currently receiving the following error:

Error: 'tls: oversized record received with length 20527'
Trying to reach: 'https://10.128.0.16:8161/jolokia/?maxDepth=7&maxCollectionSize=500&ignoreErrors=true&canonicalNaming=false'

Because openshift console proxy tries to use https, but it is configured by default as http (some other parameters are also needed).

lulf:
I had this working a while ago, but I didn't move through with enabling https yet, because the hawkular-openshift-agent

   * Needs to be configured with a client certificate trusted by the address space
   * Needs to be configured to trust the address space CA and have hostname verification turned off

That being said, the fact that we're using Artemis as a broker for queues in the standard address space should not be exposed to the end user. So I view this more as a convenience for the messaging service operator rather than something a tenant of the messaging service would use.

gordillo-ramon-redhat:
I mean it for developer debugging, as it is the way to get JMX info from the java pods.

Other than that, it is very strange to have the "open java console" link in ocp, and get an error when we got there.

lulf:
@vbusch have you been able to enable the open shift console for Artemis? Is this something we can reuse here?

ctron:
I think the OpenShift Jolokia console won't work the way it is set up currently. There is a port named "jolokia" and the "Open Java Console" appears in the OpenShift Web UI.

However then the connection fails and I think it may be due to the fact that the Artemis broker deploys this to "/console/jolokia" instead of "/jolokia" as expected by the OpenShift Web UI.

vbusch:
The broker team has been working on a plugin to get jolokia working with Artemis.  It is almost ready to be integrated https://bugzilla.redhat.com/show_bug.cgi?id=1610592

PR #1456, opened 2018-08-01
Enable Jolokia Agent (#349)

ctron:
This enables the Jolokia agent by properly deploying to an address
that the OpenShift embedded hawt.io web console understands.

### DIFF
--- a/artemis/Makefile
+++ b/artemis/Makefile
@@ -30,6 +30,10 @@ build_tar:
 	mkdir -p $(ARTIFACT_BASE)/jmx_exporter
 	cp -f jmx_exporter/target/lib/jmx_prometheus_javaagent-0.1.0.jar $(ARTIFACT_BASE)/jmx_exporter/
 
+	mkdir -p $(ARTIFACT_BASE)/jolokia-agent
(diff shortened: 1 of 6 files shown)


lulf:
@ctron One second thought... Does this expose jolokia without any authentication?

lulf:
@ctron I guess this means we'll be running 2 jolokia endpoints with this change? Is there a way not to do that? :)

k-wall:
@vbusch was saying there is an upstream change (to Artemis?) to fix the console link from openshift console.  I think she was doubting if this change was required.

vbusch:
I have been looking into the changes that have been made so far for the broker image.  There are several issues created.  https://issues.jboss.org/browse/ENTMQBR-1575 has the best details.
The console is now at the endpoint expected by openshift (with a few bugs).
Roddie has another update in progress so that the avaagent can access the mbeans with rbac restrictions.

lulf:
@vbusch @k-wall Thanks for providing info. We should perhaps hold off this then.

ALso sounds like the mbeans with rbac restrictions could be useful for us in order to have the prometheus agent not use JMX for collecting metrics.

lulf:
Closing for now, as jolokia is already included in Artemis but needs to be configured to use the expected path of the openshift console.

PR #1460, opened 2018-08-06
New checkReachableWebPage method

RH-sdavey:
* New checkReachableWebPage Method
    * Refactor of existing selenium web system tests to use new method


systemtests/src/main/java/io/enmasse/systemtest/selenium/page/OpenshiftWebPage.java
```diff
@@ -174,6 +174,7 @@ public void openOpenshiftPage() throws Exception {
         if (!waitUntilConsolePage()) {
             throw new IllegalStateException("Openshift console not loaded");
         }
+        checkReachableWebPage();
```
tkratky: In this case you can probably use content of waitUntilConsolePage (3 rows above), because it seems to me like this method was implemented for the same purpose (somewhen in past). But, if new check from checkReachableWebPage is for some reason better, then condition from row 174 can be removed. So this is up to you :)

systemtests/src/test/java/io/enmasse/systemtest/common/catalog/ServiceCatalogWebTest.java
```diff
@@ -87,6 +88,7 @@ void testProvisionAddressSpaceStandard() throws Exception {
         provisionedServices.put(namespace, standard);
         OpenshiftWebPage ocPage = new OpenshiftWebPage(selenium, addressApiClient, getOCConsoleRoute(), developer);
         ocPage.openOpenshiftPage();
+        ocPage.checkReachableWebPage();
```
tkratky: openOpenshiftPage() already contains execution of checkReachableWebPage() method, so this is not necessary. This additional check can be also removed from next tests (in commits bellow)

systemtests/src/main/java/io/enmasse/systemtest/selenium/page/OpenshiftWebPage.java
RH-sdavey: line 177 removed, I think original check is better

systemtests/src/test/java/io/enmasse/systemtest/common/catalog/ServiceCatalogWebTest.java
RH-sdavey: damn I forgot to remove them before commiting! removed now

RH-sdavey:
@tkratky @kornys Updated based on your comments, sorry I forgot to remove the call from tests before pushing...  and Ive checked all files for formatting now, thanks

systemtests/src/test/java/io/enmasse/systemtest/common/catalog/ServiceCatalogWebTest.java
```diff
@@ -74,7 +74,6 @@ void testProvisionAddressSpaceBrokered() throws Exception {
         String namespace = getUserProjectName(brokered);
         provisionedServices.put(namespace, brokered);
         OpenshiftWebPage ocPage = new OpenshiftWebPage(selenium, addressApiClient, getOCConsoleRoute(), developer);
-        ocPage.openOpenshiftPage();
```
tkratky: seems to me like this row shouldn't be removed

RH-sdavey: :/ youre right man, my mistake, added again.

tkratky:
let's wait until Jenkins PR run will be green :)

PR #1461, opened 2018-08-07
Allow the name that is displayed in the console, to be set/overridden during the build

vbusch:
- Can set the name that appears in the console using -Dapplication.display.name for maven or by setting the env_variable PROJECT_DISPLAY_NAME for make.
- Defaults to EnMasse

Signed-off-by: Vanessa <vbusch@localhost.localdomain>

### DIFF
--- a/agent/pom.xml
+++ b/agent/pom.xml
@@ -9,7 +9,28 @@
   <modelVersion>4.0.0</modelVersion>
   <groupId>io.enmasse</groupId>
   <artifactId>agent</artifactId>
+  <properties>
+    <application.display.name>EnMasse</application.display.name>
+  </properties>
   <build>
+    <resources>
+      <resource>
+        <directory>www</directory>
+        <targetPath>${project.basedir}/target/www</targetPath>
+        <excludes>
+          <exclude>index.html</exclude>
+        </excludes>
+        <filtering>false</filtering>
+      </resource>
+      <resource>
+        <directory>www</directory>
+        <targetPath>${project.basedir}/target/www</targetPath>
+        <includes>
+          <include>index.html</include>
+        </includes>
+        <filtering>true</filtering>
+      </resource>
+    </resources>  
     <plugins>
       <plugin>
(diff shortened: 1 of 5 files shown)


agent/pom.xml
```diff
@@ -9,7 +9,28 @@
...
+      </resource>
+      <resource>
+        <directory>www</directory>
+        <targetPath>${project.basedir}/target/www</targetPath>
+        <includes>
+          <include>index.html</include>
+        </includes>
+        <filtering>true</filtering>
+      </resource>
+    </resources>
```
grs: Am I right in thinking this now copies everything to target, from which the image is then built? Is that essential? Would it be possible to do filtering with the output in the same directory?

vbusch: Yes, www now appears in the target directory.  When the artifact is built, only 1 copy of www will appear in the artifact (just like before).
I would prefer not to have the build modifying the copy of index.html that is checked into git.  It would increase the odds of the build modifying the file, and someone checking it in.  And the file would be dirty after the production jenkins builds.  Neither reason is a show stopper.
Options:
1) move index.html into the src dir, and have the build copy it to www, with it in .gitignore
2) have maven call sed to replace the brand-alt in the index.html

grs: I agree we don't want to modify the file that is checked in to git (and actually doing the filtering as part of packaging is entirely fine). At present, the console server can be run from the working directory which is very useful (not having to package up everything for every edit of html). Now that the help and tooltips are written direct to target, that doesn't work so smoothly (I think). Could the help and tooltips be put into www before the copying of resources?

vbusch: I have updated the commit, to continue to copy the help, tooltips, and icon to www.

vbusch:
Removed reference to jboss jira.  The jira does reference this PR.
Should I create a github issue?  Downstream is the one to benefit from this change.

lulf:
I think that’s fine for now, it’s not a strict requirement, but generally changes upstream are reflected by upstream issues.

PR #1472, opened 2018-08-09
Use UUID only for cluster id and add address as annotation

### LINKED ISSUE #1449: Invalid PVC names created when using Gluster
When using GlusterFS with auto provisioning, EnMasse might create an invalid PVC name:

From the OpenShift log:

> Failed to provision volume with StorageClass "glusterfs-storage": create volume error: failed to create endpoint/service error creating service: Service "glusterfs-dynamic-broker-data-eventdefa-967d3474-8ae1-3a59-a7d1-5dcc90396b1f-0" is invalid: metadata.name: Invalid value: "glusterfs-dynamic-broker-data-eventdefa-967d3474-8ae1-3a59-a7d1-5dcc90396b1f-0": must be no more than 63 characters

lulf:
Fixes #1449


lulf:
@grs Please review again. I had to add a [a-z] prefix to get a valid service name for the statefulset. So I chose 'br'.

PR #1490, opened 2018-08-14
Only offer SASL mechanisms which can be supported

### LINKED ISSUE #1489: Keycloak sasl plugin should not offer mechanisms it can't support
Currently the Keycloak sasl plugin will offer all mechanisms, including mechanisms offered only by password hashing algorithms not used by any enmasse-created realm.  The plugin should only offer SASL mechanisms that are supported by at least one realm that enmasse has created.

rgodfrey:
This closes #1489

### DIFF
--- a/keycloak-plugin/sasl-plugin/src/main/java/io/enmasse/keycloak/spi/SaslAuthenticator.java
+++ b/keycloak-plugin/sasl-plugin/src/main/java/io/enmasse/keycloak/spi/SaslAuthenticator.java
@@ -27,44 +27,46 @@ import io.vertx.proton.ProtonConnection;
 import io.vertx.proton.sasl.ProtonSaslAuthenticator;
 import org.apache.qpid.proton.engine.Sasl;
 import org.apache.qpid.proton.engine.Transport;
+import org.jboss.logging.Logger;
 import org.keycloak.Config;
 import org.keycloak.credential.hash.Pbkdf2PasswordHashProviderFactory;
 import org.keycloak.credential.hash.Pbkdf2Sha256PasswordHashProviderFactory;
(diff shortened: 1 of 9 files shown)


keycloak-plugin/sasl-plugin/src/main/java/io/enmasse/keycloak/spi/SaslAuthenticator.java
```diff
@@ -82,11 +84,41 @@ public void init(final NetSocket socket,
...
+
+    private Set<String> getPasswordHashAlgorithms() {
+        Set<String> hashAlgos = new HashSet<>();
+        KeycloakSession keycloakSession = keycloakSessionFactory.create();
+        KeycloakTransactionManager transactionManager = keycloakSession.getTransactionManager();
+        transactionManager.begin();
+        try {
+            List<RealmModel> realms = keycloakSession.realms().getRealms();
+            for(RealmModel realm : realms) {
+                if(realm.getAttribute("enmasse-realm",Boolean.FALSE)) {
```
calohmn: @rgodfrey In case of an external Keycloak being used as authentication service, this attribute check now adds the requirement of having the "enmasse-realm" property in the manually created realm.

In case this property hasn't been set (and `hashAlgos` is empty), would it make sense to let `getValidMechanisms` return at least the mechanisms, that don't really check the hashAlgo (i.e. PlainSaslServerMechanism, XOAUTH2SaslServerMechanism)?
If not, I think it would be good to log a WARN message that no realm with this property exists and therefore authentication will fail.

rgodfrey: @calohmn yeah, it would probably make sense to add such always available mechanisms, I'll raise an issue for that and put it in the next release.

PR #1495, opened 2018-08-15
Added local openshift cluster files to gitignore

robshelly:


kornys:
@robshelly this folder is created when you use oc 3.10, right?

robshelly:
@kornys Yeah it's created in the working directory by `oc cluster up` when using Openshift 3.10.

It appears to store cluster config there so that you can run `cluster down `and then `cluster up` again and still have your projects.

PR #1502, opened 2018-08-15
MQTT - For clean sessions, close and unsubscribe any subscriptions on disconnect

### LINKED ISSUE #1499: Closed "cleanSession" true session continues to accumulate messages matching the subscription
If an MQTT client connects to the MQTT gateway and then closes the session (either with a `DISCONNECT`, or an abruptly closed TCP/IP connection), without first sending an `UNSUBSCRIBE`, the subscription queue remains on the Broker.  It continues to accumulate matching publications.

If the client reconnects with clean session `true`, the act of connecting discards the existing state, so the accumulated messages are not seen by the client.   If however, the client reconnects with clean session `false`, it will receive the accumulated messages, in violation of the specification.  This latter part was reported in https://github.com/EnMasseProject/enmasse/issues/1464 by @jsticha.

k-wall:
Partially fixes #1499

### DIFF
--- a/mqtt-gateway/src/main/java/enmasse/mqtt/AmqpBridge.java
+++ b/mqtt-gateway/src/main/java/enmasse/mqtt/AmqpBridge.java
@@ -274,18 +274,35 @@ public class AmqpBridge {
      * Close the bridge with all related attached links and connection to AMQP services
      */
     public void close() {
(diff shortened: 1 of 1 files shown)


k-wall:
Superseded by PR https://github.com/EnMasseProject/enmasse/pull/1514

PR #1504, opened 2018-08-16
Messsaging tests inside cluster

kornys:
This PR contains tests and system for testing messaging inside openshift cluster. 
@lulf If you approve it and PR will be green, I would like to move project with app and dockerfile under enmasseproject space in github and image under docker.io/enmasseproject.


kornys:
@lulf so I thought that this is clear way to easy modify it without rebuilding java, do you think that constructing resources in java code is better?

lulf:
@kornys I'm a bit divided on the matter... For EnMasse resources in general, its nice to be able to edit the yaml after a version has been released to patch bugs since you just have docker images.

For systemtests however, I'm not sure you'll be in that position, since you will always be able to modify the test code?

kornys:
@lulf that makes sence, let me change it from yaml to java.

PR #1507, opened 2018-08-17
Custom resources User API

lulf:
This is an initial draft on the format for creating users as custom resources and hiding the Keycloak UI for the standard address space.

The proposal is intended to solve the immediate needs to reduce dependency on Keycloak UI, but may be extended to support groups or other types of credentials or authorization policies.


documentation/design_docs/design/user-resource.adoc
```diff
@@ -0,0 +1,103 @@
...
+    authentication:
+      type: federated
+      provider: openshift
+      openShiftUserId: openshift user id
+      openShiftUserName: openshift user name
+    authorization:
+    - operations: [ "create, "delete", "send", "receive", "view", "*" ]
+      addresses: [ "queue1", "queue2", "*", "queue*" ]
+```
+
```
rgodfrey: I think it would be better to have the authorization keyed by the address/address pattern and then have the operations allowed on that address defined per address

```
authorization:
  - addresses: ["queue1", "queue2"]
    operations: ["send", "receive"]
  - addresses: ["*"]
    operations: ["view"]
  ```

Comment (lulf) [reply documentation/design_docs/design/user-resource.adoc]: Not sure i understand completely. There is no key here, just arrays. There are not connection between the order of items in the operations or the addresses array.

**Comment by grs** [comment]
How do you express that user a can send to address foo and receive from address bar (but not vice versa)?

**Comment by grs** [comment]
Would having passwords in secrets not be better than having them in the clear in the yaml used for create?

**Comment by kornys** [comment]
@grs something like that I guess

```
authorization:
  - addresses: ["foo"]
    operations: ["send"]
  - addresses: ["bar"]
    operations: ["receive"]
```

**Comment by lulf** [comment]
@grs 

‘’’
authorization:
  - addresses: ["foo"]
    operations: ["send"]
  - addresses: ["bar]
    operations: ["receive"]
‘’’

**Comment by lulf** [comment]
> Would having passwords in secrets not be better than having them in the clear in the yaml used for create?

@grs 

From a security perspective, the difference is that in secrets you base64 encode the password. Maybe we should do that as well.

The issue with secrets, is that they would typically be created by the tenant, whereas the api-server would not have permissions to read that secret in the tenants namespace, so the tenant would have to grant EnMasse permissions to read it. It just seemed a lot more complex from the user POV without bringing any security benefits.

On the other hand, we can perhaps expose this way through a 'type: secret', like this:

```
authentication:
  type: secret
  secret:
    name: mysecret
    namespace: myproject
```

**Comment by lulf** [comment]
@grs @rgodfrey could you have another look at the design doc: https://github.com/EnMasseProject/enmasse/blob/c135ceb7469a004a0dd1ff5abd4b9c277e7173e0/documentation/design_docs/design/user-resource.adoc

**Comment by lulf** [comment]
@jenmalloy I've written some initial documentation on creating/deleting users.

**Comment by grs** [comment]
I'm not entirely convinced by the 'why'.

For the what/how, one point to clarify is that support for wildcards is currently limited (in router) to a single '*' at the end of the pattern (e.g. queue-* is ok, queue-*-foo is not, nor is *-queue). 

Are federated identity type users only able to use the console? If so that might be worth clarifying.

(Another question is whether/how passwords can be chosen/changed by users. With keycloak there is an option to require a new password to be created on first login, though that only works if logging into the console at present. Not arguing this is required behaviour, just pointing out it is something that keycloak offers as an option.)

**Comment by k-wall** [comment]
A couple of high level comments/questions:

1. I think this proposal gives EnMasse the ability to manage its own local database of the users their credentials and authorisations.  This reminds me of the tomcat-users.xml file of Apache Tomcat.    Is the use-case for this feature standalone developer environments/small organisations without a federated identity system?

2. The proposal does not make it clear the local database options is an alternative or something that could sit alongside the use of a federated identity system.  Does the user need to make a choice at deployment time - chose the local database option or a federated source? 

A clear identity management strategy is key to large organisation, so I think there should be the ability to turn off a local database.

3. It is not clear to me how `create`/`update`/`delete` work when using a federated source.

Does the `create` gesture just merely link the user, which must already exist in the federated source otherwise the operation will fail? Does it also mean that if a new user joins an organisation and is provisioned centrally within the federated system, that there needs to be an extra step of something making this API call before the user is ready to use say the EnMasse console?   

Or is the idea that `create` gesture actually causes the record to be created in the federated system itself.  So in the case of keycloak, it would serve as a more convenient API?  If this is the case, how would the API degrade for identity systems which did not support the CRUD of users in this way?

4. What permissioning model would govern the creation/deletion/update of users?  I need to be able to change my own password, but I probably shouldn't be able to grant myself access to additional resources.   

In regulated organisations there are often tight controls around system access.  It needs to be possible for an identity maintainer role to be segregated from other roles.

**Comment by lulf** [comment]
>  Is the use-case for this feature standalone developer environments/small organisations without a federated identity system?

Yes. The 'standard' auth service is the 'embedded' authentication service for EnMasse. This API will only work against that auth service. Other auth services supported are 'none' and 'external'. In the case of complex organizations, I think 'external' is supposed to cover that use case.

> Does the user need to make a choice at deployment time - chose the local database option or a federated source?

When you deploy EnMasse, you can deploy this 'standard' authentication service if you wish. When you create an Address Space, it can specify which authentication service to use. If the 'standard' authservice is deployed, and the auth service has not been specified in the address space, the 'standard' is used.

> It is not clear to me how create/update/delete work when using a federated source

Again, the API is only available when the standard authentication service is deployed, and currently only a user federated to OpenShift is supported in that configuration. It is not indended to cover more specialized use cases covered by the 'external' auth service. In this specific case, a 'create' creates the user in Keycloak ,and 'links' it to the federated identity provider (OpenShift) which is assumed to already exist.

> What permissioning model would govern the creation/deletion/update of users? I need to be able to change my own password, but I probably shouldn't be able to grant myself access to additional resources.

This is controlled through the OpenShift RBAC policies configured when deploying EnMasse. The service admin can dictate which OpenShift users is authorized to create/get/delete etc. Those users are isolated from those managed by the User API (unless explicitly 'linked' by creating a federated user back to OpenShift)

Again, I see the purpose of this API more along the lines of the tomcat-users.xml that can be managed as Kubernetes resources, and I the 'external' authentication service mechanism is where we should solve the more complex use cases.

**Comment by rgodfrey** [comment]
@k-wall @lulf 

> Is the use-case for this feature standalone developer environments/small organisations without a federated identity system?
>
> > Yes. The 'standard' auth service is the 'embedded' authentication service for EnMasse. This API will only work against that auth service. Other auth services supported are 'none' and 'external'. In the case of complex organizations, I think 'external' is supposed to cover that use case.

So an alternative approach would be for some way to add other identity providers which could be federated with (LDAP or whatever).  I think the key here is that the scenario is that where the authorization is being managed inside EnMasse/Keycloak

> > It is not clear to me how create/update/delete work when using a federated source

Creating would make "adding the linked identity with given permissions", deleting would be the removal of the link.  Updating would clearly just be for modifying permission.

 > > What permissioning model would govern the creation/deletion/update of users? I need to be able to change my own password, but I probably shouldn't be able to grant myself access to additional resources.

So I think we may be evolving to a model where the "OpenShift" user will either have a "view" or a "create/update/delete" role for the address space as a whole, with a clear separation between the operations "OpenShift" users and "messaging" users.

File: documentation/common/concept-user-model.adoc [hunk 2]
```diff
@@ -0,0 +1,62 @@
...
+----
+
+The following fields are required:
+
+* `metadata.name`
+* `metadata.namespace`
+* `spec.authentication`
+* `spec.authorization`
+
+The `spec.authentication` object defines how the user should be authenticated, whereas
```
jenmalloy: should be = is ? (try to avoid using "should" if possible)

documentation/common/concept-user-model.adoc
```diff
@@ -0,0 +1,62 @@
...
+== Authentication
+
+The supported values for the authentication type is `password` and `federated`. 
+
+NOTE: At present, `password` type users can be used for both applications and console users, whereas
+`federated` type users will only work when using the messaging console.
+
+=== Password authentication type
+
+For the `password` type, an additional field `password` must be set to a base64 encoded value of the
```
jenmalloy: base64-encoded

documentation/common/proc-deleting-users-cli.adoc
```diff
@@ -0,0 +1,41 @@
...
+Users can in {ProductName} can be deleted using standard command line tools.
+
+.Prerequisites
+
+* An address space must have been created.
+* A user must have been created.
+
+.Procedure
+
+. List current users:
```
jenmalloy: List the current users:

documentation/tenant/assembly-managing-users.adoc
```diff
@@ -0,0 +1,27 @@
...
+:context: managing-users
+// The `context` attribute enables module reuse. Every module's ID includes {context}, which ensures that the module has a unique ID even if it is reused multiple times in a guide.
+
+{ProductName} user management is supported when using the standard authentication service. Users can
+be managed using a kubernetes native REST API and custom resources using the command line tools.
+
+.Prerequisites
+
+* {ProductName} is installed with custom resources enabled
+* You have created an <<address_space>>
```
jenmalloy: Is <<address_space>> an xref? If so, better to use xref: syntax. 
Same with <<restapi-reference>> below

k-wall:
Looks good to me.  Sorry, I was missing the fact that this only targeted the standard address space (despite you having written just that in the introduction).

PR #1509, opened 2018-08-17
Design doc for infrastructure resources

lulf:


lulf:
@grs @rgodfrey @k-wall @vbusch @robshelly @kornys @RH-sdavey @calohmn 

This is a proposal for how the service admin can manage the messaging infrastructure using kubernetes resources, as well as perform upgrades of the messaging infrastructure by changing these.

Please have a look.

documentation/design_docs/design/infra-resource.adoc
```diff
@@ -0,0 +1,134 @@
...
+  broker:
+    addressFullPolicy: PAGE
+    resources:
+      memory: 2Gi
+      disk: 50Gi
+  admin:
+    resources:
+      memory: 512Mi
+```
+
```
rgodfrey: So, is it possible to have more than one instance of the same "kind" but with different names? e.g. "small-brokered-infra-0.21.2" and "large-brokered-infra-0.21.2" ?  what does the namespace here represent?

documentation/design_docs/design/infra-resource.adoc
```diff
@@ -0,0 +1,134 @@
...
+  template:
+    kind: StandardInfraTemplate
+    name: standard-infra-0.21.2
+    version: 0.21.2
+  authenticationService:
+    host: example.com
+    port: 5671
+    caBundle: // Base64 encoded CA cert
+```
+
```
rgodfrey: How many "Infra" definitions can you have within a single installation of EnMasse,  would all the templates be enumerated in a single "infra" ?

documentation/design_docs/design/infra-resource.adoc
```diff
@@ -0,0 +1,134 @@
...
+As a first step, AddressSpacePlans can still contain the 'defined-by' annotation which can refer to
+a StandardInfraTemplate or a BrokeredInfraTemplate. When an address-space is created, this can be
+the way to find which template is used for creating the `Infra` resource.
+
```
rgodfrey: Why do the AddressSpacePlans not reference a named template rather than the template kind?  (presumably this would require each template definition in the Infra to have its own name)

documentation/design_docs/design/infra-resource.adoc
lulf: The namespace would typically be the namespace where the EnMasse infra-controller is deployed I think, so 'enmasse-infra' or whatever.

Yes, you can have multiple instances with different names. This is to support multiple versions but also different configurations. They can be handled the same way, except different versions can have some different meanings in what schema is supported. The 'version' attribute is mainly to be able to specify what schema one can expect to follow.

documentation/design_docs/design/infra-resource.adoc
lulf: You can have as many as is needed to support the number of address spaces needed. I.e. the number of infra and address space instances would be 1:1 at first, and then 1:M when 1 Infra can support multiple AddressSpaces.

documentation/design_docs/design/infra-resource.adoc
lulf: So I was thinking a 'defined-by' would be 'standard-infra-0.21.2' for instance, the above comment was maybe unclear.

lulf:
@rgodfrey I've updated the doc to (hopefull) be clearer + some sequence diagrams. Best viewed here: https://github.com/lulf/enmasse/blob/infra-custom-resources/documentation/design_docs/design/infra-resource.adoc

rgodfrey:
A question on the interaction of address space plans and the infrastructure template resources - I presume that it will not be possible to change the infrastructure associated with the address space plan (which would need to somehow migrate a plan from one configuration of infrastructure to another)?  How do we manager this restriction?

lulf:
@rgodfrey 

> A question on the interaction of address space plans and the infrastructure template resources - I presume that it will not be possible to change the infrastructure associated with the address space plan (which would need to somehow migrate a plan from one configuration of infrastructure to another)? How do we manager this restriction?

I assume you mean how we prevent the 'defined-by' annotation on the address-space-plan to be changed? Assuming we migrate AddressSpacePlan to a CRD, it should be possible to handle that kind of restrictions using https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#admission-webhooks , which allow you to plugin that kind of policy on k8s resources.

lulf:
Created issue with proposal. Please add additional comments in #1585

PR #1524, opened 2018-08-21
Update scripts for deployment openshift 3.10

kornys:


Jenkinsfile-long
```diff
@@ -40,7 +40,7 @@ pipeline {
             steps {
                 timeout(time: 10, unit: 'MINUTES') {
                     sh './systemtests/scripts/setup-openshift.sh "systemtests"'
-                    sh 'sudo chmod -R 777 /var/lib/origin/openshift.local.config'
+                    sh 'sudo chmod -R 777 /var/lib/origin/ || true'
```
lulf: How can this command fail?

systemtests/scripts/test_func.sh
```diff
@@ -270,3 +271,29 @@ function check_if_ansible_ready() {
...
+
+function get_openshift_version() {
+    echo $(oc version | sed -nre '/^oc/s/^.*v(([0-9]+\.)*[0-9]+).*$/\1/p')
+}
+
+function get_kubeconfig_path() {
+    OC_39='/var/lib/origin/openshift.local.config/master/admin.kubeconfig'
+    OC_310='./openshift.local.clusterup/kube-apiserver/admin.kubeconfig'
+
+    if [[ $(get_openshift_version) == '3.9.0' ]]; then
```
lulf: Maybe match on major.minor, not major.minor.micro. Ideally we should track the latest micro version anyway, wdyt?

Jenkinsfile-long
kornys: So openshift 3.9 is by default installed to /var/lib/origin, but oc 3.10 is installed into jenkins workspace, which is cool because after tests folder is cleaned. So I leave there chmod when job is running with 3.9.

systemtests/scripts/test_func.sh
kornys: Makes sense, but do you think that openshift team will release some new micro version od 3.9? Anyway I will update it to match only major.minor, thanks for suggestion 👍

Jenkinsfile-long
kornys: @lulf so when job is running with oc 3.10, /var/lib/origin foler does not exists

Issue #1474, opened 2018-08-10
system-tests: create new webconsole test for address type "subscription"


tkratky:
Create new web console test for durable subscription. This test should create and delete address via web console. New test will be very similar to following test: https://github.com/EnMasseProject/enmasse/blob/master/systemtests/src/test/java/io/enmasse/systemtest/standard/web/FirefoxWebConsoleTest.java#L21

grs:
Only wrinkle is that creating a subscription type address requires choosing a topic that the subscription is on.

RH-sdavey:
https://github.com/EnMasseProject/enmasse/commit/1705f69e72a2b0b4409879fb66552395bc43ab66

PR #1526, opened 2018-08-21
 New systemtest for durable subscriptions in webui

RH-sdavey:
issue #1474

    * New test for creating/deleting durable subs in web console
    * New public enum addressType : SUBSCRIPTION
    * Update to systemtest/Destination.java

### DIFF
--- a/systemtests/src/main/java/io/enmasse/systemtest/Destination.java
+++ b/systemtests/src/main/java/io/enmasse/systemtest/Destination.java
@@ -104,10 +104,6 @@ public class Destination {
         return new Destination(address, SUBSCRIPTION, plan, Optional.of(topic));
     }
 
-    public static Destination subscription(String address) {
-        return new Destination(address, SUBSCRIPTION, "standard-subscription") ;
-    }
-
     public static boolean isQueue(Destination d) {
         return QUEUE.equals(d.type);
     }
@@ -126,6 +122,10 @@ public class Destination {
 
     public String getAddress() {
         return address;
+    }
+
+    public String getTopic() {
+        return topic;
     }
 
     public String getQualifiedSubscriptionAddress() {
--- a/systemtests/src/main/java/io/enmasse/systemtest/Destination.java
+++ b/systemtests/src/main/java/io/enmasse/systemtest/Destination.java
@@ -104,6 +104,10 @@ public class Destination {
         return new Destination(address, SUBSCRIPTION, plan, Optional.of(topic));
     }
 
+    public static Destination subscription(String address) {
+        return new Destination(address, SUBSCRIPTION, "standard-subscription") ;
+    }
+
     public static boolean isQueue(Destination d) {
         return QUEUE.equals(d.type);
     }
--- a/systemtests/src/main/java/io/enmasse/systemtest/selenium/page/ConsoleWebPage.java
+++ b/systemtests/src/main/java/io/enmasse/systemtest/selenium/page/ConsoleWebPage.java
@@ -13,6 +13,7 @@ import io.enmasse.systemtest.selenium.resources.*;
 import org.openqa.selenium.By;
 import org.openqa.selenium.WebElement;
 import org.openqa.selenium.support.ui.ExpectedConditions;
+import org.openqa.selenium.support.ui.Select;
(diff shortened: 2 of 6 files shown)


systemtests/src/test/java/io/enmasse/systemtest/bases/web/WebConsoleTest.java
```diff
@@ -74,6 +74,28 @@ protected void doTestCreateDeleteAddress(Destination... destinations) throws Exc
...
+                sharedAddressSpace, defaultCredentials);
+        consoleWebPage.openWebConsolePage();
+        consoleWebPage.openAddressesPageWebConsole();
+
+        for (Destination dest : destinations) {
+            //create topic
+            consoleWebPage.createAddressWebConsole(dest);
+
+            //create subscription
+            Destination subscription = Destination.subscription(dest.getType() + "-subscriber");
```
grs: I think the subscription Destination instance should reference the topic address i.e. dest. As it is, I believe you are creating all the subscriptions against the first topic in the drop down list.

RH-sdavey:
@grs Updated the destination instance to reference the correct destination name, and the combobox selection is updated also.  Thanks for review, if you see anything else please let me know, thanks

systemtests/src/main/java/io/enmasse/systemtest/selenium/page/ConsoleWebPage.java
```diff
@@ -584,8 +585,16 @@ public void createAddressWebConsole(Destination destination, boolean openConsole
...
         selenium.clickOnItem(selenium.getWebElement(() -> selenium.getDriver().findElement(By.id(destination.getType()))), "Radio button " + destination.getType());
 
-        WebElement nextButton = selenium.getWebElement(() -> selenium.getDriver().findElement(By.id("nextButton")));
+        //if address type is subscription, fill in the topic dropdown box
+        if (destination.getType().equals(AddressType.SUBSCRIPTION.toString())) {
+            log.info("Selecting topic to attach subscription to");
+            WebElement topicDropDown = selenium.getWebElement(() -> selenium.getDriver().findElement(By.className("combobox")));
+            selenium.clickOnItem(topicDropDown);
+            Select combobox = new Select(topicDropDown);
+            combobox.selectByVisibleText(destination.getAddress().replace("-subscriber", ""));
```
grs: Rather than extracting the name of the topic from the subscribers address, I would add a getTopic() method to Destination and then pass in a Destination with the topic set explicitly.

systemtests/src/main/java/io/enmasse/systemtest/Destination.java
```diff
@@ -104,6 +104,10 @@ public static Destination subscription(String address, String topic, String plan
         return new Destination(address, SUBSCRIPTION, plan, Optional.of(topic));
     }
 
+    public static Destination subscription(String address) {
+        return new Destination(address, SUBSCRIPTION, "standard-subscription") ;
+    }
+
```
grs: A subscription with no address is not useful.

systemtests/src/test/java/io/enmasse/systemtest/bases/web/WebConsoleTest.java
```diff
@@ -74,6 +74,28 @@ protected void doTestCreateDeleteAddress(Destination... destinations) throws Exc
...
+                sharedAddressSpace, defaultCredentials);
+        consoleWebPage.openWebConsolePage();
+        consoleWebPage.openAddressesPageWebConsole();
+
+        for (Destination dest : destinations) {
+            //create topic
+            consoleWebPage.createAddressWebConsole(dest);
+
+            //create subscription
+            Destination subscription = Destination.subscription(dest.getAddress() + "-subscriber");
```
grs: I would create a subscription destination with an address and the topic specified separately. E.g. 

```
Destination subscription = Destination.subscription(dest.getAddress() + "-subscriber", dest.getAddress(), "standard-subscription");
```

Then in the createAddressWebConsole() method (as commented above) you can use the topic directly.

kornys:
@RH-sdavey please wait until PR job will be completed before merge.

PR #1531, opened 2018-08-22
Refactor MQTT system tests to use Paho MqttClient API directly so that it is possible for tests to excert precise control over MQTT interaction

k-wall:
New tests added for issues 1499/1503 but marked as ignored for now


k-wall:
I retained the ideas behind the the original tests.   There are quite a few gaps in the tests, for instance pub/sub to hierarchal topic trees, subscriptions to > 1 topics, multiple subscribers etc.  I intend to add these later a separate task

PR #1537, opened 2018-08-24
UPD categorize_docker_logs - added check if dir is not empty

tkratky:


tkratky:
Yes, this happened on our bare metal machines for tck and marathon tests. That issue was/is related to https://github.com/EnMasseProject/enmasse/pull/1542

lulf:
@tkratky So i guess this PR would mask such an error instead of failing the build, so I'm wondering if its better to leave it in and ensure nodes are setup correctly? Or is there a use case for running systemtests without getting the logs?

tkratky:
I wouldn't say that build should fail when docker logs are not stored. Docker logs are only additional information for identification of some issue (logging configuration of docker do not affect SUT). 

categorize_docker_logs from test_func.sh could be also use for categorization of docker logs when we are running tests locally. In this case could be usefull to check if directory passed in parameter is not empty.

PR #1541, opened 2018-08-27
Use latest selenium java

kornys:


kornys:
@tkratky please I need another approve, I switched to stable version of selenum standalone containers instead of using the latest one, and  also use selenium containers without debug tools, which should save some memory and time.

tkratky:
@kornys and you don't need another approve, you need another review 🤣

Issue #1387, opened 2018-06-20
Attempt on delete of not existing address should cause error


tkratky:
Attempt on delete of not existing address via oc cli should cause error. Now it returns "address <address> deleted"

Example:
```
$ oc get addresses -a -n myproject
> NAME                                                                         AGE
crd-address-test-shared.mytopic-agent.b9a93fd8-e00f-6c4a-adb5-030d9632e4da   1m

$ oc delete addresses crd-address-test-shared.mytopic-agent -n myproject
> address "crd-address-test-shared.mytopic-agent" deleted

$ oc delete addresses crd-address-test-shared.blah -n myproject
> address "crd-address-test-shared.blah" deleted

$ oc get addresses -a -n myproject
> NAME                                                                         AGE
crd-address-test-shared.mytopic-agent.b9a93fd8-e00f-6c4a-adb5-030d9632e4da   1m
```

Issue #1365, opened 2018-06-14
API Server: PUT requests should create or replace resources


calohmn:
`PUT` requests on the API server currently don't create a resource if it doesn't exist already.

According to K8s-API conventions, it should do so though.

See https://github.com/kubernetes/community/blob/master/contributors/devel/api-conventions.md#verbs-on-resources.

calohmn:
Related: with a `PUT` request, the addressName from the request URL is not taken into account. I think it should be checked against `metadata.name` in the request JSON - or it should be used to set `metadata.name` if it isn't set.

I can create a PR here (I'll wait for #1350).

calohmn:
Some testing on minikube with K8s API server `PUT` requests on ConfigMaps and a CustomResource (and tests on OpenShift with ConfigMaps) have shown, that such requests on non-existing resources have all lead to 404 responses. 
So it seems Kubernetes itself doesn't stick to the API conventions in this regard and doesn't create resources with PUT requests. 
That means we probably better stick to the current behaviour (with the added object name in URL validation mentioned above).

PR #1544, opened 2018-08-27
PUT on non-existing address(space) now returns 404 (#1384)

### LINKED ISSUE #1384: API-Server: PUT on non-existing address returns 200
A PUT on
`/apis/enmasse.io/v1alpha1/namespaces/{namespace}/addresses/{addressName}`
or
`/apis/enmasse.io/v1alpha1/namespaces/{namespace}/addressspaces/{addressSpace}/addresses/{addressName}`
, where the given address doesn't exist, returns status 200 - without having done anything.

Comparing that to corresponding K8s API server requests suggests that 404 should be returned here instead (see #1365).

Same applies to a PUT on
`/apis/enmasse.io/v1alpha1/namespaces/{namespace}/addressspaces`.

calohmn:
This fixes #1384.

### DIFF
--- a/api-server/src/main/java/io/enmasse/api/v1/AddressApiHelper.java
+++ b/api-server/src/main/java/io/enmasse/api/v1/AddressApiHelper.java
@@ -103,7 +103,9 @@ public class AddressApiHelper {
         AddressSpace addressSpace = getAddressSpace(address.getNamespace(), addressSpaceId);
         validateAddress(addressSpace, address);
         AddressApi addressApi = addressSpaceApi.withAddressSpace(addressSpace);
-        addressApi.replaceAddress(address);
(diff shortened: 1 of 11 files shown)


lulf:
I'm wondering if it should instead be 'create if not exist'? I.e. PUT on a resource should be idempotent

calohmn:
From a general REST API perspective, I agree.
The idea of this change was to make the behaviour consistent with that of the K8s API server, though (see https://github.com/EnMasseProject/enmasse/issues/1365#issuecomment-397614333).

(And I'll fix the failing test.)

calohmn:
As the `Address[Space]Api.replaceAddress[Space]` methods return a boolean now, it would make sense to change the `deleteAddress[Space]` methods accordingly. I would do that in a new PR as part of a fix for #1387.

PR #1552, opened 2018-08-28
New tests for creating addr space under non admin user

kornys:


systemtests/src/main/java/io/enmasse/systemtest/apiclients/AddressApiClient.java
```diff
@@ -30,15 +30,29 @@
     protected static Logger log = CustomLogger.getLogger();
     private final int initRetry = 10;
     private final String schemaPath = "/apis/enmasse.io/v1alpha1/schema";
-    private final String addressSpacesPath;
-    private final String addressNestedPathPattern;
-    private final String addressResourcePath;
+    private String addressSpacesPath;
+    private String addressNestedPathPattern;
+    private String addressResourcePath;
```
lulf: Not sure I like this being mutable, I find it cleaner to add another private constructor instead of the initializeAddressApiClient

PR #1555, opened 2018-08-29
reduce waiting in agent tests

grs:


lulf:
@grs I don't see how the failing systemtest is related to this change, so perhaps we should merge it?

grs:
@lulf I don't see why it is failing yet either, but I suspect there is something in this change that is responsible. I don't see this test failing elsewhere. (It does pass for me locally however). I'll dig in a bit more to try and understand what is happening.

grs:
@lulf thanks!

PR #1563, opened 2018-08-30
Update systemtests to use user api

lulf:


systemtests/src/main/java/io/enmasse/systemtest/apiclients/UserApiClient.java
```diff
@@ -0,0 +1,148 @@
...
+        User user = new User()
+                .setUsername(credentials.getUsername())
+                .setPassword(credentials.getPassword())
+                .addAuthorization(new User.AuthorizationRule()
+                        .addAddress("*")
+                        .addOperation(User.Operation.SEND)
+                        .addOperation(User.Operation.RECEIVE)
+                        .addOperation(User.Operation.VIEW))
+                .addAuthorization(new User.AuthorizationRule()
+                        .addAddress("*")
```
lulf: @kornys This was intentional. Manage does not apply to individual addresses.

lulf: @kornys This was intentionally to be without addresses. Manage does not apply to individual addresses, and the api allows empty list of addresses to signal that kind of operation

lulf:
@kornys sorry for the duplicate comments, github on phone not fully working :)

kornys:
@lulf I see, and np :) Ive just used that for testing because now server response is still 500 but body or response is null. I think that we need to update userApi to return better error messages. I will continue with debugging tomorrow.

kornys:
@lulf I made change in keycloak-user-api to creating manage_# group for brokered when manage group is specified.

kornys:
@lulf I made change in keycloak-user-api to creating manage_# group for brokered when manage group is specified. And another changes in keycloak-controller, now keycloak-controller adding namespace to realm attributes.

kornys:
@lulf really good job 👍  there is only the last test failure, Im going to try it locally and check what is wrong.

kornys:
@lulf I cannot run this test locally, it still throws decode exception service unavailable

lulf:
@kornys Rebase completed, please have a look

kornys:
@lulf LGTM! anyway what with PUT issue? Merge this branch and continue with testing/fixin in my new branch user-api-tests?

kornys:
@lulf wow it's green :) so I will fix authz tests, push, and then merge this branch without CI, because wildcard tests are not in PR.

kornys:
@lulf I found an issue in authz wildcards, it is fixed, but I found another issue, but not sure if it fixed in user-api-test branch. tomorrow we can check it.

PR #1527, opened 2018-08-22
Refactor manual install docs

lulf:
* Use the term namespace which is a bit more precise in the context of
deploying singletenant
* Clarify prerequisites for deploying to a single namespace
* Fix references to manual steps so that Kubernetes and OpenShift steps
are aligned
* Delete unused old doc files

PR #1564, opened 2018-08-30
Build and validate documentation on build

lulf:


lulf:
@jenmalloy @k-wall This PR is based on #1527, so only the last commit requires review for this PR. (The reason is that #1527 has some fixes that this PR would fail if not applied).

.travis/check_docs.sh
```diff
@@ -0,0 +1,36 @@
...
+  if [ -n "$x" ]; then
+    echo "$description:"
+    echo "$x"
+    y=$(echo "$x" | wc -l)
+    ((fatal+=fatalness*y))
+  fi
+}
+
+# Check for latin abbrevs
+grep_check '[^[:alpha:]](e\.g\.|eg)[^[:alpha:]]' "Replace 'e.g'. with 'for example, '"
```
k-wall: The regexs fail to report errors if the problematic word is at line beginning (or line end).  The problem is masked if you happen to have already generated the HTML form. 

Separately the .travis directory seems like an odd location of the check_docs.sh, as the check is not bound to Travis.

lulf: @k-wall Both script and location is just copied from https://github.com/strimzi/strimzi-kafka-operator/tree/master/.travis, so I hoped to minimize the changes from that

lulf: @k-wall I can move the script to `scripts`. For the regexp... I'm inclined to say its ok that its not a catch-all, but a best effort thing. Unless you have quick solution in place to fix it :)

k-wall: It seems good enough.

documentation/service_admin/ref-ansible-config-settings.adoc
```diff
@@ -2,7 +2,7 @@
 //
 // assembly_.adoc
 
-[id='ref-ansible-config-settings-{context}']
+[id='ref-ansible-config-settings-messaging']
```
jenmalloy: I think this should be {context} (since it's not an assembly file)

documentation/getting_started/mqtt-client-openshift.adoc
```diff
@@ -1,5 +1,5 @@
-[[mqtt]]
-==== Connecting using MQTT
+[[mqtt-{context}]]
```
jenmalloy: [id=mqtt-{context}]

documentation/getting_started/deploy-procedure.adoc
```diff
@@ -9,7 +9,7 @@ Invoke the deployment script to deploy {ProductName}
 ./deploy.sh -m "https://{DeployHost}:8443" -n enmasse {DeployOptions} -t {ClusterManager}
 ....
 
-===== Using ansible
+==== Using ansible
```
jenmalloy: Ansible ?

documentation/common/configuring-using-cli.adoc
```diff
@@ -3,7 +3,7 @@
 == Configuring {ProductName} using the command line
 
 {ProductName} can be configured to support manipulating address spaces and addresses using the
-Kubernetes and OpenShift command line tools. See <<oc-register-apiservice>> for how to setup the API
+Kubernetes and OpenShift command line tools. See xref:register-api-service-okd[] for how to setup the API
```
jenmalloy: command-line tools (since it's an adjective here)

documentation/getting_started/mqtt-client-openshift.adoc
lulf: I've just removed this reference for now, as i think it would require a larger restructure of the docs due to this being used in multiple getting started contexts (and thus failing the build).

lulf:
@jenmalloy Please have another look

documentation/service_admin/con-address-plans.adoc
```diff
@@ -7,7 +7,7 @@
 
 Address plans specify the expected resource usage of a given address. The sum of the resource usage for all resource types determines the amount of infrastructure provisioned for an address space. A single router and broker pod has a maximum usage of one. If a new address requires additional resources and the resource consumption is within the address space plan limits, a new pod will be created automatically to handle the increased load.
 
-In the xref:con-address-space-plans-messaging[] section, the address space plan references two address plans: `small-queue` and `small-anycast`. These address plans are stored as `ConfigMaps` and are defined as follows:
+In the xref:con-address-space-plans-{context}[] section, the address space plan references two address plans: `small-queue` and `small-anycast`. These address plans are stored as `ConfigMaps` and are defined as follows:
```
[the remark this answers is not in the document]
lulf: That would resolve in an invalid link according to the checker.

PR #1583, opened 2018-09-05
User api tests

kornys:


kornys:
@lulf thanks for enabling wildcard tests 👍

kornys:
@lulf I found this https://github.com/keycloak/keycloak/blob/812e76c39b1e693e8f11e5549cca2c90631f372e/testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/admin/UserTest.java#L388 it looks like that we use bad search in createUser method in keycloak-user-api, Im going to change it and try it in PR.

PR #1584, opened 2018-09-05
Don't require certs mandatory for external auth service

lulf:


lulf:
We need to decide on a way that we can do a 'if cert, use router setting' logic to enable this. Right now, router is configured via a configmap, and it is desirable to not change the router image startup scripts to handle this.

One alternative I like is to create the configmap in the address-space-controller when creating the address space. The conditional logic could then be applied. Need to think some more about this.

lulf:
Closing the PR for now until we have decided what to do

PR #1586, opened 2018-09-05
Agent test fix

grs:


kornys:
@grs :( 6 failed test in one travis build

grs:
@kornys with latest changes I have now had 6 consecutive runs of both travis checks with all the npm tests passing (the second run, which was overnight, had a failure after the npm tests had all passed, due to some minikube issue). I think it is worth considering merging this. If there are further failures we can push new PRS for them. (I also ran more than 200 iterations locally without any failures, but the CI exposes different races)

PR #1596, opened 2018-09-06
Add the ability for Maven build time properties to be centralised in …

k-wall:
…a pom.properties file.

Provides a single convenient single location which can be customised for downstream offerings.


k-wall:
This change requires Maven 3.3.1 (April 2015)

Issue #1606, opened 2018-09-10
Topic receiver count purpose unclear


k-wall:
In the standard address space, if I create a topic and then a durable subscription, and then attach a non-durable subscription to the topic and observe the console, the topic's receiver count reports 1.  If you detach the non-durable subscriber, the receiver count drops to 0.

The purpose the topic's receiver count is unclear.  I think a user may expect it to be sum total of all the subscribers at any point.

![topicreceivercount](https://user-images.githubusercontent.com/18440250/45308135-aa8cd900-b518-11e8-99f8-9d6219bce57d.png)

PR #1610, opened 2018-09-11
include durable subscription receivers in count of topic receivers 

grs:
(see issue #1606)


grs:
@k-wall this change will include active receivers in the count for the topic as well as the subscription. Note that at present, it will not show up if the subscription exists but has no consumer. I.e. the receiver count is the number of current receiver links for the topic (including those on any durable subscriptions).

PR #1613, opened 2018-09-11
Username validation

kornys:


kornys:
@lulf lgtm! just question username with underscore is restricted too?

lulf:
@kornys For the .metadata.name underscores are not allowed. However, I think we should allow it for .spec.username. I will raise another PR to fix that.

PR #1615, opened 2018-09-11
Issue 1587 single namespace

lulf:


agent/bin/subserv.js
```diff
@@ -208,5 +208,10 @@ if (process.env.MESSAGING_SERVICE_HOST) {
     });
 }
 
-var pod_watcher = require('../lib/pod_watcher.js').watch('role=broker,addresstype=topic');
+var selector = 'role=broker,addresstype=topic';
+if (process.env.INFRA_UUID) {
+    selector += ",infraUuid=" + process.env.INFRA_UUID;
+}
```
grs: This could perhaps be moved into the pod_watcher module itself.

grs:
One other comment, I *think* the MESSAGING_SERVICE_HOST in bin/subserv will also need to be qualified (for mqtt to work)?

agent/lib/internal_address_source.js
```diff
@@ -280,29 +286,43 @@ function extract_plan_details (plan) {
...
-            list.push(plan);
-            return map;
-        }, {});
-        var results = [];
-        types.forEach(function (type) {
-            results.push({name:type, plans:by_type[type].map(extract_plan_details)});
+    var address_space_plan_options = myutils.merge({selector: 'type=address-space-plan'}, this.config);
+    var address_plan_options = myutils.merge({selector: 'type=address-plan'}, this.config);
+    var address_space_plan_name = this.address_space.plan;
+    return kubernetes.get('configmaps', address_space_plan_options).then(function (configmaps) {
```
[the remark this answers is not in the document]
lulf: @grs So changing this to 

```
kubernetes.get('configmaps/address-space-plan-' + address_space_plan_name, ...)
```

?

agent/lib/ragent.js
```diff
@@ -384,7 +384,7 @@ Ragent.prototype.listen = function (options) {
 }
 
 Ragent.prototype.subscribe_to_addresses = function (env) {
-    var address_source = new AddressSource('foo', env);
+    var address_source = new AddressSource({name: 'foo', uuid: '1234'}, env);
```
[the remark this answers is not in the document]
lulf: @grs To me it seems like env is actually options passed on to kubernetes.watch( in AddressSource() constructor, so I wasn't sure if that would break it. Wdyt?

lulf:
> One other comment, I _think_ the MESSAGING_SERVICE_HOST in bin/subserv will also need to be qualified (for mqtt to work)?

I went through the route of adding these as environemnt variables in the subserv deployment. Wdyt of this approach as to doing extra lookups in the code? One problem with doing it in the code is that it has to do env.INFRA_UUID.toUpperCase() and then assemble the environment variable. I thought it might be simple to fix it in the templates like now.

grs:
Ah, yes, that makes sense, sorry I missed it.

kornys:
@lulf In the last test run I saw that testQuotaLimitPooled failed on address status which should be in pending state but was active, so I think that credit limitation does not work correctly or agent wrongly updated status.

lulf:
@kornys Fixed, it was an actual bug in standard-controller that assumed it could only use the first address space plan available in its namespace. Now there are multiple :)

documentation/modules/proc-deploying-api-server.adoc
```diff
@@ -3,15 +3,16 @@
 // assembly-installing-manual-steps.adoc
 
 [id='deploying-api-server-{context}']
-= (Optional) Deploying the API server
+= Deploying the API server
 
-The API server provides a REST API for creating address spaces and addresses. It can also serve as
-a Kubernetes API server if it is registered as an APIService.
+The API server provides a REST API for creating address spaces and addresses. It can also serve as an aggregated API server if it is registered as an APIService.
```
jenmalloy: Is APIService a literal?

documentation/modules/proc-deploying-api-server.adoc
```diff
@@ -3,15 +3,16 @@
...
 
-NOTE: To install {ProductName} on {PlatformName}, you must have `cluster-admin` access to set up the
-required roles for delegating authentication to the Kubernetes master; otherwise, you are restricted
-to a single address space. For more information about how to deploy without `cluster-admin` access,
-which restricts {ProductName} to a single address space, see xref:deploying-single-namespace-{context}[].
+ifeval::["{PlatformName} == "OpenShift"]
+NOTE: To install {ProductName} API Server on {PlatformName}, you must have `cluster-admin` access to set up the
```
jenmalloy: Change NOTE: to .Prerequisites
(Current template guidance is to use the plural, even if just one prereq is specified

To install the {ProductName}

documentation/modules/proc-deploying-service-broker.adoc
```diff
@@ -7,7 +7,12 @@
 
 The service broker provides an implementation of the Open Service Broker API that integrates with the Kubernetes Service Catalog. The service broker requires the standard authentication service to be deployed.
 
-NOTE: To install {ProductName} on {PlatformName}, you must have `cluster-admin` access to set up the required roles for delegating authentication to the Kubernetes master; otherwise, you are restricted to a single address space. For more information about how to deploy without `cluster-admin` access, which restricts {ProductName} to a single address space, see xref:deploying-single-namespace-{context}[].
+ifeval::["{PlatformName} == "OpenShift"]
+NOTE: To install {ProductName} Service Broker on {PlatformName}, you must have `cluster-admin` access to set up the
```
jenmalloy: Same comments as above

documentation/modules/proc-installing-without-cluster-admin-privileges.adoc
```diff
@@ -0,0 +1,41 @@
...
+//
+// assembly-installing-manual-steps.adoc
+
+[id='installing-without-cluster-admin-privileges-{context}']
+= Installing {ProductName} without cluster-admin privileges
+
+.Prerequisite
+
+* At least one authentication service must be installed. For more information see xref:deploying-auth-services-{context}[].
+* Address Space Controller must be installed. For more information see xref:deploying-address-space-controller-{context}[]
```
jenmalloy: .Prerequisites

The Address Space Controller must be...

Issue #1591, opened 2018-09-05
Filtering addresses with star symbol in console giving wrong results


RH-sdavey:
When filtering addresses by symbol * (star ), all addresses from the address list disappear, however the results box show x results,

steps to reproduce:
in console, create 4 new topics with names:
*addr
addr\*addr
addr\*
addrwithnostars
and then filter addresses by * symbol

(for bonus points, you can remove the filter for * and add another one now for . and observe that all addresses remain even though they have no . in the names, its understood this may be due to some regex happening)

workaround:
escape the * when filtering produces the correct number of filtered addresses and results value

k-wall:
We need some clarity on exactly how wildcards are supposed to work.   What are the wildcard symbols?   Are there restrictions on where why are placed?  Are there different rules for the the different filters in the console? 

We should also make sure that they are documented.

grs:
At present the filter can be a regex (not a simple wildcard though, so .* will match everything * alone will not). This was mostly an accident of implementation rather than any particular choice, so it can be changed if desired.

vbusch:
We can give an error message as below.  

![screenshot from 2018-09-19 08-51-40](https://user-images.githubusercontent.com/1995405/45754404-99635c80-bbe9-11e8-92c7-9811a5df06dc.png)

Alternatively, I can highlight the input filter box in red, and give a message similar to
![screenshot from 2018-09-19 08-55-48](https://user-images.githubusercontent.com/1995405/45754554-f101c800-bbe9-11e8-8369-4bd803c1c873.png)
This second options requires extending (copying the patternfly filter components), into our project and modifying them.  The patternfly team, would consider adding the ability to do it for patternfly version 4, but not in version 3.

jenmalloy:
@vbusch IMO, that first version works fine. Slight rewording suggestion: "**abc:" is not valid. Only regular expressions are supported." Does that second sentence make sense? If not, just the first sentence then. (Just trying to give the user a clue as to why it's not valid)

RH-sdavey:
https://github.com/EnMasseProject/enmasse/pull/1746
regex alert working as expected

PR #1620, opened 2018-09-11
don't apply invalid filter(s) (helps with issue #1591)

grs:


grs:
@RH-sdavey this at least prevents the inconsistent count/listing. It doesn't indicate very clearly what the error is - as absurd as it seems I can't find an easy way to do that with the angular-patternfly toolbar component (@vbusch do you have any suggestions on somehow indicating the actual error more clearly?)

RH-sdavey:
@grs  I still see the same results after your changes here Gordon, not sure if Im doing something wrong but Ill ping you tomorrow if you wouldnt mind walking  me through and making sure Im testing it correctly. 
The code looks like it should be doing the trick just fine though.

PR #1622, opened 2018-09-12
Allow underscores in .spec.username field

lulf:
@kornys Should probably add a systemtest that sets different values for .metadata.name and .spec.username and uses underscores in .spec.username.


kornys:
@lulf ok will create one

lulf:
@kornys Do you think this can be merged? We can also keep it open until a systemtest has been added to verify.

Issue #1607, opened 2018-09-10
The dashboard pie chart "Messages" does not account for messages sent to durable subscriptions.


k-wall:
**Describe the bug**
The dashboard pie chart "Messages" which is labeled "Totals for all addresses" does not account for messages sent to durable subscriptions.

**To Reproduce**
1. Send one message to a queue
2. Send one message to a topic with a single durable subscriber
3. Observe the address tab and verify that the queue has 1 stored message and the topic and subscriber report 1.
4. Go to the Dashboard

**Expected behavior**
The pie chart ought to account the messages sent to the topic as well as those sent to queues so that the operator understands the server's resource that are utilised.   

We'd need to decide how to account for messages that have been fanned out to many topic subscribers.  Should each message be account for once or n times?

**Additional context**

![dashtopicmessagesnotaccountedfor](https://user-images.githubusercontent.com/18440250/45309872-a6fb5100-b51c-11e8-82c2-edbfbf2f907a.png)

PR #1633, opened 2018-09-13
set message_in/message_out for link routed addresses i.e. topics

grs:
fixes issue #1607

### DIFF
--- a/agent/lib/router_stats.js
+++ b/agent/lib/router_stats.js
@@ -117,10 +117,10 @@ function get_stats_for_address(stats, address) {
     return s;
 }
 
-function collect_by_address(links, stats, router, connections) {
+function collect_by_address(links, stats, router, connections, index) {
     for (var l in links) {
         var link = links[l];
-        if (link.linkType === 'endpoint' && link.owningAddr && connections[link.connectionId]) {
+        if (link.linkType === 'endpoint' && link.owningAddr && connections[link.connectionId + '-' + index]) {
             var address = clean_address(link.owningAddr);
             var counts = get_stats_for_address(stats, address);
             if (link.name.indexOf('qdlink.') !== 0) {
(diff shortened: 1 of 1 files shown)


k-wall:
I'm still seeing the problem.  With a single topic with a single durable subscription, the dashboard still shows me nothing. In addition, I see a Javascript stack trace (both Chrome and Safari).

```
angular.js:14199 TypeError: Cannot read property 'ingress' of undefined
    at saveLastLinkInfo (addresses_ctrl.js:57)
    at calcRates (addresses_ctrl.js:64)
    at addresses_ctrl.js:125
    at Array.forEach (<anonymous>)
    at addresses_ctrl.js:124
    at angular.js:19942
    at f (angular.js:6045)
    at angular.js:6324
(anonymous) @ angular.js:14199
(anonymous) @ angular.js:10707
(anonymous) @ angular.js:19945
f @ angular.js:6045
(anonymous) @ angular.js:6324
setTimeout (async)
k.defer @ angular.js:6322
e @ angular.js:19940
forceDeliveryRateUpdate @ addresses_ctrl.js:123
setTimeout (async)
on_update @ addresses_ctrl.js:156
(anonymous) @ addresses_ctrl.js:170
invoke @ angular.js:4771
R.instance @ angular.js:10592
n @ angular.js:9469
g @ angular.js:8810
(anonymous) @ angular.js:8690
(anonymous) @ angular-ui-router.min.js:7
(anonymous) @ angular.js:1259
sa @ angular.js:10152
n @ angular.js:9551
g @ angular.js:8810
(anonymous) @ angular.js:8690
(anonymous) @ angular.js:9047
l @ angular-ui-router.min.js:7
(anonymous) @ angular-ui-router.min.js:7
$broadcast @ angular.js:18320
(anonymous) @ angular-ui-router.min.js:7
(anonymous) @ angular.js:16696
$eval @ angular.js:17994
$digest @ angular.js:17808
$apply @ angular.js:18102
(anonymous) @ angular.js:26597
dispatch @ jquery.js:4435
r.handle @ jquery.js:4121

```

subsstat1
substats2

grs:
I believe what you are seeing is that when there are no active links for a topic, the message count drops to 0 as at that point the router will no longer have any stats for the associated addresses. This is true also of anycast and multicast addresses. For queues, because they use autolinks, there is always an active link unless the broker is restarted.

This may also not be desirable, but it is I think a distinct issue, affecting all but the queue address type.

grs:
While you have an active link on the topic, you should see messages showing up in the address list and also as part of the pie chart on the dashboard.

screenshot from 2018-09-18 20-28-40

but once you close the links only the queue messages will remain

screenshot from 2018-09-18 20-30-35

a, b, c and d are queue, topic, anycast and multicast respectively

screenshot from 2018-09-18 20-36-53

grs:
@k-wall you were right! I was using subscribers on all addresses when testing. The messages pie chart only counts messages out (not sure what the rationale was there). The messages out from the durable subscription were not being included. I have now updated to fix that as well.

PR #1651, opened 2018-09-18
Parameterise Agent to allow a kc_idp_hint query argument t…

k-wall:
…o be included in the redirection to the authorize endpoint. Allows intermediate keycloak login page to be skipped.


k-wall:
@vbusch @lulf : To use the feature added by this patch, the address space's agent deployment needs to be configured to set the default provider using an environment variable.  See the the patch below.  I don't think we want to define a default in EnMasse. Thoughts?

[0001-Default-to-openshift-provider.txt](https://github.com/EnMasseProject/enmasse/files/2425057/0001-Default-to-openshift-provider.txt)

lulf:
@k-wall In the case of running on Kubernetes, will the hint/setting cause things to break, since the identity provider is not installed in that case?  If it does not work with Kubernetes + standard authservice (without openshift id provider), I think we need to make it an optional env var that we can set as a template parameter in address-space-controller when creating the address space if running on openshift.

vbusch:
@k-wall for EnMasse we should add documentation for enabling it.  And yes, for downstream AMQ Online should have it enabled by default.

lulf:
I think it should be set if openshift && standard authservice is used when creating the address space, and  provide some hook to disable it if desired.

lulf:
@k-wall To elaborate, if the parameter needs override, the user would create the address space like 
this:

```
kind: AddressSpace
metadata:
  name: myspace
  annotations:
    enmasse.io/kc-idp-hint: value
```

address-space-controller/src/main/java/io/enmasse/controller/TemplateInfraResourceFactory.java
```diff
@@ -43,11 +50,20 @@ public TemplateInfraResourceFactory(Kubernetes kubernetes, SchemaProvider schema
...
 
+
         if (resourceDefinition != null && resourceDefinition.getTemplateName().isPresent()) {
-            Map<String, String> parameters = new HashMap<>();
             AuthenticationService authService = addressSpace.getAuthenticationService();
             AuthenticationServiceResolver authResolver = authResolverFactory.getResolver(authService.getType());
 
+            String kcIdpHint = null;
+            if (addressSpace.getAuthenticationServiceKcIdpHint() != null) {
+                kcIdpHint = addressSpace.getAuthenticationServiceKcIdpHint();
```
lulf: Using annotations, this would instead be:

if (addressSpace.getAnnotation(AnnotationKeys.KC_IDP_HINT) != null) {
   //
}

Then the additions in AddressSpace.java, AddressSpaceDeserializer etc. can be removed.

k-wall: Thanks for the good advice, that sounds much less intrusive.   I actually pushed this up to be my branch so soon.   It is not in a working state yet.

systemtests/src/main/java/io/enmasse/systemtest/AddressSpace.java
```diff
@@ -186,14 +192,27 @@ public void setInfraUuid(String infraUuid) {
...
+
     @Override
     public String toString() {
         StringBuilder addressSpaceString = new StringBuilder()
                 .append("{name=").append(name).append(",")
                 .append("namespace=").append(namespace).append(",")
                 .append("infraUuid=").append(infraUuid).append(",")
                 .append("type=").append(type.toString().toLowerCase()).append(",")
-                .append("plan=").append(plan);
+                .append("plan=").append(plan)
```
lulf: add a .append(",") there

PR #1655, opened 2018-09-19
 NEW(UserApiTest) new tests for UserApi common tests

RH-sdavey:
# testCreateUserHyphenUsername
          - Hyphens are allowed in the username created by UserApi
      # testCreateUserMustStartWithChar
          - User name must start with Char, if not reply of 400 BAD_REQUEST


RH-sdavey:
@lulf , yes I agree, all collapsed into one, and as it was quite cheap (copy and paste)  to create a few more similar test cases, and only adds seconds on to the systemtest time,  I threw some more on there aswell, feel free to check them. 
Thanks for review.

PR #1671, opened 2018-09-24
Remove old openshift data for openshift 3.10

kornys:


kornys:
@k-wall on it is my bad, I will add these lines there.

PR #1677, opened 2018-09-25
Use fully-fledged OpenShift oauthclient rather than the serviceaccount to simplify openshift login workflow

k-wall:


templates/address-space-controller/020-ConfigMap-address-space-definitions.yaml
```diff
@@ -359,6 +359,8 @@ data:
                 value: ${CONTROLLER_RESYNC_INTERVAL}
               - name: CHECK_INTERVAL
                 value: ${CONTROLLER_CHECK_INTERVAL}
+              - name: CONFIGMAP_WATCH_INTERVAL
```
lulf: This variable does not seem to be used.

k-wall: KeycloakController reads environment variable `CONFIGMAP_WATCH_INTERVAL` and defaults it to 5 minutes if not set.   I think what is missing is a ` parameter` defaulting its value in the template.

lulf: But this is the admin deployment, not the keycloak-controller deployment, which is located in templates/standard-authservice/050-Deployment-keycloak-controller.yaml

k-wall:
@lulf, @vbusch  One query. Is it our practice to expose all environment variables used by an application up to the deployment template?  If so, in addition to CONFIGMAP_WATCH_INTERVAL, we should also be  treating OAUTH_CLIENT_GRANT_METHOD and OAUTH_CLIENT_NAME in the same way.

lulf:
@k-wall the practice I’ve tried to follow (which might not be fully consistent) is to only expose the mandatory ones in the  deployment. Optional parameters should (eventually) be documented in the standard authservice documentation. Otherwise we have to put defaults twice, in the code and in the template.

One approach I’ve taken in the api-server deployment is to map some optional parameter from an optional config map(api-server-config) which I find is a nice way to expose the parameter in the deployment without setting a value.

Whatever we choose, we should try to use the same approach.

k-wall:
@lulf I think that approach make sense. I'd prefer to keep that refactoring separate, so I propose to commit this change as it stands.

PR #1679, opened 2018-09-25
Update make deploy target

lulf:


HACKING.md
```diff
@@ -44,7 +44,7 @@ This builds all modules including java.
 *Note*: If you are using OpenShift and 'oc cluster up', you can push images directly to the builtin registry
 by setting `DOCKER_ORG=myproject` and `DOCKER_REGISTRY=172.30.1.1:5000` instead.
 
-#### Deploying to an OpenShift instance assuming already logged in
+#### Deploying to an OpenShift instance assuming already logged in with cluster-admin permissions
 
     make deploy
```
grs: I'd be inclined to just point at the two templates (none and standard auth) and not bother with deploy. Is it possible to deploy from a working build using a kubernetes flavour other than openshift?

lulf: I agree. I will update this and remove the deploy target

PR #1682, opened 2018-09-25
Fix maven build stability warnings

k-wall:


systemtests/pom.xml
```diff
@@ -77,11 +77,6 @@
             <artifactId>junit-jupiter-engine</artifactId>
             <scope>test</scope>
         </dependency>
-        <dependency>
-            <groupId>org.junit.platform</groupId>
-            <artifactId>junit-platform-launcher</artifactId>
-            <scope>test</scope>
-        </dependency>
```
lulf: Before merging this, ensure systemtests are actually running even with this removed.

k-wall: junit-platform-launcher was imported twice, once at scope compile, once at scope test.  The scope of compile is broader than test, so there should be no functional aspect to the change.  I will confirm though before merging, like you suggest.

Issue #1588, opened 2018-09-05
[console] Address name is not created when starting or ending with .


RH-sdavey:
When creating any type of address in console, 
if the name begins **or** ends with .  ( fullstop / period ) then the address creation wizard finishes with no error output, but the address creation is seemingly never begun.
No address is created and output from `oc get -o json addresses` shows no address created either

RH-sdavey:
also noticed, just for topic addresses, if the topic name ends in \#  then the topic never gets created, stuck in a pending state. all other address type ending in \# are created fine.

vbusch:
Added a PR, to keep the 'next' button disabled, until the name is valid.  https://github.com/EnMasseProject/enmasse/pull/1720

RH-sdavey:
Certain chars ( including . ) are no longer allowed and next button is disabled if they are attempted to be used. Issue can be closed

PR #1686, opened 2018-09-25
prevent address being created through console with certain chars

grs:
within them, see issue #1588


grs:
@lulf actually, I think it would be better to create a separate issue for any desired server side validation. At present there is none, so said user could set plans that don't exist or aren't valid etc. Might even consider the case where config maps are created directly, and e.g. do validation in standard-controller, moving badly formed addresses straight to error or something. A separate issue would allow the scope and logic of desired checks to be more clearly specified. This then is just a fix for the [console] issue as marked.

RH-sdavey:
@grs yes the code changes lgtm, will approve this for now and create a few tests to move the jira to done asap. Thanks

PR #1699, opened 2018-09-26
Always update endpoints at start of controller loop

lulf:
This should fix the systemtest failures. The certificates are taking longer than necessary to create due to 'stale' address space objects going through the controller loop. This ensures that the address space 'object' is always updated to the 'desired' state.


kornys:
@lulf please replace “+” for strings with String.format or StringBuilder :-D

PR #1703, opened 2018-09-27
Correct unit of watch interval used to detect keycloak config map changes

k-wall:


k-wall:
I tested it myself. Confirmed that it is possible to turn the new watcher off by using a zero interval.

PR #1704, opened 2018-09-28
Rename paths in tarball

lulf:
* Rename `examples` to `install` and `examples/install` to `install/components`
* Replace myproject in documentation examples


calohmn:
It looks like there are two occurrences of "enmasse-latest/examples" left in HACKING.md.

PR #1705, opened 2018-09-28
Reorder master.adoc for messaging app developer to improve console he…

lulf:
…lp page

This way, the help page docs also include the section explaining what address spaces is, and the indentation is fixed so that the ToC makes more sense.


lulf:
Example table of contents:

toc

PR #1706, opened 2018-09-28
Refactor client connect checks

lulf:
* Reuse code to create client, check messages sent/recv
* Print exceptions and translate to true/false
* Close clients after use


kornys:
@lulf I see

lulf:
@kornys The currently failing tests doesnt seem to be related to this change.

Issue #1445, opened 2018-07-23
Confusing screen labels Messages In / Messages Out in address rows


k-wall:
In the EnMasse console in the address view we have two counts Messages In and Message Out displayed for each address.  The values displayed are actually delta values computed over the last 5 minutes. I don't think this is obvious for to the end-user.  I think the user would assume these are absolute values and will be surprised when they suddenly drop to zero.

I think the screen labels should be clarified.

vbusch:
They drop to zero after 15 minutes.  But if you refresh the page, the original numbers reappear.  

If you send new messages after 15 minutes, the counts only reflect the new messages, until the page is refreshed and then you see the total (of all messages).

grs:
For anycast and multicast messages the stats will reset when the last link for the address is detached, as the router only tracks stats while there are active links. For queues/topics this means any stat taken from the router(s) will be affected by a router restart also.

PR #1708, opened 2018-09-28
fix and improve messages in/out for addresses

grs:
Also change the description and add tooltip giving more info.

### DIFF
--- a/agent/www/components/addresses/addresses.html
+++ b/agent/www/components/addresses/addresses.html
@@ -46,17 +46,18 @@
         </div>
       </div>
       <div class="list-view-pf-additional-info">
-        <div class="list-view-pf-additional-info-item">
-          <span title="Messages sent to this address in last 5 minutes">
+        <div class="list-view-pf-additional-info-item-stacked">
+          <div class="list-view-pf-additional-info-item" title="Messages sent to this address in last 5 minutes">
             <span class="fa fa-sign-in"></span>
-            <strong>{{item.messages_in_delta}}</strong> Incoming Traffic
-          </span>
-        </div>
-        <div class="list-view-pf-additional-info-item">
-          <span title="Messages sent from this address in last 5 minutes">
-          <span class="fa fa-sign-out"></span>
-          <strong>{{item.messages_out_delta}}</strong> Outgoing Traffic
-          </span>
+            <strong>{{item.messages_in_delta}}</strong>  Messages In
+          </div>
+          <div class="list-view-pf-additional-info-item" title="Messages sent from this address in last 5 minutes">
+            <span class="fa fa-sign-out"></span>
+            <strong>{{item.messages_out_delta}}</strong>  Messages Out
+          </div>
+          <div class="legend-pf-text">
+            <em>(values over the last 5 minutes)</em>
+          </div>
         </div>
         <div class="list-view-pf-additional-info-item" ng-if="item.type !== 'subscription'">
           <span class="fa pficon-import"></span> <strong>{{item.senders}}</strong> Senders
--- a/agent/www/components/addresses/addresses.html
+++ b/agent/www/components/addresses/addresses.html
@@ -58,32 +58,6 @@
           <strong>{{item.messages_out_delta}}</strong> Outgoing Traffic
           </span>
         </div>
-        <!--
-        <div class="list-view-pf-additional-info-item" ng-if="item.type !== 'subscription'">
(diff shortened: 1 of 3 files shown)


agent/www/components/addresses/addresses.html
```diff
@@ -38,22 +38,44 @@
...
+            <strong>{{item.messages_in_delta}}</strong> Incoming Traffic
+          </span>
+        </div>
+        <div class="list-view-pf-additional-info-item">
+          <span title="Messages sent from this address in last 5 minutes">
+          <span class="fa fa-sign-out"></span>
+          <strong>{{item.messages_out_delta}}</strong> Outgoing Traffic
+          </span>
+        </div>
+        <!--
```
k-wall: Commented out code.

agent/www/components/addresses/addresses.html
```diff
@@ -38,22 +38,44 @@
         </div>
       </div>
       <div class="list-view-pf-additional-info">
+        <div class="list-view-pf-additional-info-item">
+          <span title="Messages sent to this address in last 5 minutes">
+            <span class="fa fa-sign-in"></span>
+            <strong>{{item.messages_in_delta}}</strong> Incoming Traffic
```
k-wall: I'm not sure about "Incoming Traffic". I think the user will still be confused.   The hover over helps, but the user would need to know to look for this. I think we need to find a way to display a more descriptive field label. Maybe we could use smaller text beneath the value "Message In - Last 5 min" etc.

I think also the change needs to be reflected on the Connections tab too.

agent/www/components/addresses/addresses.html
grs: Fixed

agent/www/components/addresses/addresses.html
k-wall: > I think the connections tab is different. There the count is the actual number of messages on that connection.

I hadn't appreciated that those were absolute values.  Thanks.

k-wall: @gsim - any thoughts on my comments on the label used on the address tab?

grs: @k-wall I don't have strong feelings about the text. I think the time period it covers is really a detail. Though useful to be able to find it out easily, in my view it does not need to be part of the main label; perhaps adding an info icon as the target for the mouseover would help inform users of the extra text available.

Apart for the time period, 'message in' was what was there before which I thought was reported as being unclear.

kornys:
@grs @k-wall I rebased this branch on current "fixed" master

k-wall:
#1445 intended to complain that screen labels imply an all-time total for messages in and messages out for the address.  I think the label does need to communicate the real nature of the value somehow.  The user shouldn't need to hover/click/readthedocs to find this out.

I was wondering about adding the text "over the last 5 minutes" below the two values, spanning both columns to reduce the screen clutter.

For me the label using the word Messages is better than Traffic.

grs:
@k-wall do you want to commit the text you feel is appropriate onto this PR then?

k-wall:
@grs - here's what my suggestion looks like.
stackmessageinoutwithlegend

grs:
@kwall to me the alignment looks a bit odd there, but I don't object if that is what is wanted, just an observation.

k-wall:
I've got a really poor eye for aesthetics of a UI.   Are you referring to the alignment of the tick box on the left hand side?

grs:
Yes, but also the stacked items themselves - to me - look odd in the flow. Again though, I'm not opposing the change, just commenting. It may well be that the extra clarity/directness outweighs any subjective opinions on aesthetics.

k-wall:
As discussed, I value the extra clarity/directness.  If we can think of a way of improving the presentation, then we can improve later.

RH-sdavey:
@k-wall  I have to agree with @grs a little with the aesthetics, the tick box on the left looks really out of sorts right now... 
brokeredui
compared to previous "cleaner" version
filter1
and the stacked stats boxes dont work so well imo ( it grates on me that the arrows both point the same way for messages in and messages out... ) , any chance to put them all inline with just the text "values over the last 5 minutes" underneath ( as it is now) 
or what about a little "popup" window with text displayed
 "all values displayed here are aggregated over the last 5 minutes" or something similar.
something like this example
hawtio
it could be shown just once on the first time the address tab is loaded (I suppose thats possible)

PR #1711, opened 2018-10-01
Fix clean old oc instance

kornys:


systemtests/scripts/test_func.sh
```diff
@@ -321,19 +321,17 @@ function clean_docker_images() {
...
-        sudo find ${OC_DATA_PATH} -type d -exec mountpoint --quiet {} \; -exec umount --types tmpfs {} \;
-        sudo rm -rf ${OC_DATA_PATH}
-        sudo rm -rf /var/log/containers/*
-        sudo rm -rf /var/log/pods/*
-    else
-        info "${OC_DATA_PATH} not found"
+        info "Cleaning OpenShift work directory openshift.local.clusterup"
+        OLD_OC_LOCATIONS=$(sudo find / -type d -name 'openshift.local.clusterup')
+        for i in ${OLD_OC_LOCATIONS}; do sudo find "${i}" -type d -exec mountpoint --quiet {} \; -exec umount --types tmpfs {} \;; done
+        sudo find / -type d -name 'openshift.local.clusterup' -exec rm -rf {} +
```
k-wall: The change scans the entire file system, twice, looking for directory openshift.local.clusterup. Once on 328 and again three lines later.  This might be slow so why repeat it.    It almost seems odd to need to scan from root in the first place.  Do we not restrict the tests to a working directory?

lulf: I would say looking in /home/cloud-user and /var/lib should be sufficient.

PR #1717, opened 2018-10-01
Copy all YAML files into built documentation

lulf:
* Move all YAML files referenced in documentation to documentation/common
* Copy *.yaml to html/resources

@jenmalloy I did not find from the contributing docs where these kind of files should be located, so I assumed common would be a good place as any.


jenmalloy:
@lulf, agreed, in the common folder makes the most sense to me too.

PR #1725, opened 2018-10-02
Add error message for incorrect filters on connections tab too.

k-wall:
Also refactored algorithm so that it does not rely on the invalid filter being on the end of the array for correct behaviour.

### DIFF
--- a/agent/www/components/addresses/addresses_ctrl.js
+++ b/agent/www/components/addresses/addresses_ctrl.js
@@ -218,26 +218,33 @@ angular.module('patternfly.toolbars').controller('ViewCtrl', ['$scope', '$timeou
           }
         };
 
-        function is_valid_filter(filter) {
-            if (filter.title === 'Name') {
-                try {
-                    $scope.notification.show_alert = false;
-
-                    var dummy = new RegExp(filter.value);
-                    return true;
-                } catch (error) {
-                    $scope.notification.show_alert = true;
-                    $scope.notification.alert_msg = '"'+ filter.value + '": is not valid.  Only regular expressions are supported.';
-                    return false;
-                }
-            } else {
-                return true;
-            }
-        }
-
         var filterChange = function (filters) {
             $scope.filtersText = "";
-            var valid = filters.filter(is_valid_filter);
+
+            $scope.notification.show_alert = false;
+            var messages = [];
+            var valid = filters.filter(function (filter) {
+                if (filter.title === 'Name') {
+                    try {
+                        $scope.notification.show_alert = false;
+
+                        var dummy = new RegExp(filter.value);
+                        return true;
+                    } catch (error) {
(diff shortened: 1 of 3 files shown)


agent/www/components/addresses/addresses_ctrl.js
```diff
@@ -214,26 +214,33 @@ angular.module('patternfly.toolbars').controller('ViewCtrl', ['$scope', '$timeou
...
-            if (filter.title === 'Name') {
-                try {
-                    $scope.notification.show_alert = false;
+        var filterChange = function (filters) {
+            $scope.filtersText = "";
 
-                    var dummy = new RegExp(filter.value);
+            $scope.notification.show_alert = false;
+            var messages = [];
+            var valid = filters.filter(function (filter) {
```
grs: Not a big issue, but was there a reason to inline the is_valid_filter function? (To me the flow is more readable as a separate, named function).

k-wall: @vbusch's previous change had changed `is_valid_filter` from a pure function to be one with a side effect (namely the manipulation of the notification).   For correct behaviour the 'bad' filter always needed to appear on the end of the list, otherwise the next good filter would have reset the notification set.   Currently, as new filters are pushed to the end of the list, this was a safe assumption, but it felt fragile to me: it was easy to imagine a innocuous change elsewhere breaking this code.

Inlining the function allowed me to push the responsibility for manipulating notification state to filterChange whilst still calling new RegExp() only once per filter.  I can refactor again to to reinstate the pure function and separately build up the validation messages if you prefer.

grs: No, thats a good explanation!

k-wall:
We'll need to reflect changes from #1740 in this patch, or vice versa.

k-wall:
@vbusch - I made the same changes you made to addresses to fix the issue @RH-sdavey reported ( #1736) in the connection tab.

agent/www/components/addresses/addresses_ctrl.js
```diff
@@ -218,26 +218,33 @@ angular.module('patternfly.toolbars').controller('ViewCtrl', ['$scope', '$timeou
...
+                if (filter.title === 'Name') {
+                    try {
+                        $scope.notification.show_alert = false;
+
+                        var dummy = new RegExp(filter.value);
+                        return true;
+                    } catch (error) {
+                        $scope.notification.show_alert = true;
+                        $scope.notification.alert_msg = '"' + filter.value + '": is not valid.  Only regular expressions are supported.';
```
vbusch: Should we be using messages, instead of $scope.notification.alert_msg?   As it stands now, line 243 messages.length > 0, will always be false, since messages isn't being populated.

k-wall:
@vbusch thanks for spotting my stupid copy/paste error.

PR #1738, opened 2018-10-03
New webconsole systemtest  doTestFilterAddressWithRegexSymbols

RH-sdavey:
* tests if filtering addresses in console is valid with regex
    * if non valid regex is used the console shows an alert (positive)

test passes


RH-sdavey:
@kornys  I can see jenkins failed but my test in the build passed and other (not related to mine) tests failed, whats best to do in that situation please?

PR #1739, opened 2018-10-03
Fix authorization tests

lulf:
* Disable test for sending to multicast address


kornys:
@lulf 👍

lulf:
Just FYI the multicast test was reenabled after a fix in the system test client

kornys:
I saw that test is enabled :)

PR #1741, opened 2018-10-03
Issue 1585 infra upgrades

lulf:
Add support for upgrading address space infrastructure

* Define AddressSpacePlan, AddressPlan, StandardInfraConfig and BrokeredInfraConfig as custom resources
* Annotate address space with current configuration
* Trigger update in different configuration
* Use latest api versions for k8s resources


lulf:
Will reopen in EnMasseProject organization

PR #1760, opened 2018-10-09
Plot of test duration + jenkins pipelines refactor

kornys:


kornys:
@lulf plot is visible in new left menu item called "Plots" you can see it now in job PR1760 in Jenkins.

PR #1762, opened 2018-10-09
Test client certificate before trusting

lulf:
This works around OpenShift 3.9 cert issues when using oc cluster up

### DIFF
--- a/api-server/src/main/java/io/enmasse/api/server/ApiServer.java
+++ b/api-server/src/main/java/io/enmasse/api/server/ApiServer.java
@@ -19,7 +19,6 @@ import io.enmasse.user.keycloak.KubeKeycloakFactory;
 import io.fabric8.kubernetes.api.model.ConfigMap;
 import io.fabric8.kubernetes.api.model.IntOrString;
 import io.fabric8.kubernetes.api.model.Service;
-import io.fabric8.kubernetes.client.KubernetesClientException;
 import io.fabric8.openshift.api.model.Route;
 import io.fabric8.openshift.client.DefaultOpenShiftClient;
 import io.fabric8.openshift.client.NamespacedOpenShiftClient;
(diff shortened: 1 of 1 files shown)


api-server/src/main/java/io/enmasse/api/server/ApiServer.java
```diff
@@ -76,10 +78,14 @@ public void start(Future<Void> startPromise) throws Exception {
...
             ConfigMap extensionApiserverAuthentication = client.configMaps().inNamespace(options.getApiserverClientCaConfigNamespace()).withName(options.getApiserverClientCaConfigName()).get();
-            clientCa = extensionApiserverAuthentication.getData().get("client-ca.file");
+            clientCa = extensionApiserverAuthentication.getData().get("client-ca-file");
             requestHeaderClientCa = extensionApiserverAuthentication.getData().get("requestheader-client-ca-file");
-        } catch (KubernetesClientException e) {
+            validateCert(clientCa);
+            validateCert(requestHeaderClientCa);
+        } catch (Exception e) {
             log.info("Unable to retrieve config for client CA. Skipping", e);
+            clientCa = null;
```
k-wall: Would it be a good idea to include whether clientCa or requestHeaderClientCa had failed?

api-server/src/main/java/io/enmasse/api/server/ApiServer.java
```diff
@@ -93,6 +99,11 @@ public void start(Future<Void> startPromise) throws Exception {
         });
     }
 
+    private static void validateCert(String ca) throws CertificateException {
+        CertificateFactory cf = CertificateFactory.getInstance("X.509");
+        cf.generateCertificate(new ByteArrayInputStream(ca.getBytes(StandardCharsets.UTF_8)));
```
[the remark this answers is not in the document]
lulf: I'll change it to use generateCertificates(). I think DER is unlikely, as tthe certs are typically generated by kubernetes/openshift which always uses PEM.

api-server/src/main/java/io/enmasse/api/server/ApiServer.java
lulf: It is unlikely that one of them fail, but I will add that.

PR #1764, opened 2018-10-10
Fix pipelines due to upgrade pipelines plugin

kornys:


kornys:
@lulf important fix for jenkinsfile-test pipeline I made bug in syntax.

kornys:
@lulf I need your admin rights :D

PR #1765, opened 2018-10-10
Fix quote args for systemtests

kornys:


kornys:
@lulf this should fix master tests pipeline. Right now master-brokered, master-standard and master-common run all tests instead of running specific tests for job.

PR #1772, opened 2018-10-11
Return AlreadyExists instead of Conflict for 409 response codes

lulf:
This aligns the response with that given by Kubernetes API server for 409 responses. I'm not sure if there is a case where we should set it to 'Conflict', but I think this will be correct for most foreseeable cases.


lulf:
> I did wonder whether this change would have been better expressed by subclassing ConflictExeption <- AlreadyExistsException and replacing the existing uses with the narrower type. This would save the surprise that would arrive if someone started using a 409 more generally elsewhere.

I agree. Attempting to do this revealed a need to add a dependency betweeen keycloak-user-api and api-common which contains the predefined exceptions.  I would like to cleanup/move some parts of api-common before doing that.

PR #1775, opened 2018-10-11
Exclude the tools jar, fix IDE errors

ctron:
For me, all projects in the Eclipse IDE are flagged with error, due to a missing dependency of `com.sun:tools`. From what I read, this can be excluded.


ctron:
> I suspect a misconfiguration of the Eclipse IDE, or an IDE bug.

I don't think that is the case. System dependencies are discouraged for exactly this reason.

I am running on JDK-10, and starting with Java 9+ the tools.jar is gone. So my IDE properly resolved the dependency to `<systemPath>/home/jreimann/jvm/jdk-10.0.2+13/../lib/tools.jar</systemPath>`, but the file no longer exists.

So, do you need this dependency? Or can it simply be excluded? Because right now, when I import EnMasse, the whole workspace lights up red. Hide all real errors behind that. So fixing this would really improve things IMHO.

k-wall:
@ctron I was wrong.  I interpreted the sentence from [1] 'The artifact is always available and is not looked up in a repository.' to mean that the tooling would neither try to resolve the dependency in the repository nor verify the JARs existence.  It turns out I am wrong on the last point.   Maven 3.3.9 (at least) complains that file does not exist when running Java > 8.

Can you rebase you PR and I'll approve?

You'll need to move your dependency exclusion from kubernetes-server-mock to io.sundr:builder-annotations.  The latter is now depended on directly, and it is it that has the dependency on com.sun:tools.

I note that on Java 10, the build actually fails later (below).    #1818  introduces a variable for 
 jboss-logging-processor.  If you override the jboss.logging.processor.version=2.1.0.Final, you should see the build completed and the tests pass.


```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.2:compile (default-compile) on project amqp-connector: Compilation failure
[ERROR] java.lang.NoClassDefFoundError: javax/annotation/Generated
[ERROR] at org.jboss.logging.processor.generator.model.ClassModel.generateModel(ClassModel.java:134)
[ERROR] at org.jboss.logging.processor.generator.model.MessageLoggerImplementor.generateModel(MessageLoggerImplementor.java:108)
[ERROR] at org.jboss.logging.processor.generator.model.ClassModel.generateAndWrite(ClassModel.java:119)
[ERROR] at org.jboss.logging.processor.apt.ImplementationClassGenerator.processTypeElement(ImplementationClassGenerator.java:63)
```

k-wall:
I opened an issue on the underlying project https://github.com/sundrio/sundrio/issues/117

ctron:
> Can your rebase you PR and I'll approve?

I just did. And I also move the exclusion to the other dependency. Thank you for investigating this!

PR #1793, opened 2018-10-15
Issue 1616 update address api group

lulf:


lulf:
Closing this, as we've decided it is premature and no real benefit at this point.

Issue #1728, opened 2018-10-02
Console standard: address web item does not shows count of senders/receiver <random-error>


kornys:
**Describe the bug**
Address item still shows that count of senders/receivers, message-in/out are 0.

**To Reproduce**
1. Deploy enmasse
2. Create standard address space
3. Create pooled-queue address
4. install cli-rhea:
```
npm install cli-rhea -g
```
5. Run cli-rhea connector which creates 1 connections with 5 senders and 10 receivers
```
cli-rhea-connector --conn-ssl true --receiver-count 10 --obj-ctrl CESR --log-lib TRANSPORT_FRM --broker username:password@messaging-route:443 --count 1 --address pooled-queue-name --timeout 120 --sender-count 5 --conn-property connection_property1~50 --conn-property connection_property2~testValue
```

![io enmasse systemtest standard web firefoxwebconsoletest testclientsmetrics_2018-10-02_15 53 48 108](https://user-images.githubusercontent.com/11318335/46356004-2e426d00-c662-11e8-8092-42949d41b6e6.png)
![io enmasse systemtest standard web firefoxwebconsoletest testclientsmetrics_2018-10-02_15 53 48 498](https://user-images.githubusercontent.com/11318335/46356005-2e426d00-c662-11e8-9da6-ffada866ae67.png)
![io enmasse systemtest standard web firefoxwebconsoletest testclientsmetrics_2018-10-02_15 53 48 617](https://user-images.githubusercontent.com/11318335/46356006-2edb0380-c662-11e8-957f-2f0f7d7c93d8.png)
![io enmasse systemtest standard web firefoxwebconsoletest testclientsmetrics_2018-10-02_15 53 49 068](https://user-images.githubusercontent.com/11318335/46356007-2edb0380-c662-11e8-9ffe-c7efff9f06ca.png)
![io enmasse systemtest standard web firefoxwebconsoletest testclientsmetrics_2018-10-02_15 54 56 294](https://user-images.githubusercontent.com/11318335/46356008-2edb0380-c662-11e8-95d3-9e2481012044.png)
![io enmasse systemtest standard web firefoxwebconsoletest testclientsmetrics_2018-10-02_15 54 56 707](https://user-images.githubusercontent.com/11318335/46356009-2edb0380-c662-11e8-8216-6ba5e5fd6740.png)
[...]

PR #1797, opened 2018-10-16
Avoid error situation recorded in console log for CI failure.

grs:
Potential fix for issue #1728.


kornys:
@grs I run testClientsMetrics many times on my server with your change from this branch and I'm not able to reproduce issue #1728. Jenkins runs this test without issue as well

PR #1798, opened 2018-10-16
Avoid standard controller state race with watches and resync

k-wall:


lulf:
I think it would be good to capture this pattern in a class that can be used by ConfigMapAddressSpaceApi (and others) as well, since stale data could be an issue there at some point as well.

Also I’m not sure we can rely on resourceVersion after reading https://github.com/kubernetes/community/blob/master/contributors/devel/api-conventions.md#types-kinds

“resourceVersion: a string that identifies the internal version of this object that can be used by clients to determine when objects have changed. This value MUST be treated as opaque by clients and passed unmodified back to the server. Clients should not assume that the resource version has meaning across namespaces, different kinds of resources, or different servers. (See concurrency control, below, for more details.)”

The documentation refer to a generation filed, which may be more correct to use in this case:

“generation: a sequence number representing a specific generation of the desired state. Set by the system and monotonically increasing, per-resource. May be compared, such as for RAW and WAW consistency.”

k-wall:
@lulf Thanks for the link.  I hadn't found it.  I agree that the usage of the `resourceVersion` proposed by the patch is not the correct way to go.

We could introduce a `resourceVersion` and use it to determine staleness instead. If we took this approach, then the correct course to do would be to introduce it uniformly across the whole API.  However, the whole approach feels odd.   

I have been looking for alternatives.

1) At the moment standard-controller is a singleton.  If this will always be true, then one alternative would be keep track of state within AddressController itself.  Once an address is known to the controller it would from that point refer to its own state cache and use this to drive the provisioning process.  I think this would just require AddressController to maintain a map of Address=> Status.  It would add keys to the map as new addresses are created and remove them as they disappeared from the addressList.  It would continue to update the address config maps as it does today so the rest of the world could continue to observe state.

2) @lulf your original initial suggestion was make sure that address provision process was deterministic and always produce the same provisioning outcome even if the address state is stale.  I suspect the non-determinism will be coming from FifoQueue#store (a HashMap).  I think a [LinkedHashMap](https://docs.oracle.com/javase/8/docs/api/java/util/LinkedHashMap.html) (which retains iteration order when a key/vaue pair is replaced), would reduce the problem (untested).   However, this would not solve the flurry of needless config map updates that would occur (io.enmasse.k8s.api.AddressApi#replaceAddress) when an address's state is stale.  Not sure  if this actually causes a problem.

Thoughts welcome.

lulf:
@k-wall I think that we might want to be able to scale standard-controller in a way where we partition the addresses it needs to manage between them. However, I don't think 1) is in conflict with that depending on how were to do the partitioning. I think the assumption "a single address will be handled by 1 controller" should hold true.

lulf:
@k-wall I think that we might want to be able to scale standard-controller in a way where we partition the addresses it needs to manage between them. However, I don't think 1) is in conflict with that depending on how were to do the partitioning. I think the assumption "a single address will be handled by 1 controller" should hold true.

k-wall:
Taking alternative approach

PR #1804, opened 2018-10-17
New systemtest, deleting already filtered addresses removes them from…

RH-sdavey:
… filtered list

### DIFF
--- a/systemtests/src/test/java/io/enmasse/systemtest/bases/web/WebConsoleTest.java
+++ b/systemtests/src/test/java/io/enmasse/systemtest/bases/web/WebConsoleTest.java
@@ -189,6 +189,42 @@ public abstract class WebConsoleTest extends TestBaseWithShared implements ISele
                 String.format("Console failed, does not contain %d addresses", addressCount));
     }
 
+    protected void doTestDeleteFilteredAddress() throws Exception {
+        String testString = "addressName";
+        List<AddressWebItem> items;
+        int addressTotal = 2;
+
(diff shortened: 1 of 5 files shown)


systemtests/src/test/java/io/enmasse/systemtest/bases/web/WebConsoleTest.java
```diff
@@ -189,6 +189,42 @@ protected void doTestFilterAddressesByName() throws Exception {
...
+
+        Destination destQueue = Destination.destination(AddressType.QUEUE, testString + "Queue",
+                getDefaultPlan(AddressType.QUEUE), Optional.empty());
+
+        Destination destTopic = Destination.destination(AddressType.TOPIC, testString + "Topic",
+                getDefaultPlan(AddressType.TOPIC), Optional.empty());
+
+        consoleWebPage = new ConsoleWebPage(selenium, getConsoleRoute(sharedAddressSpace), addressApiClient,
+                sharedAddressSpace, defaultCredentials);
+        consoleWebPage.openWebConsolePage();
```
kornys: you don't need to open web page if you then call createAdderrWebConsole which already contains this operation

systemtests/src/test/java/io/enmasse/systemtest/bases/web/WebConsoleTest.java
```diff
@@ -189,6 +189,42 @@ protected void doTestFilterAddressesByName() throws Exception {
...
+        consoleWebPage.openAddressesPageWebConsole();
```
kornys: same as above, this is in createAddressWebConsole method

PR #1811, opened 2018-10-18
Added basic monitoring configuration

robshelly:


address-space-controller/pom.xml
```diff
@@ -44,6 +44,11 @@
       <artifactId>vertx-proton</artifactId>
       <scope>compile</scope>
     </dependency>
+    <dependency>
+      <groupId>io.vertx</groupId>
+      <artifactId>vertx-web</artifactId>
+      <scope>compile</scope>
+    </dependency>
```
lulf: I would prefer og we didnt introduce another dependency here, but used the existing vertx http server

api-server/src/main/java/io/enmasse/api/v1/http/HttpHealthService.java
```diff
@@ -28,4 +30,21 @@ public HealthResponse(int status, String message) {
             this.message = message;
         }
     }
+
+    @GET
+    @Produces({MediaType.TEXT_HTML})
+    @Path("metrics")
+    public Response getMetrics() {
+        String name = this.getClass().getPackage().getImplementationTitle();
+        String version = this.getClass().getPackage().getImplementationVersion();
```
lulf: Is this the name and version of the component? If so, I would prefer to hardcode the name here, and get version from the image environment. Then we don’t need the extra maven stuff either right?

templates/address-space-controller/040-Service-address-space-controller.yaml
```diff
@@ -0,0 +1,15 @@
...
+  labels:
+    app: enmasse
+spec:
+  ports:
+  - name: health
+    port: 8080
+    protocol: TCP
+    targetPort: http
+  selector:
+    component: address-space-controller
```
lulf: Do we actually need a service or can we configure Prometheus to scrape the pod instead as done with the router and broker

templates/prometheus/030-Deployment-prometheus.yaml
```diff
@@ -12,7 +12,7 @@ spec:
         name: prometheus
     spec:
       containers:
-      - image: prom/prometheus:v2.0.0
+      - image: prom/prometheus:latest
```
lulf: This should be configurable. Replace with ${PROMETHEUS_IMAGE} and change Makefile to to a substituttion

templates/grafana/030-Deployment-grafana.yaml
```diff
@@ -0,0 +1,49 @@
...
+spec:
+  replicas: 1
+  template:
+    metadata:
+      labels:
+        name: grafana
+    spec:
+      containers:
+      - name: grafana
+        image: grafana/grafana:latest
```
lulf: Make it configurable using ${GRAFANA_IMAGE} and change Makefile to perform substitutiton.

ansible/inventory/with-standard-authservice.txt
```diff
@@ -9,3 +9,9 @@ service_catalog=False
 register_api_server=True
 keycloak_admin_password=admin
 authentication_services=["standard"]
+enable_monitoring=True
+smtp_server= localhost
+smtp_username= username
+smtp_password= password
+smtp_from_address= alerts@example.com
+sys_admin_email= root@localhost
```
lulf: I would say we leave these settings out in this default config for now.

templates/address-space-controller/040-Service-address-space-controller.yaml
robshelly: As discussed, service included to allow Prometheus to alert pods which are down

ansible/inventory/with-standard-authservice.txt
```diff
@@ -9,3 +9,4 @@ service_catalog=False
 register_api_server=True
 keycloak_admin_password=admin
 authentication_services=["standard"]
+enable_monitoring=True
```
lulf: I would say False by default until we have it installing with the operator.

Issue #1810, opened 2018-10-17
Ready addresses seen to flip back to isReady=false


k-wall:
When sequentially creating a large number of addresses (1000-2000) on my machine, I sometimes I see a situation after many minutes of creating addresses normally (and becoming ready), suddenly the all address flip to `ready=false`.    The addresses then seem to stick in the ready=false state.  I notice if I restart the admin pod, the addresses come back ready again.

On querying the addresses with the following command I see that all addresses are reported as a missing auto links on router 4 and a subset are missing the address on the same router.

` oc get address  --output=jsonpath='{range .items[*]}{.metadata.name}{"\t"}{.status.isReady}{"\t"}{.status.messages}{"\t"}{"\n"}{end}'
`

```

standard.a95	false	[Address a95 is missing autoLinks on qdrouterd-64u4p9jwnw-4]
standard.a950	false	[Address a950 not found on qdrouterd-64u4p9jwnw-4 Address a950 is missing autoLinks on qdrouterd-64u4p9jwnw-4]
```

Examining the router directly:
oc exec -it qdrouterd-64u4p9jwnw-4 -c router -- qdstat  --autolinks
shows no auto links.

oc exec -it qdrouterd-64u4p9jwnw-4 -c router -- qdstat --address
shows some addresses.

[agent_admin.64u4p9jwnw-8569bfd475-2nvms.log](https://github.com/EnMasseProject/enmasse/files/2489539/admin.64u4p9jwnw-8569bfd475-2nvms.log)
[qdrouterd-64u4p9jwnw-4.log](https://github.com/EnMasseProject/enmasse/files/2489540/qdrouterd-64u4p9jwnw-4.log)

I notice that the agent is periodically reporting "still waiting for responses for" for qdrouterd-64u4p9jwnw-4.

```
2018-10-17T22:34:22.951Z agent warn [mgmt:qdrouterd-64u4p9jwnw-4] still waiting for responses for ["$management52","$management53"
```

The router is showing the agent's connection (ragent-64u4p9jwnw:5671).

```
Connections
  id   host                                                                                    container                             role             dir  security                              authentication                            tenant
  =================================================================================================================================================================================================================================================
  1    qdrouterd-64u4p9jwnw-3.qdrouterd-headless-64u4p9jwnw.myproject.svc.cluster.local:55672  qdrouterd-64u4p9jwnw-3                inter-router     out  TLSv1.2(ECDHE-RSA-AES256-GCM-SHA384)  x.509
[...]

Issue #1566, opened 2018-08-30
Some brokered addresses not available after broker/agent restart


calohmn:
Using EnMasse 0.21.2:
We have restarted the admin pod and shortly after restarted the broker pod.
After that, several brokered addresses were not available - in the EnMasse console they were shown with the "in progress" icon.

This didn't get fixed automatically - only a restart of the admin pod resolved the situation.

This has happened a few times before (we restart these pods on a daily basis). 

We've assembled some logs files (see [logs_180829.zip](https://github.com/EnMasseProject/enmasse/files/2337227/logs_180829.zip)) (unfortunately no qdstat or broker stats for that).

Some observations:
- there are 362 brokered addresses in total ("event/*")
- concerning timing: it looks like while the agent is starting up, it fetches broker statistics while the broker itself is still starting up - probably not getting the eventual number of queues
- as part of the agent restart to fix the situation, the agent pod log (`agent__afterRestart.log`) contains 73 "Creating queue" entries. So these queues must have been missing before - but the broker log shows no indication of that.
- standard-controller logs show nothing special (always the same "Total: 1086, Active: 1086" output)

lulf:
By not available, do you mean that you can't send/recv to them or that they are not active in the console?

grs:
Are the addresses all 'pooled' (rather than 'sharded')?

grs:
also, if pooled, is there one broker or more than one? How do you restart the pods (e.g. delete the pod, or scale down and up again)?

calohmn:
Yes, addresses are of type pooled queue. There is only one broker. 
Restart is done as part of a "kops rolling-update cluster".

grs:
As yet I have not managed to reproduce this. However looking at the logs, of the 73 queues created by the agent after restart, 69 were attempted before the restart but failed, 49 due to 'Binding already exists LocalQueueBinding' 20 due to 'Queue xxx already exists'. For those that failed with the binding error, there are (at least) two concurrent entries in the broker log indicating queue creation. My hunch is that there are two concurrent attempts to create the queue and associated address, and one of them fails. However this doesn't explain why the queue then needs to be recreated later - seemingly it gets deleted for some reason in between, possibly due to the auto-delete behaviour that appears to be enabled by default. I'll continue to try to reproduce, and perhaps experiment with a patch that turns that auto-delete off.

grs:
One other question: did you restart the router also? Which is the router log for?

[...]

PR #1815, opened 2018-10-19
throttle requests to router

grs:
Potential fix to issues #1566 and #1810

### DIFF
--- a/agent/lib/qdr.js
+++ b/agent/lib/qdr.js
@@ -18,6 +18,8 @@
 var log = require("./log.js").logger();
 var util = require('util');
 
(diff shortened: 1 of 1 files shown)


grs:
@junezhang sorry, misclicked and added you as review by mistake!

agent/lib/qdr.js
```diff
@@ -42,6 +44,7 @@ var Router = function (connection, router, agent) {
         this.tracking = {
             sent: 0,
             recv: 0,
+            outstanding: 0,
```
k-wall: Is it worth adding outstanding to the info reported periodically by Router.prototype.log_info?

Issue #1605, opened 2018-09-10
Sender and receiver links not identifiable to the end user


k-wall:
In the console,. the address tab lists each address along with a table enumerating the links to that address.  The rows in this table includes a id column column showing what is presumably a internal link name assigned within the router.    As an operator, for problem resolution purposes, I would expect to be able to determine the application from which the link originated.

Perhaps the id should display the name of the link assigned by the application.  Perhaps too there should be navigation back to the connection associated that formed that link.

![unhelpfullinknames](https://user-images.githubusercontent.com/18440250/45303946-e884ff80-b50e-11e8-8706-d06287f92558.png)

grs:
Should be fixed by https://github.com/EnMasseProject/enmasse/pull/1614 which exposes the link name (the id is the id of the router, which may not be something we want to expose)

k-wall:
@yih-wang The IP/ephemeral port will help identify the application - the ephemeral part makes it unique even if many apps run on the same host.  I would support showing the Container ID too.  We already display the Container ID on the Connection tab.  We ought to show it on the Address tab too.  We currently show the Connection ID created by the router in column 1 of the Incoming Links/Outgoing Links tables.  It would be better if this was the Connection ID created by the application.  @grs what do you think?

grs:
Agreed, ip/port and container-id of connection would be more useful than router's container id.

Another approach to identifying connections with links to/from a given address would be to add a filter to the connections view where you could specify source or target address and the list would be filtered to only those connections with one or more matching links. That would be better for cases where there might be a large number of connections as you could then sort and/ro filter further to manage the data. The table in the address detail really assumes a small number of connections.

yih-wang:
@k-wall @grs It's more user-friendly that navigate users to Connection screen through a link. In that case, it's not necessary to display both IP/Port and Container ID in Address table because they can both navigate to the same connection. The process and interfaces could be like this:

1 - Users click on the hostname to navigate to the connection that created the link. I remove the Id column as it's not too useful, and use Name as the primary attribute instead. It can keep consistency with Connection table in this way.
![1 - addresses](https://user-images.githubusercontent.com/23331885/47142296-31da2300-d2f5-11e8-9c2c-0a7658e61a27.png)

2 - When navigating to Connections screen, it automatically applies a filter with this hostname so that the user can locate to this connection. (I suppose there will be only one result, because connections with the same hostname show in same row.) The result row is expanded by default.
[...]

PR #1836, opened 2018-10-23
Display the Container ID of the client's connection in Console.

k-wall:
Partly addresses #1605.

Screenshot attached.

issue-1605-use-containerid-of-clientcon

### DIFF
--- a/agent/lib/router_stats.js
+++ b/agent/lib/router_stats.js
@@ -74,7 +74,10 @@ function backlog (link_stats) {
 function routerName (link_stats, router) {
     return router && router.target ? router.target.split('/')[3] : undefined;
 }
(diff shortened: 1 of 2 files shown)


yih-wang:
@k-wall Happy to see your updates, wondering 'kw1' refers to which Container ID (see my attachment)? 
artboard 2

grs:
@yih-wang kw1 would be the container id of the connection i.e. the top number in your picture 267f943f-... A lot of libraries default that to a uuid but it can also be supplied by the application making the connection (as in the case for Keith's screenshot).

The link names are sometimes also uuids, though again that is dependent on the library used to connect with and it can also be overridden by the application.

yih-wang:
@grs Got it, thanks!

Issue #1550, opened 2018-08-28
Agent should lock down Node dependencies


k-wall:
The agent build should ship a `shrinkwrap.json` in order that all dependencies, including transitive dependencies, are already locked down to specific versions and repeatable builds result.

https://docs.npmjs.com/cli/shrinkwrap

PR #1843, opened 2018-10-23
include shrinkwrap definition for runtime deps

grs:
closes issue #1550

### DIFF
--- a/agent/pom.xml
+++ b/agent/pom.xml
@@ -64,6 +64,7 @@
                              <directory>${project.basedir}</directory>
                              <includes>
                                 <include>package.json</include>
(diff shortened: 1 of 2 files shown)


agent/pom.xml
```diff
@@ -94,6 +95,7 @@
       <plugin>
         <groupId>com.github.eirslett</groupId>
         <artifactId>frontend-maven-plugin</artifactId>
+        <version>0.0.27</version>
```
lulf: This seems to be a very old version of the plugin, latest is 1.6, which is defined in the root pom.xml. It should't be necessary to specify the version here.

PR #1844, opened 2018-10-23
remove dashboard as it causes confusion without adding value

grs:


k-wall:
@grs @rgodfrey  @lulf - one thought, with this change we have lost the ability to download the messaging certificate.   To be honest, the dash board was alway an odd place for it.   I think you'd normally expect a CA signed cert to be used, so maybe this is not an issue.  However, for the case where self signed/private CA certs are in use, perhaps a section in the docs describing how to get the server certificate from the oc/kubectl command line would be required

grs:
You can still get the cert from the help menu I believe.

rgodfrey:
@k-wall @grs I think documenting how to get the certificate from the command line makes a lot of sense given our move more towards command-line style interaction for setting up EnMasse.  Where the address space has been created via the service catalog (and thus the cert information would be harder to get from openshift) the certificate should be available in the binding. This should definitely all be added to our docs.

PR #1846, opened 2018-10-23
Issue 1837 embed default plans in address space controller

lulf:


pom.xml
```diff
@@ -48,6 +48,7 @@
         <jackson-annotations.version>2.9.6</jackson-annotations.version>
         <jackson-core.version>2.9.6</jackson-core.version>
         <jackson-databind.version>2.9.6</jackson-databind.version>
+        <jackson-dataformat-yaml.version>2.9.6</jackson-dataformat-yaml.version>
```
k-wall: We already have a transitive dependency on jackson-dataformat-json coming from kubernetes-client.  Since  jackson-dataformat-json and  jackson-dataformat-yaml belong to the same project, it is not safe to allow their versions to deviate.   I think it is best we define jackson-dataformat.version and dependency manage both libraries to the same version.

Issue #1832, opened 2018-10-23
Add metrics for address space


lulf:
Metrics should be exposed through prometheus http endpoint and configured in alertmanager and grafana dashboard.

PR #1856, opened 2018-10-25
Add address space status metrics to address-space-controller

lulf:
* Add Java API for reporting metrics in controller loop
* Report address space status and number of address spaces
* Format metrics as prometheus metrics

Issue #1832


lulf:
@robshelly Should build finally :) . I changed the api-server to use the metrics-api library as well.

PR #1865, opened 2018-10-26
Reduce build time dependencies

k-wall:


k-wall:
Avoid the pulling in (at build time) of unnecessary Artemis dependencies (and its dependencies).  This simplifies EnMasse's build time footprint.  No functional change.

k-wall:
I just use `mvn dependency:tree`

k-wall:
215fecb avoids build time dependencies on a bunch of dependencies of keycloak-services itself (such as Twitter).

lulf:
I was more thinking how you determine which dependencies are only required build time vs run time

k-wall:
> I was more thinking how you determine which dependencies are only required build time vs run time

Nothing fancy.  I was just examining the contents of the Jar artefact produced and reasoning from there.  If we are producing a plugin for a product, then the only classes that need to exist on the class path at compile time are those needed to compile for the plugin.  Everything else is superfluous.  At runtime, it is just the plugin's classes and any dependencies the plugin introduces itself.

PR #1866, opened 2018-10-26
Mqtt internal tests

kornys:
Tests for base testing internal mqtt messaging.


kornys:
@lulf could you please re-approve ? :) I had to fix port for amqps.

PR #1874, opened 2018-10-28
De-springify the MQTT components.

k-wall:
Fixes 1529


k-wall:
@lulf one thing I'd like to mention.  I opted to programatically tell Vertx to use SLF4J (io.vertx.core.logging.LoggerFactory.LOGGER_DELEGATE_FACTORY_CLASS_NAME), rather than including the JUL to SLF4J bridge that was previously the case (it was a transitive dependency of Spring Boot).  This seemed simpler to me, but does mean if something else were to use JUL, it would not get bridged.

PR #1878, opened 2018-10-29
Remove option to add console access to service binding

### LINKED ISSUE #1756: Remove the ability to request console access on binding
We need to remove the Service Broker's current ability to request console access on binding.   When running EnMasse on OpenShift, the default is now to log on to the Console using OpenShift credentials, so the ability to create console users (that won't be federated to an openshift identity) no longer makes sense.

rgodfrey:
Remove console access options and always add external host info
This closes #1756

### DIFF
--- a/service-broker/src/main/java/io/enmasse/osb/api/ServiceMapping.java
+++ b/service-broker/src/main/java/io/enmasse/osb/api/ServiceMapping.java
@@ -4,17 +4,28 @@
  */
 package io.enmasse.osb.api;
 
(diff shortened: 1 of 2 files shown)


kornys:
I will have to update tests, because tests will fail they count with access console rights in binding.

rgodfrey:
@kornys ok  - I didn't find anything mentioning the text of the options, do they just do a count somewhere?

kornys:
@rgodfrey test failures are not related to your changes, so it can be merged.

PR #1879, opened 2018-10-29
Add new default address space plans and address plans

lulf:


rgodfrey:
@lulf @grs @k-wall So I don't think this is a blocker, but I do find the ordering of the parts of the name slightly odd:

> large-standard-anycast

to me parses as a large "standard anycast" instance. 

As such I think I prefer "standard-large-anycast" or, perhaps even clearer, "standard-anycast-large"

lulf:
@rgodfrey I'm fine with changing that, the current approach is easier to misread I agree.

lulf:
@rgodfrey In that case, should we also change the address space plan names from 'small-standard', 'medium-standard' to 'standard-small', 'standard-medium' etc.?

rgodfrey:
@lulf personally I'd be good with the address space plans being "standard-small", "standard-medium"...  in general _\<what it is\>-\<size\>_

PR #1881, opened 2018-10-29
Import all CA certs

lulf:


calohmn:
@lulf When using this keytool command, I still only get the first cert in the keystore.

lulf:
@calohmn Can you try with this change?

lulf:
@calohmn I think importing the first cert is fine for now.

PR #1882, opened 2018-10-29
Implement deletion of bindings

### LINKED ISSUE #1751: Binding credentials can still be used after the secret is deleted
**Describe the bug**
I am still able to send messages to a queue in brokered addr space (haven't tried standard addr space yet, will add to comments when I do) using binding credentials that I have deleted. 

**To Reproduce**
1. Create brokered addr space and create a queue "myqueue"
2. In Openshift UI, create a binding for "Enmasse(brokered)" provisioned service
3. Use the given secret credentials send messages to myqueue  ( works as expected ) 
4. delete binding
5. use the same credentials to send messages to myqueue  (works, I dont expect it should) 
optional. wait 15 minutes and try step 5 again, (just to confirm its not a "config refresh" issue, its not)

**Expected behavior**
once the binding is deleted, the username and password should not be valid, authentication should fail, and the connection should not be made, no messages sent

**Additional context**
so far only tested with cli-proton-python sender client,

cli-proton-python-sender --timeout 2 --broker amqps://user-99825c64-c8b0-11e8-a8c1-0242ac110008:o4I8TT6tbOSfoSA2EgjL-DILn5QUoPvQ@messaging-jwgjhr63mp-myproject.<<ip>>.nip.io:443/q2 --log-msgs dict --count 10

rgodfrey:
this closes #1751

### DIFF
--- a/service-broker/src/main/java/io/enmasse/osb/api/bind/OSBBindingService.java
+++ b/service-broker/src/main/java/io/enmasse/osb/api/bind/OSBBindingService.java
@@ -215,23 +215,14 @@ public class OSBBindingService extends OSBServiceBase {
     }
 
     private boolean deleteUser(AddressSpace addressSpace, String username) throws Exception {
-            UserSpec.Builder specBuilder = new UserSpec.Builder();
-            specBuilder.setUsername(username);
-            User user = new User.Builder()
-                .setMetadata(new UserMetadata.Builder()
-                                 .setNamespace(addressSpace.getNamespace())
-                                 .setName(addressSpace.getName() + "." + username)
-                                 .build())
-                .setSpec(specBuilder.build())
-                .build();
 
-            String realmName = addressSpace.getAnnotation(AnnotationKeys.REALM_NAME);
-            if (userApi.getUserWithName(realmName, username).isPresent()) {
-                userApi.deleteUser(realmName, user);
-                return true;
-            } else {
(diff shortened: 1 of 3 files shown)


service-broker/src/main/java/io/enmasse/osb/api/bind/OSBBindingService.java
```diff
@@ -196,12 +196,44 @@ private String generatePassword() {
...
+            User user = new User.Builder()
+                .setMetadata(new UserMetadata.Builder()
+                                 .setNamespace(addressSpace.getNamespace())
+                                 .setName(addressSpace.getName() + "." + username)
+                                 .build())
+                .setSpec(specBuilder.build())
+                .build();
+
+            String realmName = addressSpace.getAnnotation(AnnotationKeys.REALM_NAME);
+            if (userApi.getUserWithName(realmName, username).isPresent()) {
```
lulf: Why not  use the user returned here in the call to delete instead of constructing one?

rgodfrey: because I was doing a quick C&P without actually looking at the code :-)

I did notice this could be simplified after I pushed it

kornys:
@rgodfrey tests passed, the error was caused by archiving artifacts. So it can be merged.

PR #1884, opened 2018-10-29
Add maven profile to perform a CVE check using OWASP's dependency-check-maven and RH victims.

k-wall:
Note that currently RH victims can't fail the build (there's an open PR https://github.com/victims/maven-security-versions/pull/5  to do this).  You need to parse the output and fail if find the string CVE is found.


k-wall:
There is more concise way suppress the false positive than the one I chose.  It is a suppression rule based on the  Common Platform Enumeration (CPE).  https://jeremylong.github.io/DependencyCheck/general/suppression.html.  It should be possible to tell the Engine concisely that Selenium is not Opera using this method and that should reduce the size of the suppression file somewhat.  Same applies to the other misidentifications.

I'll turn off the CVE check for systemtests.  This should be just the case of setting skip on the module.   
 https://jeremylong.github.io/DependencyCheck/dependency-check-maven/configuration.html

It is also possible to turn off scanning of artefacts that are test scope.

PR #1888, opened 2018-10-30
Migrate default plans

lulf:
In order for users to upgrade from 0.23.0 to 0.24.0 out of the box, the following PR is needed. It will change the address space plan and address plan names to the appropriate new name if they don't exist.


lulf:
@vbusch No, we will only support N+1 versions for now.

PR #1892, opened 2018-10-30
Docs: using qdstat docs

jenmalloy:


documentation/modules/proc-view-connections-qdstat.adoc
```diff
@@ -0,0 +1,25 @@
+// Module included in the following assemblies:
+//
+// assemblies/assembly-qdstat.adoc
+
+[id='view-connections-qdstat-{context}']
+= Viewing router connections using qdstat
+
+A connection to the container running `qdstat` is created by default on the router. You can view the router connections using `qdstat`.
```
robshelly: As the use case here is to inspect the state of the router I would say we can leave out the explanation of the connection to the container running qdstat.

> You can view the router connections using `qdstat`.

documentation/modules/proc-view-link-routes-qdstat.adoc
```diff
@@ -0,0 +1,24 @@
+// Module included in the following assemblies:
+//
+// assemblies/assembly-qdstat.adoc
+
+[id='view-link-routes-qdstat-{context}']
+= Viewing link routes using qdstat
+
+Two link routes are created by default to manage last will and testament. You can view these link routes using `qdstat`.
```
robshelly: These link routes only apply when MQTT is enabled so they are not always created by default. So again, I'd say just leave out the explanation of the default link routes.

> You can view the link routes using `qdstat`.

PR #1897, opened 2018-10-31
Docs: download instructions edit

jenmalloy:


ncbaratta:
Can we handle this with conditionals? We don't want to be creating downstream only files if we can avoid it.

jenmalloy:
@lulf, this PR reflects how Strimzi is handling it (I got it a little wrong in the original way I did it, so this PR is to get it right)

PR #1899, opened 2018-10-31
Include license bodies, license summary (xml) and third-party info

k-wall:
All shaded JARs now include:

```
/licenses.xml    - summary that lists the licenses of all dependencies used by the artefact
/licenses/*      - license body (txt/html)
/THIRD-PARTY.txt - identifies the license associated with each third party dependency
```

### DIFF
--- a/agent/pom.xml
+++ b/agent/pom.xml
@@ -195,6 +195,14 @@
           </execution>
         </executions>
       </plugin>
+      <plugin>
+          <groupId>org.codehaus.mojo</groupId>
+          <artifactId>license-maven-plugin</artifactId>
+          <configuration>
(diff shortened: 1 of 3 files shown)


k-wall:
@lulf  I'll be raising a separate PR for the equivalent change to the agent module (which will use https://www.npmjs.com/package/license-reporter), however, I need to refactor agent first to avoid the use of bower.

One thing I wish to point out with this change:  The  text of the http://creativecommons.org/publicdomain/zero/1.0/ license is not downloadable by the plugin.  I *guess* the site is blocking the user agent.  This affects org.reactivestreams dependency and means its license text won't appear in the licenses directory.  The license will be referenced in the THIRDPRTY party file and appear in the license.xml.  I think we can live with this wrinkle.

PR #1900, opened 2018-11-01
Use npm to install console dependencies

k-wall:


k-wall:
Console is functioning properly but there is one issue that I am not completely satisfied which I wish to highlight.

Previously the index.html pulled in the minified `lodash.min.js` from the Bower Lodash module 3.10.1.  The NPM release of the same dependency (same version) does not include the minified version, so I have had to pull in the un-minified version (node_modules/angular-patternfly/node_modules/lodash/index.js).   This is functioning, but as I am not expert in this technology comments would be appreciated.

PR #1904, opened 2018-11-01
Create license metadata for agent too

k-wall:
Uses license-reporter to generate a license.xml for both agent and the console.  These artefacts
are then merge into a single file that is included in the distribution.  license-reporter is also
used to create files containing the text of the license themselves. These too are include in the
distribution artefact.

Note that unlike the Maven license-plugin, license-reporter does not actually download the license.  Instead it takes it from the npm module itself.


lulf:
The failing tests seem unrelated to this change, so I think it can be merged. I am looking into the unstable tests.

PR #1907, opened 2018-11-01
Docs: additional info about address spaces

jenmalloy:


documentation/modules/con-standard-address-space.adoc
```diff
@@ -4,11 +4,11 @@
 
 [id='con-standard-address-space-{context}']
 = Standard Address Space
-The standard address space is the default address space in {ProductName}. It consists of an AMQP router network in combination with attachable storage units. The implementation of a storage unit is hidden from the client and the routers with a well-defined API. This address space type is appropriate when you have many connections and addresses. However, it has the following limitations: 
+The standard address space is the default address space in {ProductName}. It consists of an AMQP router network in combination with attachable storage units. The implementation of a storage unit is hidden from the client and the routers with a well-defined API. This address space type is appropriate when you have many connections and addresses. However, the standard address space has the following limitations:
 
 * No transaction support
 * No message ordering
-* No selectors on queues
+* No browsing or selectors on queues
```
grs: I'd possibly leave these as two separate items ('no selectors on queues' ''no browsing of queues '). Not a big issue though if you prefer this form.

documentation/modules/con-standard-address-space.adoc
```diff
@@ -4,11 +4,11 @@
 
 [id='con-standard-address-space-{context}']
 = Standard Address Space
-The standard address space is the default address space in {ProductName}. It consists of an AMQP router network in combination with attachable storage units. The implementation of a storage unit is hidden from the client and the routers with a well-defined API. This address space type is appropriate when you have many connections and addresses. However, it has the following limitations: 
+The standard address space is the default address space in {ProductName}. It consists of an AMQP router network in combination with attachable storage units. The implementation of a storage unit is hidden from the client and the routers with a well-defined API. This address space type is appropriate when you have many connections and addresses. However, the standard address space has the following limitations:
```
grs: I know you didn't add this but 'The implementation of a storage unit is hidden from the client and the routers with a well-defined API.' seems odd to me. What does that mean? @lulf @rgodfrey ?

documentation/modules/con-standard-address-space.adoc
jenmalloy: Nope, that was a typo on my part. Will fix and keep them as two separate items.

documentation/modules/con-standard-address-space.adoc
lulf: @grs I think it is a bit too abstract :) How about replacing storage units with broker and reword to something like "Clients connect to a message router, which forwards messages to/from one or more message brokers" ?

grs: Yes, that is easier to understand.

jenmalloy: @grs, @lulf, Thanks, I'll edit this to reflect the new verbiage.

PR #1908, opened 2018-11-01
disable the dynamic management of connectors by agent as this is now …

grs:
…done by the router image on startup

This resolves errors in the logs of routers of the form Connection from 127.0.0.1:40540 (to 0.0.0.0:55672) failed: amqp:connection:framing-error SSL Failure: error:1417C0C7:SSL routines:tls_process_client_certificate:peer did not return a certificate


lulf:
Test looks unrelated, but a general instability i've seen some times after the new plans got merged. I think this can be merged though.

PR #1910, opened 2018-11-01
Docs: dynamic subs on topics and wildcard behavior

jenmalloy:


documentation/modules/con-brokered-topic.adoc
```diff
@@ -9,3 +9,18 @@
 The topic address type supports the publish-subscribe messaging pattern in which there are 1..N producers and 1..M consumers. Each message published to a topic address is forwarded to all subscribers for that address. A subscriber can also be durable, in which case messages are kept until the subscriber has acknowledged them.
 // !brokered.address.topic.longDescription:stop
 
+== Dynamic subscriptions on topics and wildcard behavior
+
+You can specify a dynamic subscription on topics in the brokered address space using wildcards. The wildcard behavior follows the MQTT syntax:
```
grs: I think I would avoid the term 'dynamic subscription' here if possible (I know I was the one who used it in discussion). I would say something like:

A client receiving from a topic address can specify a wildcard address with the topic address as the root. The wildcard behaviour follows...

Also maybe change the section title to just 'hierarchical topics and wildcards'

documentation/modules/con-standard-topic.adoc
```diff
@@ -9,4 +9,19 @@
 The topic address type supports the publish-subscribe messaging pattern where there are 1..N producers and 1..M consumers. Each message published to a topic address is forwarded to all subscribers for that address. A subscriber can also be durable, in which case messages are kept until the subscriber has acknowledged them.
 // !standard.address.topic.longDescription:stop
 
+== Dynamic subscriptions on topics and wildcard behavior
+
+You can specify a dynamic subscription on topics in the standard address space using wildcards. The wildcard behavior follows the MQTT syntax:
```
grs: As above, I would try and avoid the term dynamic subscription here.

documentation/modules/con-brokered-topic.adoc
jenmalloy: @grs, excellent, thank you! Will fix

documentation/modules/con-standard-topic.adoc
jenmalloy: Done

PR #1917, opened 2018-11-02
Remove the 'min' field from address space plans

lulf:
The 'min' field only applies to routers, and should not be dictated by
address space plans. Instead the plans should set the limits of how much
router credit addresses can use.

Instead, specify minimum router replicas in the standard infra config .router.minReplicas

@rgodfrey I think it would be good to make this change before 0.24.0 so that we don't have to maintain compatibility for this change in the future.


k-wall:
Removing brokers.min makes sense to me for the reasons you mention.   I'm not sure about the rename of `router.min`.   What's wrong with the name it has now?

lulf:
> Removing brokers.min makes sense to me for the reasons you mention. I'm not sure about the rename of router.min. What's wrong with the name it has now?

@k-wall 

Yes, I would want to remove it from routes, brokers, and 'aggregated' actually. Its only the 'router.min' that is in use right now, and I think it fits more naturally in the infraConfig than in the address space plan. The reason it is there is that we didn't have infra configs before.

If we're going to use AddressSpacePlans that shares the underlying infrastructure, it makes even less sense to keep 'min' there IMO.

PR #1920, opened 2018-11-02
Docs: address filtering behavior in the console

jenmalloy:


documentation/modules/ref-console-address-filtering.adoc
```diff
@@ -0,0 +1,18 @@
...
+
+The address name filtering feature in the {ProductName} Console uses regular expressions. Also, filters are cumulative.
+
+.{ProductName} Console address name filtering behavior
+[cols="34%a,33%a,33%a",options="header"]
+|===
+|To match... |Use this |Result
+|An expression exactly |Use a period followed by an expression: `.my` |The address `my`
+|The beginning of an expression |Use a caret followed by an expression: `^my` |All addresses that begin with `my`
+|All addresses containing an expression |Use a period followed by an asterisk and an expression: `.*my` |All addresses containing `my`
```
grs: you can actually just use the expression here

documentation/modules/ref-console-address-filtering.adoc
```diff
@@ -0,0 +1,18 @@
...
+[id='ref-console-address-filtering-{context}']
+= Using the {ProductName} Console address filtering
+
+The address name filtering feature in the {ProductName} Console uses regular expressions. Also, filters are cumulative.
+
+.{ProductName} Console address name filtering behavior
+[cols="34%a,33%a,33%a",options="header"]
+|===
+|To match... |Use this |Result
+|An expression exactly |Use a period followed by an expression: `.my` |The address `my`
```
grs: I'm not sure this is right. By 'exactly', do you mean match my but not amy or mya?

jenmalloy: @grs, yes, that is what I meant. I just tested it on my console, and it looks like to match "mya" you'd need to use `.my*` instead of `.my` ?

grs: I would expect 'my' to match 'mya' and 'amy' (anything with my in it). To match my and nothing else, I would use ^my$

documentation/modules/ref-console-address-filtering.adoc
```diff
@@ -0,0 +1,18 @@
...
+
+[id='ref-console-address-filtering-{context}']
+= Using the {ProductName} Console address filtering
+
+The address name filtering feature in the {ProductName} Console uses regular expressions. Also, filters are cumulative.
+
+.{ProductName} Console address name filtering behavior
+[cols="34%a,33%a,33%a",options="header"]
+|===
+|To match... |Use this |Result
```
ncbaratta: Change 'use this' to 'use' and remove use from the options below.  And maybe 'Result' should be 'Returns' or 'Results in' ... think of the headings as part of the sentence.

documentation/modules/ref-console-address-filtering.adoc
jenmalloy: @grs, got it. Did some further testing and got that result. Making edits now

documentation/modules/ref-console-address-filtering.adoc
jenmalloy: Done

PR #1921, opened 2018-11-02
Reduce number of routers to 1 by default

lulf:


address-space-controller/src/main/resources/addressspaceplans/020-AddressSpacePlan-standard-medium.yaml
```diff
@@ -14,7 +14,7 @@ uuid: 3610533c-db6e-11e8-a390-d45ddf138840
 addressSpaceType: standard
 resources:
 - name: router
-  min: 2.0
+  min: 1.0
```
calohmn: The `(minimum 2)` bit in the `longDescription` above should be changed as well. Same goes for the other AddressSpacePlan files.

PR #1925, opened 2018-11-03
Separate deploy and tests

kornys:


kornys:
@lulf I splitted deployment of enmasse and system tests into separated stages. Also added functionality to choose oc version by env var for prod testing jobs.

PR #1926, opened 2018-11-03
More license metadata automation

k-wall:
Move licenses.xml to within the licenses/ directory.
Generate a per-module `licenses.html` providing a human readable summary and link to the remote/local license.
Allow license artifact production to be surpressed (for reasons of speed during a developer's build/test cycle).

### DIFF
--- a/templates/Makefile
+++ b/templates/Makefile
@@ -76,7 +76,7 @@ component_install: replace_images
 
 ansible_install: component_install
 	cp -r $(PACKAGE_ANSIBLE_DIR) $(INSTALLDIR)/
-	gln -srf $(INSTALLDIR)/install/components $(INSTALLDIR)/ansible/playbooks/openshift/components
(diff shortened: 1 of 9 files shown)


templates/Makefile
```diff
@@ -76,7 +76,7 @@ component_install: replace_images
 
 ansible_install: component_install
 	cp -r $(PACKAGE_ANSIBLE_DIR) $(INSTALLDIR)/
-	ln -srf $(INSTALLDIR)/install/components $(INSTALLDIR)/ansible/playbooks/openshift/components
+	gln -srf $(INSTALLDIR)/install/components $(INSTALLDIR)/ansible/playbooks/openshift/components
```
lulf: Need an OS ifdef (or alternative command)

k-wall: Sorry - that local hack wasn't intended to be part of this PR.

PR #1943, opened 2018-11-06
Docs: issue 1942: add attribute

jenmalloy:


ncbaratta:
The problem is that that attribute is only in certain files.  We need to stop the practice of defining attributes only in specific files and come up with a way to have them all in attributes.adoc.  For now I'm not particular about what it's named, but this is something we do need to talk about before we end up with too much of a mess.

jenmalloy:
@lulf, I'll change the attribute name to KubePlatform here. We can tackle the other issue @ncbaratta brings up later

lulf:
Yes, we should split the OpenShift and Kubernetes books and use separate attribute files. This should fix the issue and avoid per-doc attributes longer term.

PR #1951, opened 2018-11-07
Patch secret if cert bundle has changed

lulf:
This allows address spaces to have annotations, labels and endpoints replaced. The certBundle endpoints also switch the certificate in the secret, allowing users to perform certificate rotation without downtime.

### DIFF
--- a/address-space-controller/src/main/java/io/enmasse/controller/auth/CertBundleCertProvider.java
+++ b/address-space-controller/src/main/java/io/enmasse/controller/auth/CertBundleCertProvider.java
@@ -8,13 +8,13 @@ import io.enmasse.address.model.AddressSpace;
 import io.enmasse.config.AnnotationKeys;
 import io.enmasse.config.LabelKeys;
 import io.fabric8.kubernetes.api.model.Secret;
+import io.fabric8.kubernetes.api.model.SecretBuilder;
 import io.fabric8.openshift.client.OpenShiftClient;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.util.HashMap;
 import java.util.Map;
-import java.util.Objects;
 
 public class CertBundleCertProvider implements CertProvider {
     private static final Logger log = LoggerFactory.getLogger(CertBundleCertProvider.class);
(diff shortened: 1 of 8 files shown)


address-space-controller/src/main/java/io/enmasse/controller/auth/CertBundleCertProvider.java
```diff
@@ -28,34 +28,40 @@ public CertBundleCertProvider(OpenShiftClient client) {
...
+                .addToLabels("app", "enmasse")
+                .endMetadata()
+                .withType("kubernetes.io/tls")
+                .withData(data)
+                .build();
+
+        Secret existing = client.secrets().inNamespace(namespace).withName(endpointInfo.getCertSpec().getSecretName()).get();
+        if (existing == null) {
+            client.secrets().inNamespace(namespace).createOrReplace(secret);
+        } else if (!existing.getData().equals(secret.getData())) {
```
k-wall: The API does not make is clear if existing.getData(), could return null, but the code would be robust to this possibility if you flipped the equality around.

address-space-controller/src/main/java/io/enmasse/controller/auth/CertBundleCertProvider.java
```diff
@@ -41,8 +41,8 @@ public void provideCert(AddressSpace addressSpace, EndpointInfo endpointInfo) {
             return;
         }
 
-        data.put("tls.key", tlsKey);
-        data.put("tls.crt", tlsCert);
+        data.put("tls.key", tlsKey.replace("\n", ""));
```
k-wall: I'm not sure I understand the intent by stripping the newlines?  I think the data of cert is PEM formatted, so the newlines (which are allowed to be an line ending convention) are required by the format.

address-space-controller/src/main/java/io/enmasse/controller/auth/CertBundleCertProvider.java
k-wall: Resolution looks good.

address-space-controller/src/main/java/io/enmasse/controller/auth/CertBundleCertProvider.java
```diff
@@ -28,34 +28,40 @@ public CertBundleCertProvider(OpenShiftClient client) {
...
+        }
+
+        if (tlsCert == null) {
+            log.warn("tlsCert not present, not providing cert for {}", endpointInfo.getServiceName());
+            return;
+        }
+
+        data.put("tls.key", tlsKey);
+        data.put("tls.crt", tlsCert);
+        log.info("Creating cert secret with certBundle input");
```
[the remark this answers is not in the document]
k-wall: resolution looks good.

address-space-controller/src/main/java/io/enmasse/controller/auth/CertBundleCertProvider.java
lulf: This is base64 encoded. If the input is taken as a newline formatted base64 (which is common if you use the base64 command line tools etc), comparing desired vs current state will always trigger a replace because kubernetes strips the newlines when creating the secret.

lulf: The input is the base64 encoded string of the PEM, i.e. almost encoded twice, which made me think maybe we should change the expected input to be the PEM itelf instead requiring it to be base64-encoded?

lulf: @k-wall I think the source of confusion here is that the input format was a base64 encoded string of the original PEM. But that is not really required, so I pushed a fix to allow the input to be PEM-formatted as well.

k-wall:
I wonder if a better approach would be for EnMasse’s CertSpec#validate to apply the same validation to `tlsKey` and `tlsCert` that a OpenShift TLS Secret would apply?  i.e. assert that it has to be Base64 encoded (PEM string).  This would give the user a consistent interface.   I would not support the adding of the un-base64 encoded PEM from CertSpec.

lulf:
@k-wall Added input validation (requiring base64 encoded) and CRLF and LF replace.

PR #1952, opened 2018-11-07
Allow more characters in MessagingUser resource

### LINKED ISSUE #1950: After provisioning from the catalog I cannot login to the console with my OpenShift user 0.23.0
**Describe the bug**
Using EnMasse 0.23.0 when provisioning from the catalog everything looks to have worked but I cannot login to the EnMasse console.

Digging further I see my OpenShift user has not been added the keycloak realm.

Looking at the keycloak controller I see the following exception

```

io.enmasse.user.model.v1.UserValidationFailedException: io.enmasse.user.model.v1.UserValidationFailedException: Invalid resource name 'mine.admin@example.com', must match ^[a-z]+([a-z0-9\-]*[a-z0-9]+\|[a-z0-9]*)\.[a-z0-9]+([a-z0-9\-]*[a-z0-9]+\|[a-z0-9]*)$
--
  | at io.enmasse.user.model.v1.User.validate(User.java:61)
  | at io.enmasse.user.keycloak.KeycloakUserApi.createUser(KeycloakUserApi.java:155)
  | at io.enmasse.keycloak.controller.KeycloakManager.onUpdate(KeycloakManager.java:114)
  | at io.enmasse.k8s.api.ResourceChecker.doWork(ResourceChecker.java:48)
  | at io.enmasse.k8s.api.ResourceChecker.run(ResourceChecker.java:40)
  | at java.lang.Thread.run(Thread.java:748)
  | Caused by: io.enmasse.user.model.v1.UserValidationFailedException: Invalid resource name 'mine.admin@example.com', must match ^[a-z]+([a-z0-9\-]*[a-z0-9]+\|[a-z0-9]*)\.[a-z0-9]+([a-z0-9\-]*[a-z0-9]+\|[a-z0-9]*)$
  | at io.enmasse.user.model.v1.UserMetadata.validate(UserMetadata.java:70)
  | at io.enmasse.user.model.v1.User.validate(User.java:58)

```

**To Reproduce**
1. Go to 'catalog'
2. Click on 'EnMasse standard and provision'
3. Once provisioned try to login it will fail
4. look at logs of Keycloak controller

**Expected behavior**
To be able to login with my OpenShift user

lulf:
Fixes #1950

### DIFF
--- a/user-api/src/main/java/io/enmasse/user/model/v1/UserSpec.java
+++ b/user-api/src/main/java/io/enmasse/user/model/v1/UserSpec.java
@@ -17,7 +17,7 @@ public class UserSpec {
     private final String username;
     private final UserAuthentication authentication;
     private final List<UserAuthorization> authorization;
-    private final Pattern usernamePattern = Pattern.compile("^[a-z0-9]+([a-z0-9_\\-]*[a-z0-9]+|[a-z0-9]*)$");
+    private final Pattern usernamePattern = Pattern.compile("^[a-z0-9]+([a-z0-9_@.\\-]*[a-z0-9]+|[a-z0-9]*)$");
 
     @JsonCreator
     public UserSpec(@JsonProperty("username") String username,
--- a/address-model-lib/src/main/java/io/enmasse/address/model/KubeUtil.java
+++ b/address-model-lib/src/main/java/io/enmasse/address/model/KubeUtil.java
@@ -11,13 +11,23 @@ import io.enmasse.config.AnnotationKeys;
  */
 public class KubeUtil {
     private static int MAX_KUBE_NAME = 63 - 3; // max length of identifier - space for pod identifier
-    private static final String pattern = "[^a-z0-9\\-]";
+    private static final String addressPattern = "[^a-z0-9\\-]";
+    private static final String usernamePattern = "[^a-z0-9\\-.@_]";
+
     public static String sanitizeName(String name) {
-        if (name == null) {
+        return sanitizeWithPattern(name, addressPattern);
+    }
+
+    public static String sanitizeUserName(String name) {
+        return sanitizeWithPattern(name, usernamePattern);
+    }
+
+    private static String sanitizeWithPattern(String value, String pattern) {
+        if (value == null) {
             return null;
         }
 
-        String clean = name.toLowerCase().replaceAll(pattern, "");
+        String clean = value.toLowerCase().replaceAll(pattern, "");
         if (clean.startsWith("-")) {
             clean = clean.replaceFirst("-", "1");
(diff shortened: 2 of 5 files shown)


k-wall:
If the intention is to allow email addresses to be used as user identifiers, the the restrictions would need to be relaxed further to be useful (_ for instance).   Furthermore, shouldn't we be basing the validation on what openshift considers to be a legal user identifier?

lulf:
@k-wall So I believe '@' be allowed in k8s resource names, but not '_'. Therefore if you wish to create users with special characters, you must do so in the .spec.username field (similar to Address metadata.name vs .spec.address).

For keycloak-controller, I added auto-sanitation of the .metadata.name field, but kept .spec.username as original. The .metadata.name.

lulf:
@k-wall Updated PR to support both @, _ and . in resource names. Keycloak-controller also sanitizes the usernames in case it is not using valid characters. Api server validation of user names is extended.

user-api/src/main/java/io/enmasse/user/model/v1/UserSpec.java
```diff
@@ -17,7 +17,7 @@
     private final String username;
     private final UserAuthentication authentication;
     private final List<UserAuthorization> authorization;
-    private final Pattern usernamePattern = Pattern.compile("^[a-z0-9]+([a-z0-9_\\-]*[a-z0-9]+|[a-z0-9]*)$");
+    private final Pattern usernamePattern = Pattern.compile("^[a-z0-9]+([a-z0-9_@.\\-]*[a-z0-9]+|[a-z0-9]*)$");
```
k-wall: Are we correct to still disallow uppercase?  What if my organisation uses uppercase usernames?

lulf: All usernames in keycloak are lowercased, so I'm not sure if we should allow it.

k-wall: I was thinking about the case where the openshift user id (which may or may not be ultimately stored in a keycloak instance), happens to include uppercase.  May be this is not valid concern.

lulf: @k-wall The federated user fields have no restrictions on the name. This affects only the name of the user we create in keycloak. It can still reference openshift users with any characters in them.

In the case of the realm admin user, the .metadata.name and .spec.username will be lowercased by keycloak-controller, so that it will be accepted by the user api.

k-wall: Thanks for the clarification.

PR #1955, opened 2018-11-08
Allow host volumes to be recycled on 3.11

lulf:


k-wall:
Is this change harmless on earlier openshift versions?

PR #1956, opened 2018-11-08
Allow browser security headers to be configured in keycloak-config

### LINKED ISSUE #1953: Allow setting the X-FRAME-OPTIONS and Content-Security-Policy for auto-created realms
## why

We have a Walkthroughs webapp that takes a user through multiple tasks for integrating different services to solve a larger problem.
EnMasse is one of those services.
We've been progressing some spikes around showing the different services alongside the walkthrough steps in an iframe. e.g. when I get to the task that has EnMasse bits, we open the EnMasse console in an iframe.

## Proposal

Allow configuring what values are set for the X-FRAME-OPTIONS and Content-Security-Policy whenever a new Realm is created for an Address Space.
See https://www.keycloak.org/docs/3.2/server_admin/topics/threat/clickjacking.html for more info

> What are the components involved in this change and how will they change?

Probably the keycloak controller, when a realm is created?

> Describe any new components introduced by this change

None proposed.

## Testing

> What is the test plan for the feature/change?  

Configure EnMasse to set the headers to specific values .
Create a new Address Space (which will create a new Realm in Keycloak with those header values)
Create a simple webpage that has an iframe with src set to this EnMasse instance.
Verify the iframe loads correctly.

## Documentation

> What documentation will be created for this feature? 

How to configure EnMasse to set these default realm settings

## Tasklist

> Enumerate the tasks required to complete the feature

I'll leave this to someone more familiar with the project

lulf:
Closes #1953

It looks like keycloak uses the following format for headers:

xContentTypeOptions, xRobotsTag, xFrameOption, xXSSProtection, contentSecurityPolicy etc. I didn't like the thought of converting, so users of this feature would need to know what keycloak expects... wdyt?


david-martin:
> Does the above format look ok?

Sure, I can work with those.

keycloak-controller/src/main/java/io/enmasse/keycloak/controller/KeycloakRealmParams.java
```diff
@@ -0,0 +1,97 @@
...
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+import java.util.*;
+
+public class KeycloakRealmParams {
+
+    private static final Logger log = LoggerFactory.getLogger(KeycloakRealmParams.class);
+    private static final ObjectMapper mapper = new ObjectMapper();
+    public static final KeycloakRealmParams NULL_PARAMS = new KeycloakRealmParams(null, null, null, null);
```
k-wall: NULL_PARAMS.getBrowserSecurityHeaders() will NPE.   I also prefer a defensive copy of mutable structures.  So how about changing the constructor to:

this.browserSecurityHeaders = browserSecurityHeaders == null ? Collections.emptyMap() : new HashMap<>(browserSecurityHeaders);

keycloak-controller/src/main/java/io/enmasse/keycloak/controller/KeycloakRealmParams.java
```diff
@@ -0,0 +1,97 @@
...
+            }
+        }
+
+        return new KeycloakRealmParams(identityProviderUrl, identityProviderClientId, identityProviderClientSecret, browserSecurityHeaders);
+    }
+
+    @Override
+    public String toString() {
+        return "{identityProviderUrl=" + identityProviderUrl + "," +
+                "identityProviderClientId=" + identityProviderClientId + "}";
```
k-wall: For completeness, add the security headers.

keycloak-controller/src/main/java/io/enmasse/keycloak/controller/KeycloakRealmParams.java
```diff
@@ -0,0 +1,97 @@
...
+        if (browserSecurityHeadersString != null) {
+            try {
+                ObjectNode data = mapper.readValue(browserSecurityHeadersString, ObjectNode.class);
+                Iterator<String> it = data.fieldNames();
+                while (it.hasNext()) {
+                    String key = it.next();
+                    browserSecurityHeaders.put(key, data.get(key).asText());
+                }
+            } catch (IOException e) {
+                log.warn("Error parsing browserSecurityHeaders, skipping");
```
k-wall: I'd chain the exception.  The Jackson Exception will be sufficient to help the user figure out what they have done wrong.

2018-11-09 15:52:14 WARN  KeycloakRealmParams:53 - Error parsing browserSecurityHeaders, skipping
com.fasterxml.jackson.core.JsonParseException: Unrecognized token 'hello': was expecting ('true', 'false' or 'null')
 at [Source: (String)"hello"; line: 1, column: 11]

lulf:
@k-wall did you try this format: xFrameOptions
?

k-wall:
With the option name `xFrameOptions` all worked well for a new addressspace.  There was a defect when updating an existing addressspace with new headers.  After fixing that all seemed well.

lulf:
@k-wall I have not found keycloak docs for that format, although it looks like the following rules apply for converting headers:

* Lowercase first character
* Uppercase each character after the `-` character.

Since this is probably not going to be a widely used feature, I think whoever sets these headers need to know the format keycloak expects.

k-wall:
@lulf 

> Since this is probably not going to be a widely used feature, I think whoever sets these headers need to know the format keycloak expects.

That sounds reasonable to me.

PR #1967, opened 2018-11-09
Docs: issue-1691: document known issue with durable subs and topics

jenmalloy:


documentation/modules/con-brokered-topic.adoc
```diff
@@ -9,6 +9,8 @@
 The topic address type supports the publish-subscribe messaging pattern in which there are 1..N producers and 1..M consumers. Each message published to a topic address is forwarded to all subscribers for that address. A subscriber can also be durable, in which case messages are kept until the subscriber has acknowledged them.
 // !brokered.address.topic.longDescription:stop
 
+NOTE: Because subscriptions created after messages have been sent to the topic do not currently work properly, you must create the subscription before you create the topic.
```
grs: We don't define a subscription type in the brokered space, so this can be removed.

documentation/modules/con-standard-topic.adoc
```diff
@@ -9,6 +9,8 @@
 The topic address type supports the publish-subscribe messaging pattern where there are 1..N producers and 1..M consumers. Each message published to a topic address is forwarded to all subscribers for that address. A subscriber can also be durable, in which case messages are kept until the subscriber has acknowledged them.
 // !standard.address.topic.longDescription:stop
 
+NOTE: Because subscriptions created after messages have been sent to the topic do not currently work properly, you must create the subscription before you create the topic.
+
```
grs: You can't actually do that. The topic must exist before the subscription is created. I think what we should say here is something like: If you create a subscription on the topic, any senders to that topic must specify the 'topic' capability

documentation/modules/con-brokered-topic.adoc
jenmalloy: Removed

documentation/modules/con-standard-topic.adoc
jenmalloy: @grs Edited note in latest commit

PR #1974, opened 2018-11-12
Change keycloak deployment strategy

lulf:


k-wall:
Could you describe the problem being solved, and how this change solves it so I can give a meaningful review.   Having the history of why a change was made is always helpful too :)

I see we are switching from [default deployment](https://docs.openshift.com/container-platform/3.11/dev_guide/deployments/deployment_strategies.html#strategies
) (which  I read is a rolling one) to an explicit [recreate](https://docs.openshift.com/container-platform/3.11/dev_guide/deployments/deployment_strategies.html#recreate-strategy).

lulf:
@k-wall, the reason for this change is that with the Default/rolling update strategy, an upgrade can cause 2 instances of Keycloak sharing the same file system (and therefore h2 database instance). This causes an error in the upgrade tests where the upgraded Keycloak got stuck with errors accessing the database. Setting rollingUpdate explicitly to null is required to avoid error in oc apply with this change

k-wall:
> Setting rollingUpdate explicitly to null is required to avoid error in oc apply with this change. This is because the oc client is not aware of the default values.

btw.  #24198

PR #1976, opened 2018-11-12
Restore two router plans

grs:


lulf:
@grs Could you rebase the branch on master? I think we can merge this @k-wall @rgodfrey ?

grs:
@lulf @kwall @rgodfrey if someone wants to approve this I can merge it; else we can delete it and revisit another time. (All it does is revert a previous change)

PR #1983, opened 2018-11-14
Wait for all resources to be deleted

lulf:
Issue #1982


kornys:
@lulf let me try it with marathon test suite

PR #1985, opened 2018-11-14
Create single broker artifact with configuration and launch scripts

lulf:
* Add plugin module containing configuration and scripts
* Add assembly in plugin that bundles config, scripts and plugin jars
* Reuse scripts from base image instead of providing our own
* Copy jars from launch script


lulf:
@k-wall @vbusch Got it working now, ready for review :)

k-wall:
nom-shrinkwrap has crept into the patch.

PR #1988, opened 2018-11-14
Prevent the StandardController from overwriting address configmaps

k-wall:
@lulf I took an alternative approach to solving this problem to one that was discussed a few weeks ago.  I tried caching state locally but I couldn't get an implementation that wasn't ugly.   This approach guards the standard controller from acting on an address if its representation is stale.  This staleness may occur owing to the relative actions of the refresh/watchers.  This approach is also appealing because should we start to support patching of addresses (say to allow the user to add their own annotations), then this approach will mean that the standard controller cannot overwrite these changes.

I left some commented out code only for discussion purposes.  My intention is to remove it.

Also there are some analogous changes which I think need to be reflected into the address-space API.

### DIFF
--- a/k8s-api/src/main/java/io/enmasse/k8s/api/ConfigMapAddressApi.java
+++ b/k8s-api/src/main/java/io/enmasse/k8s/api/ConfigMapAddressApi.java
@@ -66,10 +66,7 @@ public class ConfigMapAddressApi implements AddressApi, ListerWatcher<ConfigMap,
                 builder.setUid(configMap.getMetadata().getUid());
             }
 
-// resourceVersion is server assigned when the update is received, so any value from the config.json will always be stale.
-//            if (address.getResourceVersion() == null) {
-//                builder.setResourceVersion(configMap.getMetadata().getResourceVersion());
-//            }
+            builder.setResourceVersion(configMap.getMetadata().getResourceVersion());
 
             if (address.getCreationTimestamp() == null) {
                 builder.setCreationTimestamp(configMap.getMetadata().getCreationTimestamp());
--- a/k8s-api/src/main/java/io/enmasse/k8s/api/ConfigMapAddressApi.java
+++ b/k8s-api/src/main/java/io/enmasse/k8s/api/ConfigMapAddressApi.java
@@ -19,6 +19,8 @@ import io.fabric8.openshift.client.NamespacedOpenShiftClient;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.io.IOException;
+import java.io.UncheckedIOException;
 import java.time.Clock;
 import java.time.Duration;
 import java.util.*;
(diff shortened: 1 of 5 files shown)


k8s-api/src/main/java/io/enmasse/k8s/api/ConfigMapAddressApi.java
```diff
@@ -151,21 +156,18 @@ private ConfigMap create(Address address) {
                 .addToLabels(LabelKeys.INFRA_UUID, infraUuid)
                 .addToLabels(LabelKeys.INFRA_TYPE, "any")
                 .addToAnnotations(address.getAnnotations())
+// I think that resourceVersion is always assigned by the server, so I think doing this unconditionally is safe.
+                .withResourceVersion(address.getResourceVersion())
```
lulf: I'm wondering if this will work in the initial create via API server, which also calls this create method, and will cause a .withResourceVersion(null) in that case.

k-wall: Looking at the implementation, it seemed to have no affect, but maybe it was better as it was.  I'm inclined to do the same to io.enmasse.k8s.api.ConfigMapAddressSpaceApi#create.

k8s-api/src/main/java/io/enmasse/k8s/api/ConfigMapAddressApi.java
```diff
@@ -64,9 +66,10 @@ private Address getAddressFromConfig(ConfigMap configMap) {
                 builder.setUid(configMap.getMetadata().getUid());
             }
 
-            if (address.getResourceVersion() == null) {
-                builder.setResourceVersion(configMap.getMetadata().getResourceVersion());
-            }
+// resourceVersion is server assigned when the update is received, so any value from the config.json will always be stale.
```
lulf: I think that with this change, address setResourceVersion is never called, so the config.json will never contain that info, which means lockResourceVersion(address.getResourceVersion()) will always be called with 'null' as the argument.

How it has worked previously is that address resourceVersion always gets set on deserialization  like above, so it will always match the resourceVersion of the underlying configmap. So i would instead remove the 'if address.getResourceVersion() == null), because it should always be set to the configmap resourceversion IMO.

k-wall: You are indeed right and it was the change that I had been testing.   I reverted too much when adding the comments just before making the PR.

k8s-api/src/main/java/io/enmasse/k8s/api/ConfigMapAddressApi.java
```diff
@@ -132,9 +132,11 @@ public boolean replaceAddress(Address address) {
             return false;
         }
         ConfigMap newMap = create(address);
-        if (newMap != null) {
-            client.configMaps().inNamespace(namespace).withName(name).replace(newMap);
-        }
+        client.configMaps()
+                .inNamespace(namespace)
+                .withName(name)
+                .lockResourceVersion(address.getResourceVersion())
```
lulf: address.getResourceVersion() may be null if replaced via api server

PR #1993, opened 2018-11-15
Remove vertx as a dependency for address space controller

lulf:
* Use com.sun.net.httpserver instead of vertx
* Install signal handlers to improve container shutdown


address-space-controller/src/main/java/io/enmasse/controller/AddressSpaceController.java
```diff
@@ -109,8 +103,7 @@ public void start(Future<Void> startPromise) throws Exception {
         controllerChain.addController(authController);
```
k-wall: I think there is now nothing calling controllerChain.start()

lulf:
@k-wall Explicit closing is probably more portable, so I'm fine with that change.

lulf:
@k-wall Had to add a missing dependency on resteasy-jackson mapping that was implicit from vertx earlier ( i think)

PR #1996, opened 2018-11-15
Remove redundant calls to inNamespace() when interacting with openshift API.

k-wall:
Also make use of openshift API the same between ConfigMapAddressApi/ConfigMapAddressSpaceApi implementations.


k-wall:
The belt and braces in `io.enmasse.k8s.api.ConfigMapAddressSpaceApi#replaceAddressSpace` results from https://github.com/fabric8io/kubernetes-client/issues/1264

PR #1999, opened 2018-11-19
Check opened files in artifact dir before archive

kornys:
Should avoid the problem with archiving artifacts in jenkins.

ERROR: Failed to archive artifacts: artifacts/**
java.io.IOException: This archives contains unclosed entries.


jenkins-functions.groovy
```diff
@@ -56,8 +56,9 @@ def postAction(String coresDir, String artifactDir) {
     junit testResults: '**/TEST-*.xml', allowEmptyResults: true
     //archive test results and openshift logs
     archiveArtifacts artifacts: '**/TEST-*.xml', allowEmptyArchive: true
-    archiveArtifacts artifacts: 'templates/install/**', allowEmptyArchive: true
+    archiveArtifacts artifacts: 'templates/build/**', allowEmptyArchive: true
     sh "sudo ./systemtests/scripts/compress_core_dumps.sh ${coresDir} ${artifactDir}"
+    sh 'sleep 60' //wait before archive artifact
```
lulf: I don't understand why this is needed. The commands in the `compress_core_dumps.sh` script are run synchronously.

kornys: Because job sometimes fails with exception in archive artifacs due to unclosed files

PR #2010, opened 2018-11-20
[System Tests] Fix array splicing to prevent duplicate address creati…

k-wall:
…on from tail of array


k-wall:
"service unavailable" suggests an overload - just one with different symptoms to mine.   With #2013, I find that the test successfully creates all 3900 addresses without issue (even relaxing the health checks).   Can you conveniently repeat your test with that patch applied?

PR #2012, opened 2018-11-20
Switch order of commands to ensure uninstall is successfull

lulf:


ncbaratta:
I might be missing something - but this looks exactly the same ... what has changed?

lulf:
@ncbaratta the steps are reordered, i.e. the delete project/namespace step is moved to the end instead of as the 2nd step.

PR #2015, opened 2018-11-21
new marathon systemtest testCreateHighAddressCountCheckStatusDeleteLong

RH-sdavey:
* new marathon systemtest testCreateHighAddressCountCheckStatusDeleteLong      
  * removed outdated/unused systemtest/marathon/PlansMarathonTest.java


systemtests/src/test/java/io/enmasse/systemtest/marathon/PlansMarathonTest.java
```diff
@@ -1,143 +0,0 @@
-/*
```
kornys: Please don't delete this file, we should keep these tests.

RH-sdavey:
Test is now failing on my local testing server. Messages are not able to be sent and received on different addresses on each test run, seems to be no fault of the test, but with the RAM limitations on the testing node. I will merge this test for now, and if its failing in CI, I will disable the test while working on a possible fix.

PR #2016, opened 2018-11-21
Issue 2000 network policies

lulf:


lulf:
@k-wall Addressed the in-code comments.

Regarding uninstall, the networkpolicy objects are namespaced and part of the enmasse-infra namespace, so they will be deleted in the current uninstall docs.

lulf:
@k-wall Example infra config

```
kind: StandardInfraConfig
...
spec:
   networkPolicy:
      ingress:
         - from:
              - podSelector:
                   my: label
```

address-space-controller/src/main/java/io/enmasse/controller/NetworkPolicyController.java
```diff
@@ -0,0 +1,153 @@
...
+        if (infraConfig != null) {
+            networkPolicy = infraConfig.getNetworkPolicy();
+        }
+
+        if (addressSpace.getNetworkPolicy() != null) {
+            networkPolicy = addressSpace.getNetworkPolicy();
+        }
+
+        String infraUuid = addressSpace.getAnnotation(AnnotationKeys.INFRA_UUID);
+        List<Service> services = kubernetesClient.services().withLabel(LabelKeys.INFRA_UUID, infraUuid).list().getItems();
```
k-wall: The service declaration should be moved to within the first branch of the if block.

k-wall:
I'm having problems with an end to end test.  I keep meeting deserialisation issues.   The example document you have above doesn't match the structure of the Java code.  In the structure above the `networkPolicy` has an array of `networkPolicyIngressRule`, but the Java code the `NetworkPolicy` has a single `NetworkPolicyIngressRule` instance.

lulf:
> I'm having problems with an end to end test. I keep meeting deserialisation issues. The example document you have above doesn't match the structure of the Java code. In the structure above the `networkPolicy` has an array of `networkPolicyIngressRule`, but the Java code the `NetworkPolicy` has a single `NetworkPolicyIngressRule` instance.

No, there is only one `ingress` field in the yaml above. According to json2yaml.com, it is equivalent to the following json:

```
{
  "spec": {
    "networkPolicy": {
      "ingress": [
        {
          "from": [
            {
              "podSelector": {
                "my": "label"
              }
            }
          ]
        }
      ]
    }
  }
}
```

k-wall:
It still looks wrong to me..  I am definitely running the latest code from your branch. Take a look at the attach test.  The first test case   (which includes your YAML translated to JSON) fails with a deserialisation error. In the second, I've modified the input to give ingress an object rather than an array.

```
com.fasterxml.jackson.databind.exc.MismatchedInputException: Cannot deserialize instance of `io.fabric8.kubernetes.api.model.networking.NetworkPolicyIngressRule` out of START_ARRAY token
 at [Source: (String)"{
  "spec": {
    "networkPolicy": {
      "ingress": [
        {
          "from": [
            {
              "podSelector": {
                "my": "label"
              }
            }
          ]
        }
      ]
    }
  }
}"; line: 4, column: 18] (through reference chain: io.enmasse.admin.model.v1.StandardInfraConfig["spec"]->io.enmasse.admin.model.v1.StandardInfraConfigSpec["networkPolicy"]->io.enmasse.admin.model.v1.NetworkPolicy["ingress"])
```

[InfraConfigDeserializerTest.txt](https://github.com/EnMasseProject/enmasse/files/2617331/InfraConfigDeserializerTest.txt)

lulf:
@k-wall You are right, i was confusing the examples from the NetworkPolicy docs[1] where ingress is an array, with the format expected by 'our' NetworkPolicy. The format in the deserialiseSucceeds is the right format, sorry for misleading you.

[1] https://kubernetes.io/docs/concepts/services-networking/network-policies/

k-wall:
> Regarding the leaks, does it print the message only once or each time the controller is run? I'll do a test as well to check

It appears to be a leak per request. The leak reports seem to collate with GC cycles, so you may not see the leak report immediately.

lulf:
@k-wall pushed an update that should comply with the schema

address-model-lib/src/test/java/io/enmasse/admin/model/v1/InfraConfigDeserializerTest.java
```diff
@@ -0,0 +1,45 @@
...
+    public void deserializeNetworkPolicy() throws Exception {
+
+
+        String str = "{\n" +
+                "  \"spec\": {\n" +
+                "    \"networkPolicy\": {\n" +
+                "      \"ingress\": {\n" +
+                "        \"from\": [\n" +
+                "          {\n" +
+                "            \"podSelector\": {\n" +
```
k-wall: Missing `matchLabels ` ..

- podSelector:
          matchLabels:
            my: label3

PR #2017, opened 2018-11-21
Fixed addres space metrics

robshelly:
Removed Prometheus anti-pattern of changing labels from address space metrics
Added address space metrics to Grafana dashboard


address-space-controller/src/main/java/io/enmasse/controller/ControllerChain.java
```diff
@@ -114,18 +114,27 @@ public void onUpdate(List<AddressSpace> resources) throws Exception {
...
+        
         metrics.reportMetric(new Metric(
                 "address_spaces_ready",
                 "Address Spaces Ready",
                 MetricType.gauge,
-                healthValues));
+                readyValues));
+        metrics.reportMetric(new Metric(
+                    "address_spaces_pending",
+                    "Address Spaces Pending",
```
lulf: I'm not a native english writer/speaker, but is pending the right word to use? To me, pending means it has been created and is waiting to be ready, whereas in this case it can move to pending even after it has been running successfully for a while.

lulf:
@robshelly The build fails in a unit test in the address space controller, probably because there is an additional metric now.

lulf:
@robshelly  @vbusch Added a commit to fix the unit test

PR #2032, opened 2018-11-23
Fix a possible NPE, the value can only be null at this point

ctron:


ctron:
> This should be userNameWithAddressSpace to match the previous value

Indeed. Fixed.

PR #1997, opened 2018-11-16
Remove vertx-proton dependency in standard-controller

lulf:
* Rewrite ProtonRequestClient to use Proton-J directly
* Add shutdown hooks to standard controller for graceful shutdown
* Use com.sun.httpserver instead of vertx for health endpoint

PR #2034, opened 2018-11-23
Add graceful shutdown of standard controller

lulf:
Takes the shutdown bits from #1997 and removes some unnecessary vertx use.


k-wall:
The changes are not compiling for me.  

```
[ERROR] /Users/keith/src/forks/enmasse/standard-controller/src/main/java/io/enmasse/controller/standard/StandardController.java:[65,13] cannot find symbol
[ERROR]   symbol:   class HTTPServer
[ERROR]   location: class io.enmasse.controller.standard.StandardController
[```

lulf:
@k-wall must have forgot to re-add a file

lulf:
@k-wall added missing file

PR #2035, opened 2018-11-24
Issue 1595 - display warning if connection to the server is lost

k-wall:
This PR adds a modal dialogue that pops up if the connection to the server is lost.   If the connection reestablishes itself the modal will automatically pop-down.  Scenarios where thus might happen are if the admin pod was restarted, or the user's internet connection dropped temporarily. The user currently has no way to dismiss the dialogue and cannot interact with the app whilst it is popped up.

serverconnectionlost

My original intent was to offer a [Logout] button.  However, currently the `/logout` url relies on relies on the server (agent) being available.

I think an improvement would be to allow the infra config to specify a site specific `logoutUrl`.  Currently we just redirect back to keycloak (or openshift) but this probably is not too helpful in terms of workflow. The deployer would configure `logoutUrl` to suitable landing page (outside of EnMasse).  EnMasse would direct to this logout url via a [keyclock logoff redirect](https://www.keycloak.org/docs/3.3/securing_apps/topics/oidc/java/logout.html).

If we had `logoutUrl`, and the server side made it available to the client, this facility could be used by this change.

### DIFF
--- a/agent/www/index.html
+++ b/agent/www/index.html
@@ -63,18 +63,12 @@
             Server Connection Lost
           </p>
           <p>
-            This may mean that a server component is being restarted. If this is the case, the console
-            will reconnect automatically. <!--Alternatively, the address space to which the console was connected
-            has been deleted, if this is the case, click the Logout button to end the session.-->
+            This may mean that the address space to which the console is connected has been permanently deleted, or it
+            could be indicative of a transient issue such as a network problem or server component restart.  If the
+            issue is transient, the console will reconnect automatically once the problem is resolved.
           </p>
         </div>
       </div>
-      <!--
-      TODO: if the address space is deleted we have no server to handle the logout and no location to direct the browser to.
-      <div class="modal-footer">
-        <a href="logout" class="btn btn-default">Logout</a>
-      </div>
-      -->
     </div>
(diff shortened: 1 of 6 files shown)


agent/www/index.html
```diff
@@ -51,6 +51,37 @@
...
+        <h4 class="modal-title" id="myErrorDialogLabel">Server Connection Lost</h4>
+      </div>
+      <div class="modal-body">
+        <div>
+          <p class="lead pficon pficon-error-circle-o">
+            Server Connection Lost
+          </p>
+          <p>
+            This may mean that a server component is being restarted. If this is the case, the console
+            will reconnect automatically. <!--Alternatively, the address space to which the console was connected
```
lulf: Perhaps 'This may mean that a server component is being restarted or deleted. If it is restarted, the console will reconnect automatically.' ?

k-wall: I don't think server component being delete makes sense to the user.
How about?

```
This may mean that the address space to which the console is connected has been permanently deleted, or it
could be indicative of a transient issue such as a network problem or server component restart.  If the
issue is transient, the console will reconnect automatically once the problem is resolved.
```

PR #2041, opened 2018-11-27
Upgrade JBoss loggin processor as the current version does not work w…

dejanb:
…ell with JDK > 8


dejanb:
The problem is described in https://issues.jboss.org/browse/LOGTOOL-105

PR #2051, opened 2018-11-28
Set initial router replicas

lulf:


kornys:
@lulf you have to increase timeouts for create address space like in downstream tests

lulf:
@rgodfrey @kornys The test failures were actually a bug in kubernetes-client that was unable to load template objects containing parameterized integer fields (complaining cannot parse ${ROUTER_INITIAL_REPLICAS} as integer). I've updated this PR with a workaround to set the initial replicas.

PR #2061, opened 2018-11-30
Create resource if patch fails

lulf:


lulf:
@kornys the failing test should be fixed now. It exposed a bug that was surfaced by this initial commit, so good that we found it now.

PR #2063, opened 2018-11-30
Never call onUpdate on watches before items has received the initial …

### LINKED ISSUE #2025: Address space serialization/deserialization issue can cause address spaces to be deleted
**Describe the bug**
If an exception occurs during deserialization in the ResourceChecker, one can end up with 0 address space being passed to the controll loop, in which case the address spaces get deleted

**To Reproduce**
1. Create address space
2. Wait for address space to be created
3. Edit address space configmap and inject syntax error
4. Address space gets deleted

**Expected behavior**
Be more conservative and don't delete address space.

lulf:
…update

This could cause the address-space-controller and standard-controller
loops to be invoked with an empty list of resources in the case of
parsing errors, causing them to delete infrastructure.

Closes #2025


lulf:
Merging this as the tests passsed but failed on artifact collection

PR #2064, opened 2018-11-30
Add uninstall playbook

### LINKED ISSUE #1297: Create an uninstall ansible playbook
As we have a install playbook, it would be great to have an uninstall that takes the same inventory and removes everything from the openshift cluster.

lulf:
Closes #1297


kornys:
@lulf playbook works fin, but one note we should make note in doc about this playbook and how it run (inventory/extra vars)

kornys:
@lulf another note: when I install enmasse, and create an address space and then run uninstall playbook, it stuck in delete project phase and project is marked for deletion but it isn't deleted.

PR #2070, opened 2018-12-03
Fix test gaps

kornys:


kornys:
@lulf last commit should fix the client tests fail from PR build

Issue #1280, opened 2018-05-29
Add support for watching addressspaces and addresses to api server


lulf:
This is required in order to use the API server by our controller components. This is a required step in the direction of introducing alternative persistence for address spaces and addresses.

secondsun:
I'm working on an operator that uses enmasse and having this would be really useful.  We'd like to be able to recreate queues / topics if they are deleted.  Currently we can, but something else has to trigger the reconcile loop, and I would like to be more reactive.

secondsun:
Update: So this problem is actually really bad if you use enmasse in a golang operator client.  In k8s watches are sent to controllers with a truthy `watch` as a query parameter and the response is a stream of `WatchEvents`.  In the enmasse controller instead the `$RESOURCEList` result is returned (ie AddressList, AddressSpaceList, etc).  In client go this is reported as a error and the goroutine is relaunched.  This relaunched goroutine immediately fails and logs an error and relaunches.  Infinitely.

Additionally if you set an owner and controller to Addresses etc then when those resources are created in your namespace this behavior kicks off as well.

kornys:
api-server was removed from enmasse

PR #2079, opened 2018-12-04
Add support for listing all address spaces if allowed

### LINKED ISSUE #2074: Unable to use generated Kubernetes Go client informer with "AddressSpace" resource
**Describe the bug**

It look to me as if it is not possible to use the Kubernetes generated Go client information implementation to list `AddressSpace` resources.

Instantiating a new informer fails with the follow error when refreshing:

~~~
E1203 14:46:50.687772 1 reflector.go:134] github.com/enmasseproject/enmasse/pkg/client/informers/externalversions/factory.go:107: Failed to list *v1alpha1.AddressSpace: RESTEASY003210: Could not find resource for full path: https://172.30.0.1:443/apis/enmasse.io/v1alpha1/addressspaces?limit=500&resourceVersion=0 
~~~

**To Reproduce**
1. Generate Go client resources
1. Instantiate Informer and list

**Expected behavior**

Kubernetes generated Go resources seems to support this operation. My expectation would be that EnMasse supports this as well.

lulf:
Closes #2074


ctron:
So listing seems to work for me. However it looks like I indeed need the "Watch" functionality as well.

lulf:
@ctron Watching is a more significant effort, so I would suggest you go ahead with 'polling' using listing for now until #1280  is scheduled for work.

PR #2086, opened 2018-12-05
Fix stale address update

lulf:


address-model-lib/src/main/java/io/enmasse/address/model/v1/AddressV1Serializer.java
```diff
@@ -13,6 +13,7 @@
 import io.enmasse.address.model.Address;
 
 import java.io.IOException;
+import java.util.HashMap;
```
vbusch: I'm not seeing why this imported is needed.

PR #2090, opened 2018-12-06
Check init image also

kornys:


systemtests/scripts/wait_until_up.sh
```diff
@@ -23,26 +23,30 @@ fi
...
             else
-                info "Pod ${pod_id} is not upgraded, current image: ${image}"
+                info "Pod ${pod_id} is not upgraded, current image: ${image}, initImage: ${initImage}"
             fi
         else
-            upgraded="true"
+            imageUpgraded="true"
         fi
-        if [[ "${ready}" == "false" ]] || [[ "${upgraded}" == "false" ]]
+        if [[ "${ready}" == "false" ]] || [[ "${imageUpgraded}" == "false" ]] && [[ "${initImageUpgraded}" == "false" ]]
```
lulf: Shouldn't this be a || b || c and not a || b && c ?

kornys: yes, it should, will fix it

PR #2092, opened 2018-12-07
Add support for selecting storage class for broker volumes

lulf:
This will allow service admins to select which storage class to use for broker volumes.


lulf:
@vbusch sorry, should be fixed now

Issue #2100, opened 2018-12-10
Pooled addresses are removed in scale down process


kornys:
**Describe the bug**
In progress of scale down of pooled broker statefull set, broker pods with active addresses are deleted.

**To Reproduce**
1. Create a lot of addresses to scale up pooled brokers (for example to 4 pods)
2. Delete **first** half of addresses and wait for scale down (for example to 2 pods)

**Expected behavior**
enmasse should keep broker pods with not deleted addresses 

```
oc get address many-addresses-standard.xxs-queue-2000 -o wide
NAME                                     ADDRESS          ADDRESSSPACE              TYPE      PLAN               READY     PHASE     AGE       STATUS
many-addresses-standard.xxs-queue-2000   xxs-queue-2000   many-addresses-standard   queue     pooled-xxs-queue   false     Active    2h        Address xxs-queue-2000 is missing active autoLink (active in dirs: [])
```

lulf:
CC @calohmn

lulf:
Fixed in #2101

PR #2101, opened 2018-12-11
Scale horizontally using multiple deployments

lulf:
This changes the scaling model of pooled brokers from single deployment
to multiple deployments. This is the first step in providing HA brokers.

This also fixes issue #2100 where brokers are scaled down with
addresses still assigned to them.


standard-controller/src/main/java/io/enmasse/controller/standard/AddressProvisioner.java
```diff
@@ -493,4 +481,22 @@ private BrokerInfo(String brokerId, double credit) {
         }
 
     }
+
+    public static class BrokerIdGenerator {
+        private final Random random = new SecureRandom();
```
rgodfrey: Does this need to be a SecureRandom or just a Random?  I don't know if it is still an issue but using too many SecureRandoms used to cause pauses while the JVM waited for entropy

systemtests/src/main/java/io/enmasse/systemtest/TestUtils.java
```diff
@@ -87,7 +87,7 @@ public static void waitForNReplicas(Kubernetes kubernetes, int expectedReplicas,
     public static void waitForNBrokerReplicas(Kubernetes kubernetes, String infraUuid, int expectedReplicas, boolean readyRequired,
                                               Destination destination, TimeoutBudget budget, long checkInterval) throws InterruptedException {
         String statefulsetName = destination.getDeployment().startsWith("broker") ?
-                String.format("broker-pooled-%s", infraUuid) : destination.getDeployment();
+                String.format("broker-%s-0", infraUuid) : destination.getDeployment();
```
kornys: why is there suffix "-0" ?

standard-controller/src/main/java/io/enmasse/controller/standard/AddressProvisioner.java
lulf: Fixed

systemtests/src/main/java/io/enmasse/systemtest/TestUtils.java
lulf: Fixed. It now resolves the actual cluster id of the destination instead

Issue #2107, opened 2018-12-11
API server drops ownerReference information


ctron:
**Describe the bug**

Creating a new addressspace/user, the API server silently drops the provided `ownerReference` information.

**To Reproduce**
1. Create a new resource with the `ownerReference` set
2. Read the resource

**Expected behavior**

Retain the `ownerReference` information.

PR #2109, opened 2018-12-11
[#2107] Use the builder and common fabric8 base classes for User

ctron:
This change refactors the "User" model to use the generated builders
instead of custom ones. Re-use the fabric8 ObjectMeta definition. And
simplifies the JSON parts a bit.

This also prepares the way to store ownerReferences. I have an initial version for that with the `MessagingUser`, which seems to work with the IoT project.


address-model-lib/pom.xml
```diff
@@ -10,6 +10,10 @@
   <groupId>io.enmasse</groupId>
   <artifactId>address-model-lib</artifactId>
   <dependencies>
+    <dependency>
+      <groupId>io.enmasse</groupId>
+      <artifactId>common-api-model</artifactId>
```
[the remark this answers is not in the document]
ctron: That would mean merging in "user-model" as well, is that ok for you?

common-api-model/src/main/java/io/enmasse/common/api/model/AbstractList.java
```diff
@@ -0,0 +1,62 @@
...
+import java.util.List;
+
+import com.fasterxml.jackson.annotation.JsonIgnore;
+
+import io.fabric8.kubernetes.api.model.HasMetadata;
+import io.fabric8.kubernetes.api.model.KubernetesResource;
+import io.fabric8.kubernetes.api.model.KubernetesResourceList;
+import io.fabric8.kubernetes.api.model.ListMeta;
+
+public abstract class AbstractList<T extends HasMetadata>
```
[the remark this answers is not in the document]
ctron: Same here, absolutely.

ctron:
> I would prefer to just drop the VirtualBeanPropertyWriter and CustomResource stuff. Since we know upfront what resources we have up front, having a static initializer like in AdminCrd is less magic and easier to read IMO.

I know there is a bit of magic included. But right now the information is repeated and spread all over the place. A future change might easily break things. This way you only have one location where stuff is happening and one location where the information is.

If your concern is performance, then I think this can be improved to resorting to a bit of static information, like the `CRD` constant.

address-model-lib/pom.xml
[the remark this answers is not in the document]
lulf: Yes, thats fine, as long as keycloak-user-api is kept separate.

ctron:
I pushed a commit which removes the annotation processing, but still tries to keep all information and logic in one place. Adding a `UserCrd` in the process, just to show how this would work.

if this is ok, then I would start working on refactoring the other parts as well, creating a common `model` artifact.

address-model-lib/pom.xml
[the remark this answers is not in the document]
ctron: Ok, that part is done. I will move on the migrating the other API models.

lulf:
Excellent work @ctron , thanks! By 'core' do you mean Address and AddressSpace? They were written before fabric8 was used, so they are using the 'Builder' style and with explicit serializer/deserializer classes for the 'v1' version. IMO they should be converted to the same type of model-mixed-with-serialization as the other types, but take care to about the nitty gritty details on the Serializer/Deserializer classes which handles some backward compatibility cases.

I'm fine with leaving those as is for now if you want follow up that work in a second PR.

ctron:
Thank you! I saw there seems to be some glitch in the current PR, both CIs fail. I will fix that. And then I would say it would be a good point to merge this indeed.

I would still need the `ObjectMetadata` in the `AddressSpace` as well. So I will need to work on that. However I am not sure if I will manage to get this "merge ready" by the end of tomorrow. After that I will be out of office until early next year.

I would prefer to finish up this PR, and then get it merged. So I can start working on the rest, when I come back from vacation.

address-space-controller/src/main/java/io/enmasse/controller/AddressSpaceController.java
```diff
@@ -133,21 +133,21 @@ private void stop() {
...
         KubeResourceApplier.applyIfDifferent(new File(resourcesDir, "brokeredinfraconfigs"),
-                client.customResources(AdminCrd.brokeredinfraconfigs(), BrokeredInfraConfig.class, BrokeredInfraConfigList.class, DoneableBrokeredInfraConfig.class).inNamespace(namespace),
+                client.customResources(AdminCrd.brokeredInfraConfigs(), BrokeredInfraConfig.class, BrokeredInfraConfigList.class, DoneableBrokeredInfraConfig.class).inNamespace(namespace),
                 BrokeredInfraConfig.class,
                 Comparator.comparing(BrokeredInfraConfig::getVersion));
 
         KubeResourceApplier.applyIfDifferent(new File(resourcesDir, "standardinfraconfigs"),
-                client.customResources(AdminCrd.standardinfraconfigs(), StandardInfraConfig.class, StandardInfraConfigList.class, DoneableStandardInfraConfig.class).inNamespace(namespace),
+                client.customResources(AdminCrd.brokeredInfraConfigs(), StandardInfraConfig.class, StandardInfraConfigList.class, DoneableStandardInfraConfig.class).inNamespace(namespace),
                 StandardInfraConfig.class,
```
rgodfrey: why has this changed from standard to brokered?  was this intentional - it looks wrong

ctron: It is wrong! And fixed.

api-model/src/main/java/io/enmasse/address/model/AddressList.java
```diff
@@ -12,6 +12,9 @@
  * Type for address lists.
  */
 public class AddressList extends ArrayList<Address> {
+
+    private static final long serialVersionUID = 1L;
+
```
rgodfrey: Do we actually care about serialization (i.e. we are going to modify this if we change the class in a non-compatible way) or are we just trying to suppress warnings (in which case a @SuppressWarnings annotation might express the intent better)

api-model/src/main/java/io/enmasse/address/model/AddressSpaceList.java
```diff
@@ -12,6 +12,8 @@
  */
 public class AddressSpaceList extends ArrayList<AddressSpace> {
 
+    private static final long serialVersionUID = 1L;
+
```
rgodfrey: Same comment as previously about serialVersionUID

api-model/src/main/java/io/enmasse/address/model/UnresolvedAddressException.java
```diff
@@ -5,6 +5,9 @@
 package io.enmasse.address.model;
 
 public class UnresolvedAddressException extends RuntimeException {
+
+    private static final long serialVersionUID = 1L;
+
```
rgodfrey: As before - and I can't believe we ever want to serialise an exception object

api-model/src/main/java/io/enmasse/address/model/AddressList.java
ctron: Yes this all about suppressing warnings. The latter case would work fine as well. Whatever you prefer.

api-model/src/main/java/io/enmasse/address/model/UnresolvedAddressException.java
ctron: I pushed a commit which switches from adding the serial IDs to simply ignoring the warning.

So we can merge (or revert) this commit.

api-model/src/main/java/io/enmasse/address/model/AddressList.java
rgodfrey: Or you could just turn off the warnings in Eclipse :-)

I generally hate the serialVersionUID being included because it looks like an intention for the class to be serialised, and that an attempt will be made to either maintain compatibility or modify the Uid when necessary.  If the intent is to suppress warnings I'd probably use the appropriate annotation (yes there are warnings but I just don't care) or simply turn them off in your IDE)

api-model/src/main/java/io/enmasse/address/model/UnresolvedAddressException.java
rgodfrey: I'm fine with ignoring/suppressing the warning

api-model/src/test/java/io/enmasse/user/model/v1/UserModelTest.java
```diff
@@ -19,32 +22,35 @@
...
                                         .build()))
                         .build())
                 .build();
 
         ObjectMapper mapper = new ObjectMapper();
         byte [] serialized = mapper.writeValueAsBytes(user);
 
+        Object foo = mapper.readValue(serialized, User.class);
+        System.out.println(foo.getClass());
+
```
rgodfrey: Was this meant to be left here? I don't see foo being used again

ctron: Nope. Gone.

ctron:
I rebased after the merge of e069e4d41abe5504b4c4dfc5e0477ff55619a37f

ctron:
Rebased after the merge of #2113

Issue #2110, opened 2018-12-11
Addreses of MessageUser get encoded, when it should not


ctron:
**Describe the bug**

Adding an address like `event/hono.default` to the list of authorized addresses to a `MessagingUser` applies a URL encoding to the address:

     event%2Fhono.default

I would expect the value to remain:

    event/hono.default

lulf:
Fixed on master

PR #2113, opened 2018-12-12
[#2110]: Add missing URL decoding of group names

ctron:
This PR fixes the URL encoding of the addresses reported in #2110 and adds a bunch of unit tests to verify the results.


keycloak-user-api/src/main/java/io/enmasse/user/keycloak/KeycloakUserApi.java
```diff
@@ -244,6 +234,49 @@ private void applyAuthorizationRules(RealmResource realm, User user, UserResourc
...
+        final Set<String> desiredGroups = new HashSet<>();
+
+        for (UserAuthorization userAuthorization : authorization) {
+
+            for (Operation operation : userAuthorization.getOperations()) {
+
+                if (userAuthorization.getAddresses() == null || userAuthorization.getAddresses().isEmpty()) {
+
+                    switch (operation) {
+                    case manage:
```
kornys: could you please format this switch?

ctron: Sure, if you can point me to the formatter rules 😁

kornys: 😁 https://google.github.io/styleguide/javaguide.html#s4.8.4-switch

lulf: IntelliJ default format 😅👍

ctron: I am sorry, but I can only handle Eclipse formats :wink:

ctron: I found a ready-to-use Eclipse file for using the Google code style for Java. I will go for that then.

ctron: I don't think you are using the Google code style, applying it causes a whole bunch of changes.

lulf: I think the closest matching is oracle java conventions or something, if there is such a formal specification

ctron: I think I got it. Looks like a modifed version of Google's code formatter, with 4 spaces indentation instead of 2.

I pushed a change, please have a look.

kornys: LGTM!

PR #2114, opened 2018-12-12
Rename node affinity labels

lulf:
This matches the form used by Openshift node groups and makes it easier to have multiple roles on the same node during openshift install.


k-wall:
wq

PR #2121, opened 2018-12-13
Added authentication to prometheus, alertmanager

robshelly:
@lulf What do you think of this for the monitoring authentication? I'm not entirely happy with it myself but it's the only way I've gotten it working. The parts I'm not happy about are the need to run `htpasswd` during installing [1] and the security context constraints [2].

[1] [https://github.com/EnMasseProject/enmasse/compare/master...robshelly:monitoring-authentication?expand=1#diff-a501728a33cc0c36febb64a9e3076673R53](https://github.com/EnMasseProject/enmasse/compare/master...robshelly:monitoring-authentication?expand=1#diff-a501728a33cc0c36febb64a9e3076673R53)

[2] [https://github.com/EnMasseProject/enmasse/compare/master...robshelly:monitoring-authentication?expand=1#diff-a501728a33cc0c36febb64a9e3076673R65](https://github.com/EnMasseProject/enmasse/compare/master...robshelly:monitoring-authentication?expand=1#diff-a501728a33cc0c36febb64a9e3076673R65)

### DIFF
--- a/ansible/inventory/with-standard-authservice.txt
+++ b/ansible/inventory/with-standard-authservice.txt
@@ -10,3 +10,5 @@ register_api_server=True
 keycloak_admin_password=admin
 authentication_services=["standard"]
 enable_monitoring=False
+monitoring_username=admin
+monitoring_password=secret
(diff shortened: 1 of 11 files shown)


lulf:
@robshelly Yeah, I agree its not ideal to use basic auth :/ What I think would be ideal is something like #2111 but would work (in a 'disabled mode') on Kubernetes as well. OTOH, the ansible playbooks only support OpenShift, so not too much to worry about wrt. Kubernetes there.

Perhaps with the install operator, we will be able to configure things appropriately for both cases.

I'd vote for using the openshift oauth proxy for now, and figure out a better way to support monitoring for Kubernetes later. If we deployed the prometheus operator, would we be better off wrt. getting auth out of the box?

PR #2124, opened 2018-12-14
Attempt to ensure AMQP connection is always closed

rgodfrey:


rgodfrey:
@lulf well the 10s is *after* the plugin should have sent the data to the party trying to authenticate, so it *should* be ok, even if the 10s timeout gets invoked... but as you say, we will see :)

Issue #2127, opened 2018-12-16
Cannot send/receive messages in standard addressspace when one qdrouter pod is deleted


kornys:
**Describe the bug**
When one of two qdrouter pods is deleted and test immediately tries to send/receive message, client fails on timetout.

**To Reproduce**
1. Create standard address space
2. Create topic/queue addresses (standard-small-queue, standard-small-topic, standard-large-queue, standard-large-topic)
3. wait for addresses are ready
4. delete one of the qdrouter pod
5. try attach client to address immediately

**Expected behavior**
Client should connect immediately because there is still one qdrouter pod active

**Test log**

[testlog.zip](https://github.com/EnMasseProject/enmasse/files/2683670/testlog.zip)

```
...
2018-12-15T17:38:34.892Z INFO  [TestUtils] Checking {name=test-queue, address=test-queue} for ready state
2018-12-15T17:38:39.893Z INFO  [TimeMeasuringSystem] End time of operation CREATE_ADDRESS-dd4cefd0 is correctly stored
2018-12-15T17:38:39.893Z INFO  [GlobalLogCollector] Collecting configmaps for namespace myproject
2018-12-15T17:38:40.965Z INFO  [AmqpClientFactory] External endpoint: messaging-9vvx8n91wl-myproject.10.37.144.48.nip.io:443, internal: messaging-9vvx8n91wl-myproject.10.37.144.48.nip.io:443
2018-12-15T17:38:41.959Z INFO  [Receiver] Receiver link 'test-queue' opened, granting credits
2018-12-15T17:38:42.536Z INFO  [Sender] Sender link 'test-queue' opened, sending messages
2018-12-15T17:38:43.556Z INFO  [AmqpClientFactory] External endpoint: messaging-9vvx8n91wl-myproject.10.37.144.48.nip.io:443, internal: messaging-9vvx8n91wl-myproject.10.37.144.48.nip.io:443
2018-12-15T17:38:44.277Z INFO  [Receiver] Receiver link 'test-topic' opened, granting credits
2018-12-15T17:38:44.803Z INFO  [Sender] Sender link 'test-topic' opened, sending messages
2018-12-15T17:38:45.847Z INFO  [AmqpClientFactory] External endpoint: messaging-y0vy94skua-myproject.10.37.144.48.nip.io:443, internal: messaging-y0vy94skua-myproject.10.37.144.48.nip.io:443
[...]

PR #2129, opened 2018-12-16
Allow MessagingUsers to be created from service account

### LINKED ISSUE #2054: Allow MessagingUsers to be created from service accounts
## why

Allow MessagingUsers to be created from service accounts.  This will allow applications running in Kubernetes to use their service account credentials to authenticate against the address space

## Proposal

Update the Keycloak plugin to recognize a new type of credential for a service account - from a SASL perspective it will need to be sent using PLAIN.  The plugin will find the user, see it is a service account based user, and then attempt to authenticate with Kubernetes using the service account name / password.

## Testing

Test that a messaging user can be created from a service account
Test that the credentials of the service account can be used to make a messaging connection
Test that using the credentials of a service account which is not a messaging user does not allow a messaging connection to be made
Test that using an incorrect password for the service account does not allow a successful connection

## Documentation

Update docs on creating messaging users

rgodfrey:
This closes #2054

### DIFF
--- a/keycloak-plugin/sasl-plugin/pom.xml
+++ b/keycloak-plugin/sasl-plugin/pom.xml
@@ -30,6 +30,26 @@
     </dependency>
 
     <dependency>
+      <groupId>io.fabric8</groupId>
+      <artifactId>openshift-client</artifactId>
+      <scope>compile</scope>
+      <exclusions>
+        <exclusion>
+          <groupId>com.fasterxml.jackson.core</groupId>
+          <artifactId>jackson-core</artifactId>
+        </exclusion>
+        <exclusion>
+          <groupId>com.fasterxml.jackson.core</groupId>
+          <artifactId>jackson-databind</artifactId>
+        </exclusion>
+        <exclusion>
+          <groupId>com.fasterxml.jackson.core</groupId>
+          <artifactId>jackson-annotations</artifactId>
+        </exclusion>
+      </exclusions>
+    </dependency>
+    
+    <dependency>
       <groupId>org.keycloak</groupId>
       <artifactId>keycloak-server-spi</artifactId>
       <scope>provided</scope>
--- a/keycloak-plugin/sasl-plugin/src/main/java/io/enmasse/keycloak/spi/K8sServiceAccountCredentialProvider.java
+++ b/keycloak-plugin/sasl-plugin/src/main/java/io/enmasse/keycloak/spi/K8sServiceAccountCredentialProvider.java
@@ -28,6 +28,7 @@ import org.keycloak.models.cache.OnUserCache;
 
 import java.io.IOException;
 import java.io.UncheckedIOException;
+import java.util.Collections;
 import java.util.Objects;
 import java.util.Set;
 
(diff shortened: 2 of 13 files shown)


keycloak-plugin/sasl-plugin/src/main/java/io/enmasse/keycloak/spi/K8sServiceAccountCredentialProvider.java
```diff
@@ -0,0 +1,141 @@
...
+        } else if(!Objects.equals(userModel.getUsername(), userName)) {
+            LOG.debug("Attempt to log in for user " + userModel.getUsername() + " in realm " + realmModel.getName() + " with token for " + userName);
+        }
+
+        return authenticated && Objects.equals(userModel.getUsername(), userName);
+    }
+
+    private JsonObject doRawHttpRequest(String path, String method, JsonObject body, boolean errorOk) {
+        NamespacedOpenShiftClient client = new DefaultOpenShiftClient();
+        OkHttpClient httpClient = client.adapt(OkHttpClient.class);
```
lulf: (If we want to use the kubernetes-client etc.) I think you need to close both of these to avoid a resource leak

keycloak-plugin/sasl-plugin/pom.xml
```diff
@@ -29,6 +29,26 @@
       </exclusions>
     </dependency>
 
+    <dependency>
+      <groupId>io.fabric8</groupId>
+      <artifactId>openshift-client</artifactId>
```
lulf: I think it might be worth using just okhttp client or some other http client already part of keycloak to perform the requests. The token can be dug up from `/var/run/secrets/kubernetes.io/serviceaccount/token`

templates/standard-authservice/010-ServiceAccount-std-authsvc.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: std-authsvc
```
lulf: To make naming consistent with the service for instance: `standard-authservice`

keycloak-plugin/sasl-plugin/src/main/java/io/enmasse/keycloak/spi/K8sServiceAccountCredentialProvider.java
rgodfrey: I literally just copied this code from api-server :-)

keycloak-plugin/sasl-plugin/pom.xml
rgodfrey: Is it just a matter of adding some headers to the HTTP request with the token for authorisation?

lulf: Yes, this is done in the doRawHttpRequest method:

             .addHeader("Authorization", "Bearer " + client.getConfiguration().getOauthToken())

keycloak-plugin/sasl-plugin/src/main/java/io/enmasse/keycloak/spi/K8sServiceAccountCredentialProvider.java
lulf: The new DefaultOpenshiftClient() is the one leaking. In the api-server it is created at the start of the server process.

rgodfrey: I've moved the creation of the DefaultOpenShiftClient to the constructor and closed it in the close method of the provider.

keycloak-plugin/sasl-plugin/pom.xml
rgodfrey: Given we include it elsewhere, I'm minded to keep as is

kornys:
@lulf I fixed what you requested and also disabled tests until #2127 will be fixed, I think it can be merged.

PR #2145, opened 2018-12-20
Allow PLAIN serviceaccount auth without username

rgodfrey:
This closes #2144

PR #2147, opened 2018-12-20
New tests for federated/serviceaccount messagingusers

kornys:
new tests for https://github.com/EnMasseProject/enmasse/pull/2145 and federated users


kornys:
@lulf could you please re-approve it? tests in CI passed, only one mqtt test failed but this changes don't affect mqtt so failure is unrelated.

PR #2164, opened 2019-01-07
Added alert for pods stuck in pending phase

robshelly:


templates/prometheus/010-ConfigMap-prometheus-config.yaml
```diff
@@ -122,10 +122,18 @@ data:
...
         expr: rate(kube_pod_container_status_restarts_total[10m]) * 600 > 5
         for: 60s
+      - alert: Pending_Pods
+        annotations:
+          description: '{{ $labels.pod }}'
+          phase: '{{ $labels.phase }}'
+          value: "{{ $value }}"
+          severity: warning
+        expr: kube_pod_status_phase{phase="Pending"} == 1
+        for: 60s
```
lulf: Suggest bumping these alerts to 180 seconds or something. At least for pending, it can be due to long container image pull time

robshelly: Good point. Done!

PR #2165, opened 2019-01-07
Improve address space status check

lulf:
* Check statefulsets in addition to deployments
* Less strict: require >= 1 ready pod per deployment/statefulset to pass status check.
* Only check components belonging to address space

### DIFF
--- a/address-space-controller/src/main/java/io/enmasse/controller/StatusController.java
+++ b/address-space-controller/src/main/java/io/enmasse/controller/StatusController.java
@@ -66,7 +66,7 @@ public class StatusController implements Controller {
             checkStatefulSetsReady(addressSpace, requiredResources);
         } catch (Exception e) {
             String msg = String.format("Error checking for ready components: %s", e.getMessage());
-            log.warn(msg);
+            log.warn(msg, e);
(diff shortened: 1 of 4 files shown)


address-space-controller/src/main/java/io/enmasse/controller/StatusController.java
```diff
@@ -55,33 +57,61 @@ private InfraConfig parseCurrentInfraConfig(AddressSpace addressSpace) throws IO
...
-                addressSpace.getStatus().appendMessage("Following deployments and statefulsets are not ready: " + missing);
-            }
+            List<HasMetadata> requiredResources = infraResourceFactory.createInfraResources(addressSpace, infraConfig);
+
+            checkDeploymentsReady(addressSpace, requiredResources);
+            checkStatefulSetsReady(addressSpace, requiredResources);
         } catch (Exception e) {
-            String msg = String.format("Error checking for ready deployments: %s", e.getMessage());
+            String msg = String.format("Error checking for ready components: %s", e.getMessage());
             log.warn(msg);
```
k-wall: We might need for the context for diagnostics purposes.  Suggest passing `e` to the `log.warn()`.

PR #2098, opened 2018-12-10
Address plan switching for queues

lulf:
Tasks:
- [x] Create broker status field in address 
- [x] Add annotation to determine currently active plan 
- [x] Add all brokers to address brokerstatus and set brokers to draining state if queue should be moved
- [x] Cut of autolink for incoming traffic in agent
- [x] Delete drained brokers with 0 message count and unacked messages (requires checking broker status, standard-controller needs to query broker through agent)
- [x] Systemtests changing address plans and verifying that messages are not lost and brokers are drained/deleted

PR #2168, opened 2019-01-08
Replace address space tests

kornys:
Base tests for verify if addressspace is implemented


systemtests/src/main/java/io/enmasse/systemtest/TestUtils.java
```diff
@@ -391,15 +391,15 @@ public static String getAddressSpaceType(JsonObject addressSpace) {
...
         JsonObject addressSpaceObject = null;
         TimeoutBudget budget = null;
 
         boolean isReady = false;
         budget = new TimeoutBudget(5, TimeUnit.MINUTES);
         while (budget.timeLeft() >= 0 && !isReady) {
-            addressSpaceObject = apiClient.getAddressSpace(addressSpace);
-            isReady = isAddressSpaceReady(addressSpaceObject);
+            addressSpaceObject = apiClient.getAddressSpace(addressSpace.getName());
+            isReady = isAddressSpaceReady(addressSpaceObject) && matchAddressSpacePlan(addressSpaceObject, addressSpace);
```
lulf: For tests that want to test that a plan cannot be applied (due to too little quota), I think it might make sense to split this into a waitForAddressSpaceReady and waitForAddressSpacePlanApplied or something like that, and invoke them separately like https://github.com/EnMasseProject/enmasse/pull/2098/files#diff-3d1b3eed8fb655a202d75bab060610d0R304

If you write a test that wants to verify that plan is not applied, splitting them up would make it possible to provide more context about why the test failed.

PR #2170, opened 2019-01-08
Ensure that JVM options are passed to the Broker's JVM.

### LINKED ISSUE #2158: Broker pod continues to run even after OutOfMemoryError
During a performance test that uses a standard addressspace using default-minimal infraconfig, if I publish messages to a queue (standard-small-queue plan) continually, the Broker goes OOME (heap) but continues to run on, seeming indefinitely, producing spurious exceptions.

This raises two questions:

1) The Broker JVM is configured with JVM -XX:+ExitOnOutOfMemoryError so why is the JVM not shutting down which would cause the openshift to restart the pod.
2) If the default size of Broker defined by the infraconfig too low?

I worked around by increasing broker.resources.memory to 2Gi.

k-wall:
Fixes #2158.


broker-plugin/plugin/src/main/sh/launch-broker.sh
```diff
@@ -2,6 +2,7 @@
 
 BROKER_DIR=${ARTEMIS_HOME}
 BROKER_CUSTOM_DIR=${BROKER_DIR}/custom
+# Deletes the last line so it wont start the broker? Necessary obscurity?
 sed -i '$ d' ${BROKER_DIR}/bin/launch.sh
```
k-wall: I'd prefer to remove the line from launch.sh rather than have the script delete it.  Is there some reason it needs to be this way?

broker-plugin/plugin/src/main/sh/init-broker.sh
```diff
@@ -16,6 +16,8 @@ cp -r ${BROKER_PLUGIN_DIR}/jmx_exporter/jmx_prometheus_javaagent-0.1.0.jar $BROK
 cp -r ${BROKER_PLUGIN_DIR}/lib $BROKER_CUSTOM
 cp -r ${BROKER_PLUGIN_DIR}/bin $BROKER_CUSTOM
 
+source /opt/run-java/dynamic_resources.sh > /dev/null
```
vbusch: Does the init-container have the same resources available as the actual container?  If so, it would be fine to move this into the init container.  The downstream broker/launch.sh will also run this.

k-wall: @vbusch you are right.   My approach was flaw as I was generating the JVM arguments from within the wrong container.  If the base image diverged the results would have been wrong and as you pointed out, settings based on resources could have been wrong.

 I took an alternative approach instead.

PR #2173, opened 2019-01-09
Docs: add Console assembly file

jenmalloy:


jenmalloy:
@lulf, yes, that is part of my larger plan. I'd like to restructure the assembly that content is in, so will do that in another PR to follow today. Thanks, Ulf!

PR #2175, opened 2019-01-09
Update API versions to v1beta1

lulf:
This updates the API version of all EnMasse resources to use v1beta1. The v1alpha1 version will still be supported for all resources, but it will default to v1beta1.


lulf:
@rgodfrey Added unit tests (and fixed bugs :))

Issue #2179, opened 2019-01-09
creating address with restricted chars is possible through api server ( address creation gets stuck on configuring ) 


RH-sdavey:
**Describe the bug**
Its possible to create an address with restricted chars ( # ! . / : and whitespace ) through api server ( oc create -f standard-small-queue.yaml), this was blocked in console as it was causing address creation to be stuck on configuring, the address never makes it to 'Active' state. 

**To Reproduce**
1. create queue address definition [1], standard-small-queue.yaml
2. oc create -f standard-small-queue.yaml
3. address.enmasse.io/seantest-addr.testqueue1 created
4. console address loading icon gets stuck
5. [2]

**Expected behavior**
Api server would reject invalid address names, and Error from server output would be shown. 

**Additional context**
[1]
> apiVersion: enmasse.io/v1alpha1
kind: Address
metadata:
    name: seantest-addr.testqueue1     <=== valid name
  spec:
    address: seantest-addr.te/st:qoe!!!uei##\#,,.   <=== invalid name
     type: queue
    plan: standard-small-queue
~                                   

[2]
>NAME                       ADDRESS                                  ADDRESSSPACE    TYPE      PLAN                   READY     PHASE         AGE       STATUS
seantest-addr.testqueue1   seantest-addr.t e/st:qoe!!!uei###,,.   seantest-addr   queue     standard-small-queue   false     Configuring   2m        Address seantest-addr.t e/st:qoe!!!uei###,,. is missing autoLinks on qdrouterd-py85uq450e-1. Address seantest-addr.t e/st:qoe!!!uei###,,. is missing autoLinks on qdrouterd-py85uq450e-0. Address seantest-addr.t e/st:qoe!!!uei###,,. is missing active autoLink (active in dirs: []). Address seantest-addr.t e/st:qoe!!!uei###,,. not found on qdrouterd-py85uq450e-0. Address seantest-addr.t e/st:qoe!!!uei###,,. not found on qdrouterd-py85uq450e-1

PR #2180, opened 2019-01-09
Add address validation

lulf:


api-model/src/main/java/io/enmasse/address/model/Address.java
```diff
@@ -130,16 +132,24 @@ public void validate() {
     private void validateName(String name) {
         String[] components = name.split("\\.");
         if (components.length < 2) {
-            throw new IllegalArgumentException("Address name must be on the form addressSpace.addressName");
+            throw new AddressValidationFailedException("Address name must be on the form addressSpace.addressName");
```
RH-sdavey: must be in the form

lulf:
Is it really not allowed to create addresses with '/' ? @grs  I think this is already used by Bosch (CC @calohmn ).

grs:
@lulf you can relax it if it is too strict. I would check how artemis behaves with a queue that has a / in it though, may be fine. (For anycast it certainly would not cause any problem)

k-wall:
What is it that motivates us to make this change?

Do we have a definition somewhere  of what constitutes a valid address in EnMasse?  I think this should appear in user-documentation too. 

I tried creating addresses and queue in Artemis containing the characters excluded by the PR.  It didn't seem to cause an issue.  I couldn't find any documentation.

lulf:
@k-wall It was motivated by the referenced issue #2179 where the console is restricting the allowed characters for addresses. I think we should perhaps leave it out and do more experimentation before we end up restricting it more than necessary.

RH-sdavey:
+1 for proper documentation of what is a valid/invalid address if possible, 
also, I noticed there were differences between what addresses were possible to be made through API server and through console, where they would be accepted in one and not the other, so maybe some standardization on that?

lulf:
Have to be rethought. Also underlying model changes in #2202

PR #2181, opened 2019-01-09
Implement non-namespaced versions for listing users and addresses (#2101)

ctron:


keycloak-user-api/src/main/java/io/enmasse/user/keycloak/KeycloakUserApi.java
```diff
@@ -443,26 +433,88 @@ public boolean realmExists(String realmName) {
         return withKeycloak(kc -> getRealmResource(kc, realmName) != null);
     }
 
-    @Override
-    public UserList listUsers(String namespace) {
+    protected UserList queryUsers(final Predicate<RealmRepresentation> realmPredicate, final Predicate<UserRepresentation> userPredicate) {
```
lulf: protected -> private?

PR #2188, opened 2019-01-10
Support moving between sharded plans

lulf:


k-wall:
Changes look good, however as discussed I notice two issues:

- When scaling up an empty queue from address plan small to medium, I see the the old broker disappear before the new broker is ready.  It would be a better behaviour for connected clients if the provisioner awaited the new broker to be ready before decommissioning the first.
- When scaling up up an empty queue from address plan small to large in a standard-small address space, I expect the address to be over quota and this to be reported when I check the queue's status with oc get address  Looking at the logs, I see this is the case internally.

```
Oslo:enmasse keith$ oc get address -o wide
NAME                ADDRESS    ADDRESSSPACE   TYPE      PLAN                    READY     PHASE     AGE       STATUS
standard.myqueue1   myqueue1   standard       queue     standard-xlarge-queue   true      Active    1h

```

lulf:
> Changes look good, however as discussed I notice two issues:
> 
>     * When scaling up an empty queue from address plan small to medium, I see the the old broker disappear before the new broker is ready.  It would be a better behaviour for connected clients if the provisioner awaited the new broker to be ready before decommissioning the first.
> 

I have fixed this by ensuring that the address status is reset to its original state if the quota check fails. Originally, it would accept the modified list of brokerstatus even if the quota check failed, which in turn got picked up in the next controller loop iteration calculating the usage. The new broker then got provisioned because the usage indicated that it were already using the new sharded quota. Resetting the state fixes this.

I have also changed so that status is reset at the beginning of the controller loop (after the initial copy used for comparing if status changed). Part 

>     * When scaling up up an empty queue from address plan small to large in a standard-small address space, I expect the address to be over quota and this to be reported when I check the queue's status with oc get address  Looking at the logs, I see this is the case internally.
> 

I have fixed this by introducing a 'Migrating' step, which holds the broker to be removed alive until all Active brokers have at least 1 ready replica.

standard-controller/src/main/java/io/enmasse/controller/standard/AddressController.java
```diff
@@ -100,6 +100,10 @@ public void onUpdate(List<Address> addressList) throws Exception {
 
         AddressProvisioner provisioner = new AddressProvisioner(addressSpaceResolver, addressResolver, addressSpacePlan, clusterGenerator, kubernetes, eventLogger, options.getInfraUuid(), brokerIdGenerator);
 
+        for (Address address : addressList) {
+            address.getStatus().setReady(true).clearMessages();
+        }
+
         Map<Status.Phase, Long> countByPhase = countPhases(addressSet);
         log.info("Total: {}, Active: {}, Configuring: {}, Pending: {}, Terminating: {}, Failed: {}", addressSet.size(), countByPhase.get(Active), countByPhase.get(Configuring), countByPhase.get(Pending), countByPhase.get(Terminating), countByPhase.get(Failed));
```
k-wall: Should Migrating addresses by counted and logged too?

PR #2220, opened 2019-01-17
Fix up a few things in the global log collector

ctron:
* Print out full path were files are being created
* Drop back and forth conversion with File and directly use Path
* pull all "create log file with parent" logic into a single method


kornys:
@ctron could you please rebase this branch on current master?

ctron:
Rebased.

PR #2228, opened 2019-01-17
Add back reseasy multipart dependency

kornys:


lulf:
@k-wall FYI this was added back because it broke keycloak controller in some cases looking for classes in this dependency

PR #2229, opened 2019-01-17
Fix HTTP server test, didn't properly handle multiple steps

ctron:
This change handle multiple steps in each test by using the checkpoint
feature, which will track progress during the test and only complete
once all required checkpoints have been completed.

Before that the first "complete" call would have finished the test.
Even if that was only the first step. Further steps were ignored and
failures not found.

This change also fixes issues brought up by actually testing.

Also, the multiple methods for the different API versions were
simplified using parameterized tests.


lulf:
Somehow I think this is already in @ctron. Closing this.

PR #2249, opened 2019-01-23
Switch to Minikube v0.33.1

ctron:
This updates Minikube to the most recent version. I want to test what happens on the Travis build.

If we want, we can still merge this later, but this is not my primary goal here.


lulf:
@ctron What is the status of this? Close or continue effort?

lulf:
Closing forn ow

PR #2266, opened 2019-01-28
Exclude bouncycastle as dependency of keycloak-admin-client (use JCE instead)

k-wall:


k-wall:
This change will mean keycloak-controller uses the standard JCE rather than BC.   It will also mean that all the other applications that included keycloak-user-api will no longer include BC on their class paths. This dependency would have been unused because the provider was never installed. These were: address-space-controller, api-server, service-broker.

PR #2269, opened 2019-01-29
Invoke thread that is GCed when completed

lulf:


lulf:
@kornys please have another look

PR #2274, opened 2019-01-29
Docs: add variable

jenmalloy:


documentation/common/proc-download-procedure.adoc
```diff
@@ -14,7 +14,7 @@ ifdef::Downloading[]
 endif::Downloading[]
 
 ifndef::Downloading[]
-* Download and extract the `install_and_examples.zip` file from the {ZipDownload}.
+* Download and extract the `{ZipInstallFile}` file from the {ZipDownload}.
```
ncbaratta: I don't see an edit to the attributes file including this attribute.

lulf: I don't believe this section will be displayed in the upstream as we have a different way to download.

PR #2276, opened 2019-01-29
Run upgrade pipeline from 0.26.0

kornys:


lulf:
@kornys needs to be rebased

kornys:
@lulf done

PR #2281, opened 2019-01-30
Use podman

lulf:


ctron:
I think you need to add the following to `.travis.yml`:

~~~yaml
dist: xenial
~~~

As the openjdk PR is merged, this should work now.

kornys:
@lulf I think you need to add `sudo add-apt-repository -y ppa:projectatomic/ppa`

kornys:
@lulf I do not understand why in travis ubuntu /etc/containers/registries.conf is not created by default and on rhel/fedora/centos is created.

lulf:
@kornys because it’s not a standard, and only something that is part of fedora/cent is/rhel

lulf:
Close until we have some spare time to look more into podman

PR #2293, opened 2019-01-31
Export CA cert in AddressSpaceStatus

ctron:
This exports the CA cert of the address space in the status section.

I based this PR on PR #2202, as it already builds on top of the refactored model API.

### DIFF
--- a/address-space-controller/src/main/java/io/enmasse/controller/EndpointController.java
+++ b/address-space-controller/src/main/java/io/enmasse/controller/EndpointController.java
@@ -7,6 +7,7 @@ package io.enmasse.controller;
 import io.enmasse.address.model.*;
 import io.enmasse.config.AnnotationKeys;
 import io.enmasse.config.LabelKeys;
+import io.enmasse.controller.auth.OpenSSLCertManager;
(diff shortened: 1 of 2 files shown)


address-space-controller/src/main/java/io/enmasse/controller/auth/CertManager.java
```diff
@@ -16,6 +18,9 @@
     Collection<CertComponent> listComponents(String uuid);
     boolean certExists(CertComponent component);
     Secret getCertSecret(String name);
+    default Secret getCertSecret(AddressSpace addressSpace) {
+        return getCertSecret(KubeUtil.getAddressSpaceExternalCaSecretName(addressSpace));
```
lulf: I think this breaks the level of abstraction in CertManager.

I would prefer to not have this default method here and instead use KubeUtil.getAddressSpaceExternalCaSecretName(addressSpace) in EndpointController. 

Also because the address space CA is specific to the SelfSignedCertProvider.

ctron: Done and rebased.

PR #2295, opened 2019-01-31
Updated the infra-config version to be the release-version

vbusch:
- Maven reads the release-version and replaces the value in the property files from release.version.

- Doesn't duplicate the release-version.  However, needed a different maven-plugin to read the release version from the release.version file.

Would it be preferred to re-write this having the release-version in maven.  And using a maven ant-task to write the release.version file?  The standard resource plugin will write it to the build-dir, which makes it use in the Makefile scripts difficult.


lulf:
If there was a way to easily 'extract' the version from pom.properties or something to be used in the Makefile, I think it would be good to move it there.

vbusch:
moved release.version to pom.properties.  For consistency, also moved maven.version.

PR #2298, opened 2019-02-01
Fix for upgradeTests for version 0.26

vbusch:
Version 0.26 has been released with controller version 0.26, not matching infra config version 0.26.0.  

This fix will allow the upgrade to happen as long as the major & minor versions match.
It is only required until after version 0.27 is released.  PR #2295 ensures version numbers match in future releases.


lulf:
@vbush I have tagged, 0.26.2 with the fix to infra configs, so this should not be needed

lulf:
@vbusch Closing since we have working upgrades now.

PR #2301, opened 2019-02-01
Fix the PlanTest failing due to missing endpoint

ctron:
This change waits after the update of the address space that the
resource version changes, only after that does it make sense to wait
to wait for the readiness of the address space.

What happened was that the update came, and immediately it was checked
that the address space is ready, which succeeded, but for the old state.
Then the test progressed, but actually ran into an "unready" address
space.

### DIFF
--- a/systemtests/src/main/java/io/enmasse/systemtest/TestUtils.java
+++ b/systemtests/src/main/java/io/enmasse/systemtest/TestUtils.java
@@ -20,6 +20,7 @@ import io.vertx.core.http.HttpMethod;
 import io.vertx.core.json.JsonArray;
 import io.vertx.core.json.JsonObject;
 import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.function.ThrowingSupplier;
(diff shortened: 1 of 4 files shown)


ctron:
So this is an additional check, it only checks that the resource version has changed, so that following checks (ready, ...) will not start checking in a stale version.

Adding a sleep(1_000) already did the trick, but that is more guessing. I also think that one minute should be fine to check for the initial update of the record. Again, after that change was detected, the normal checks of ready, etc will still be performed.

lulf:
The controller loop runs every 30 seconds in worst case. Lets say that controller loop just started before the address space was updated, which means the update was not in the current update, and for some reason Kubernetes watch does not fire (which is not a guarantee in a network environment).

Then, if for some reason the system in the systemtests are loaded and the controller loop is slow + getting watches is slow, it is starting to close in on 1 minute. Lets be conservative about the timeouts, it doesnt hurt anything other than optimizing feedback on test failure (which is not the common case).

ctron:
Ok, you convinced me 😁

PR #2315, opened 2019-02-05
Remove group wildcards special to brokered address space

lulf:
The broker sasl plugin has been updated to handle the same wildcard
syntax as the dispatch router.

### DIFF
--- a/broker-plugin/sasl-delegation/src/main/java/io/enmasse/artemis/sasl_delegation/SaslDelegatingLogin.java
+++ b/broker-plugin/sasl-delegation/src/main/java/io/enmasse/artemis/sasl_delegation/SaslDelegatingLogin.java
@@ -32,6 +32,7 @@ import java.security.NoSuchAlgorithmException;
 import java.security.Principal;
 import java.util.*;
 import java.util.function.BooleanSupplier;
+import java.util.stream.Collectors;
 import javax.naming.InvalidNameException;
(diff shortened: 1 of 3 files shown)


ctron:
Can you use a LinkedHashSet instead of the HashSet, that should keep the order, at least in that piece of code.

lulf:
rerun tests

lulf:
rerun tests

lulf:
@enmasse-ci run tests

lulf:
@enmasse-ci run tests testcase=brokered.**

lulf:
@rgodfrey Could you check that last commit I made? It seems that Artemis expects that manage_# must exist. I figure its better to put that logic into the plugin so that the auth service API doesn't have to deal with it.

lulf:
@enmasse-ci run tests

rgodfrey:
@lulf I don't really understand your last commit, is it not adding _# to every group?  We wouldn't want that for the non manage_ groups do we?

lulf:
@rgodfrey Just above it does a retainAll on (SPECIAL_AUTHZ_GROUPS) which include 'admin' and 'manage' (so actual will only be manage). What would be the appropriate place to put this you think? I don't understand the code :)

rgodfrey:
@lulf yeah - having spent another 45 seconds reading the code, I think it is fine :)

PR #2316, opened 2019-02-05
Issue 2288 restructure templates

lulf:


lulf:
@robshelly Could you have another look at the recent changes? I made it possible to deploy auth services after the fact and added some missing roles

PR #2317, opened 2019-02-06
Move selenium to kubernetes app

kornys:


kornys:
@lulf it seems to me stable instead of deploying selenium in docker manually :)

PR #2325, opened 2019-02-08
use generic command executor for minikube commands

famartinrh:
We have 2 commands for minikube so there is no need to create CmdClient for minikube so I just replaced the usage of Process and ProcessBuilder with our Executor for commands.

### DIFF
--- a/systemtests/src/main/java/io/enmasse/systemtest/executor/Executor.java
+++ b/systemtests/src/main/java/io/enmasse/systemtest/executor/Executor.java
@@ -29,11 +29,11 @@ public class Executor {
     private boolean appendLineSeparator;
 
     public Executor() {
-        appendLineSeparator = true;
+        this.appendLineSeparator = true;
     }
 
     public Executor(Path logPath) {
-        appendLineSeparator = true;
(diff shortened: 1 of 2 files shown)


systemtests/src/main/java/io/enmasse/systemtest/executor/Executor.java
```diff
@@ -26,14 +26,21 @@
     private StreamGobbler stdOutReader;
     private StreamGobbler stdErrReader;
     private Path logPath;
+    private boolean appendLineSeparator;
 
     public Executor() {
+        this(true);
```
kornys: I don't like calling another constructor, better from my POV is setting appendLineSeparator manually to true

systemtests/src/main/java/io/enmasse/systemtest/executor/Executor.java
```diff
@@ -26,14 +26,21 @@
...
     }
 
     public Executor(Path logPath) {
+        this();
```
kornys: same as above

famartinrh:
changed

PR #2335, opened 2019-02-15
Initial support for Eclipse Hono

ctron:
This PR adds initial support of IoT use cases, using Eclipse Hono.


.travis.yml
```diff
@@ -1,58 +1,76 @@
...
+  - sudo add-apt-repository -y ppa:masterminds/glide && sudo apt-get update
+  - sudo apt-get install -y glide
+  - gem install asciidoctor
+  - gimme 1.10.5
+  - source ~/.gimme/envs/go1.10.5.env
+  - go version
 install: true # would only call "mvn install", which doesn't help us at this point
+before_script:
+  - pwd
+  - export GOPATH="$HOME/gopath"
```
kornys: @ctron why we need these steps? also I don't understand why we create this path and them copy all enmasse conetent here?

ctron: Welcome to the wonderful world of Go. Go has some specific requirements of its base path. And in this case I wasn't able to fix this with a symlink.

The "verify" step is needed in general to check if the code generator creates different files than the one that are currently checked in. If that is the case, then someone should manually run the code generator, and commit the changes.

iot/pom.xml
```diff
@@ -0,0 +1,85 @@
...
+        <groupId>org.springframework</groupId>
+        <artifactId>spring-web</artifactId>
+        <version>${spring.version}</version>
+      </dependency>
+      <dependency>
+        <groupId>org.springframework</groupId>
+        <artifactId>spring-webmvc</artifactId>
+        <version>${spring.version}</version>
+      </dependency>
+
```
lulf: Can we please get rid of all of the above dependencies? No spring boot or micrometer please :) We've managed without it so far, and we just got rid of spring boot from mqtt-gateway and mqtt-lwt.

.travis.yml
```diff
@@ -1,58 +1,76 @@
...
+  - "2.7.14"
+go:
+  - "1.10.x"
 before_install:
-- sudo pip install requests[security] ansible
-- gem install asciidoctor
+  - sudo pip install requests[security] ansible
+  - gem install asciidoctor
+  - sudo pip install requests[security] ansible
+  - sudo add-apt-repository -y ppa:masterminds/glide && sudo apt-get update
```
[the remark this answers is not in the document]
ctron: Go modules doesn't work with the kubernetes apimachinery stuff from what I know ... besides, support for go modules is, complicated at best. I think you will need to wait a few more years until this gets stable enough.

iot/pom.xml
ctron: Hono uses Spring Boot and Micrometer extensively. So I don't see much benefit in dropping and rewriting all the Hono base classes that we currently use. If we use Hono, then Spring Boot and Micrometer are a part of that.

.travis.yml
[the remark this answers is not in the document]
lulf: Operator framework uses go modules...: https://github.com/operator-framework/operator-lifecycle-manager/blob/master/go.mod

iot/pom.xml
lulf: As discussed, we'll have to live with it ... for now :)

.travis.yml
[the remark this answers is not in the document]
ctron: I am not sure Kubernetes itself is a good example: https://github.com/kubernetes/kubernetes/issues/63607 … which seems to have an impact on the code generator, which again impacts all go projects which provide custom resource definitions.

I guess the correct way is to use go modules at some point. Everything else is a nightmare, and glide simply turned out to cause less pain than the other solutions.

But sure, if we have something which is more hipster than glide, and it works, then let's replace it.

PR #2343, opened 2019-02-18
Fix running tests on different server

kornys:


kornys:
ok to test

kornys:
ok to test

PR #2350, opened 2019-02-20
Allow specifying globalMaxSize for broker config

lulf:
@enmasse-ci run tests


lulf:
@enmasse-ci run tests

k-wall:
One niggle. globalMaxSize which expresses a quantity of memory is expressed in different from that other resources (memory and storage) which use the IEC binary suffix.  As a user I would probably make a mistake.  Perhaps the infraconfigs should present a consistent facade?

PR #2351, opened 2019-02-20
Support setting handshake timeout

lulf:
@enmasse-ci run tests


lulf:
@enmasse-ci run tests

PR #2356, opened 2019-02-20
Fix compiler issue when running Maven with Java 11

ctron:
Currently the S2I Java build is broken as the Maven compiler plugin 3.1 is being used (current is 3.8). And 3.1 fails building on Java 11, and Java 11 is in the fabric8 s2i image.


lulf:

PR #2357, opened 2019-02-20
Split the travis build and run a second build, validating S2I

ctron:
Split up the travis build into two jobs, running in parallel.

The first job is the existing configuration, and the second job runs the S2I builder to ensure "buildability" on the sources.

The job split is done in order to reduce the chance to run into the 50 minute job limit of travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,34 +23,52 @@ python:
   - "2.7.14"
 go:
   - "1.10.x"
-before_install:
-  - sudo pip install requests[security] ansible
-  - gem install asciidoctor
-  - sudo pip install requests[security] ansible
-  - sudo add-apt-repository -y ppa:masterminds/glide && sudo apt-get update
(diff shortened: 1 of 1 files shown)


ctron:
This should successfully compile now that #2356 was merged

ctron:
This is currently blocked by sundrio/sundrio#129 …

ctron:
Raised PR: sundrio/sundrio#131

ctron:
Well, it is not testing this, but checking "buildability", current the issue is, that the build with S2I later on fails, due to the sundrio issue. Adding this to the travis build ensure that no code gets merged which later on breaks the build.

And it doesn't impact the main build, as this is a second build job and thus doesn't count towards the 50 minute build timeout.

lulf:
Wdyt @kornys

kornys:
LGTM

PR #2361, opened 2019-02-22
Selectively update generated resources, instead of overwriting them

ctron:
This change switches a few more locations to selectively update
the created resources, by manipulating the required fields.

That leaves other fields untouched and prevent unnecessary updates.


lulf:
@enmasse-ci run tests

PR #2362, opened 2019-02-22
Split up makefile and have dedicated Go builds

ctron:


ctron:

kornys:

ctron:
I had to re-run the Jenkins job, as it had some network error, unrelated to this change.

The build looks fine now, and even the Go unit tests are part of the Jenkins result now.

ctron:
Ok, I will change that then.

ctron:
Ok, I pushed a change which makes the following change:

You can use the Makefile variable `GO2XUNIT` to control if you want to have converted reports or not, if the variable is set to `` (empty), then this will be ignored. If it is set, then the binary must exists at this location (or in the path) and the conversion will be performed.

If the `GO2XUNIT` variable is set, then `go2xunit` will fail the build, otherwise `go test` will fail the build. The default is to have `go2xunit` in the path.

lulf:
Please update HACKING.md with required dependencies to build.

ctron:
@lulf  HACKING.md is updated.

lulf:
@ctron 

Having tried this locally, I think the whole cd $(GOPROJ) and overriding $(GOPATH) was a mistake. At least it breaks on my system where i have GOPATH set to $(HOME)/dev/go. I've tried to fix this in #2379 .

Also, I would like to make go2xunit an optional dependency. Is there a good reason to make it opt-out instead of opt-in? As a developer building and testing go, i don't really care about the xunit results.

ctron:
> Also the whole reason for this change in the first place is not clear to me.

The whole reason for this was to enable testing in Go, have test results in Jenkins and allow you to only build a sub-directory containing a go binary, without the need to build the full EnMasse project.

I added go2xunit as a requirement, since I think it is better to explicitly opt-out of something that is pretty useful, compared to asking why it wasn't running for the last few months and no one noticed.

lulf:
> The whole reason for this was to enable testing in Go, have test results in Jenkins and allow you to only build a sub-directory containing a go binary, without the need to build the full EnMasse project.

You should be able to do that now running `make -C iot/iot-gc`. 



> I added go2xunit as a requirement, since I think it is better to explicitly opt-out of something that is pretty useful, compared to asking why it wasn't running for the last few months and no one noticed

Understood. I think we had the same discussion regarding mocha-junit-reporter in the past. We decided to make it opt-in to improve the local developer experience, and I think we should try to require only build dependencies that will uncover a bug in the software itself. Failing to generate the xunit report is less critical IMO, and will be discovered eventually if a unit test fails without a test report in Jenkins.

PR #2363, opened 2019-02-22
Create plain k8s deployments for some IoT components

ctron:


lulf:
Just a thought on the s2i stuff in general. With recent changes on master, I've tried to remove the concept of bundles that all duplicate the contents in the 'enmasse-with..' etc. How about that instead of having the following bundles:

* enmasse
* monitoring
* iot
* iot_s2i

We have
* enmasse
* monitoring
* iot
* s2i

Where s2i contains all the OpenShift-specific s2i stuff, for all of enmasse? This way we can add s2i for other enmasse components that will be put there as well. Then, if you want to use s2i, it will be an optional step to apply before applying the other components.

Could this work you think?

ctron:
Hm, I like the idea in general. So one nice way to do this would be to use image streams, and either will them from a build configuration, or via downloading from a registry. So when you want "S2I", then you would only need different sources of the images in the image stream. However that doesn't work with Kubernetes, so I guess this isn't a real option.

In the way you suggested above, to only have one "s2i" module, then that would contain deployments for EnMasse, Monitoring and IoT. Wouldn't it make sense to simply have two modules "enmasse" and "enmasse_s2i"? I would like to keep the subdirectory though to make it more obvious of what is what. But the final output can be a single archive/directory.

lulf:
@ctron Based on our past experience with providing different bundles, having copies of deployments for different bundles make it a bit harder to make changes as you have to do so everywhere. So instead of viewing bundles as 'all-in-one' thing, I think they should be more like modules like 'core', 'monitoring', 'iot'. Etc, and that we should try keep as few of them as possible. 

So what I was thinking is that we could have something like this:

```
install/bundles/enmasse
install/bundles/monitoring
install/bundles/iot
install/components/s2i // Maybe just buildconfigs?
install/components/imagestreams
```

I.e. we would not create bundles with the s2i stuff, but if you want to use s2i to install, the set of commands you need to run is

```
oc apply -f install/components/imagestreams
oc apply -f install/components/s2i
oc apply -f install/bundles/enmasse
oc apply -f install/bundles/iot
```

My thinking was that we would only use Kubernetes resources like Deployment, and simply put the annotations required to use imagestreams. The annotations will (hopefully) only take effect if imagestreams etc. are defined.

ctron:
Hm, so you would put the imagestrream-trigger annotations there in any case, with the assumption that they will override the provided image name. Ok, I will test if that works that way.

ctron:
So I created the remaining deployments. The current state is:

* Each component now has two directory variants (`_s2i` for S2I, and `_img` for plain image), in addition the the base directory which contains the common part. So e.g. for the "HTTP adapter" this is `http-adapter`, `http-adapter-s2i`, `http-adapter-img`. The idea is that you need to deploy `http-adapter` + either `http-adapter-s2i` or `http-adapter-img`.
* Currently the deployment still uses the internal service CA for internal TLS certificates, this is OpenShift specific and will cause some effort if we want to delete this. Most likely by writing the install operator for as a next step.
* The difference between is mostly the added ImageStream and BuildConfig, and the relatively small annotation in the Deployment.

My proposal would be to take this as an intermediate step, and do the install operator as the next task. Because that would effectively drop all of this.

Makefile.env.mk
```diff
@@ -39,7 +39,19 @@ PROMETHEUS_IMAGE ?= "prom/prometheus:v2.4.3"
 ALERTMANAGER_IMAGE ?= "prom/alertmanager:v0.15.2"
 GRAFANA_IMAGE ?= "grafana/grafana:5.3.1"
 KUBE_STATE_METRICS_IMAGE ?= "quay.io/coreos/kube-state-metrics:v1.4.0"
+QDROUTERD_BASE_IMAGE ?= "enmasseproject/qdrouterd-base:1.5-rc2"
```
lulf: This is already out of date, assuming you want to use the same as the router base :)

templates/iot/auth-s2i/050-Deployment-service-auth.yaml
```diff
@@ -0,0 +1,85 @@
...
+      [
+        {
+          "from":
+            {
+              "kind":"ImageStreamTag",
+              "name":"iot-auth-service:latest"
+            },
+            "fieldPath":"spec.template.spec.containers[?(@.name==\"auth-service\")].image"
+        }
+      ]
```
lulf: What I was imagining was that you could have this annotation and it would be ignored if you did not have an imagestreamtag that matched. Is that not the case?

PR #2364, opened 2019-02-22
Systemtests infra configs

kornys:


kornys:
@enmasse-ci run tests

kornys:
@enmasse-ci run tests

lulf:
@kornys Seems like a lot of work to maintain Java POJOs for systemtests, would it be possible to reuse classes from api-model, or do you want it to be an 'independent' implementation?

kornys:
@lulf yeah it is a lot of work, that would be a good idea, but is it posible, with current state of classes in api-model, convert obejct to json like in systemtests classes? Or we need to modify classes in api-model?

lulf:
You just need an ObjectMapper instance  and the objects support json encoding.

Use mapper.writeValueAsString. Look for it’s use in api-model

kornys:
@lulf so do you think we can reuse AddressSapceBuilder etc... in systemtests?

lulf:
@kornys yes they are part of api model(generated)

kornys:
@lulf ok I will create task for that and will do that in different PR

Issue #2366, opened 2019-02-23
Responses to Agent's management requests fail if Broker has exceeded configured global-max-size


k-wall:
https://issues.apache.org/jira/browse/ARTEMIS-2261

Testing on Master, I notice the following issue when the a Broker becomes full.

Artemis 2.6.4 includes a fix for https://issues.apache.org/jira/browse/ARTEMIS-1710, however, whilst this fix now means a full Broker accepts management requests sent to `activemq.management`, internally the Broker can still fail to route to the responses unless the response queue happens to have been created with a name prefixed by `activemq.management`.  Agent (artemis.js) doesn't do this, so the responses are lost. 

I think it would be possible to argue that the fix for ARTEMIS-1710 is incomplete, but it will probably be possible to work around by influencing the name given to the temporary reply queue.

=== 

From a messaging tenant perspective, this issue manifests as an inability for Enmasse to problem control the Broker.  An example failing use-case:

1. Create two small queue sharing a broker instance
2. Fill the first queue until the broker's global-max-size is exceeded.  The application will see rejected messages.
3. Delete the first address.
4. `oc get address` reports the address gone
5. Fill the second queue.  **Expected** behaviour: queue accepts messages until full **Actual**: queue refused messages as Broker is still full (**DEFECT**)

Connecting to the Broker management, you'll see the queue backing the address is still present and still full.

Issue would have existed in versions of EnMasse using previous versions of Artemis (2.6.3 and below), but in those versions the management request would have been rejected rather than the response lost.

==

The Broker logs look like this:

```
2019-02-19T20:38:43.356Z WARN  [ProtonServerReceiverContext] AMQ229102: Address "54111434-5686-4dbb-8a94-dd0182dbe7eb" is full.: ActiveMQAddressFullException[errorType=ADDRESS_FULL message=AMQ229102: Address "54111434-5686-4dbb-8a94-dd0182dbe7eb" is full.]
	at org.apache.activemq.artemis.core.paging.impl.PagingStoreImpl.page(PagingStoreImpl.java:780)
	at org.apache.activemq.artemis.core.persistence.impl.journal.AbstractJournalStorageManager.addToPage(AbstractJournalStorageManager.java:2019)
	at org.apache.activemq.artemis.core.postoffice.impl.PostOfficeImpl.processRoute(PostOfficeImpl.java:1203)
[...]

PR #2367, opened 2019-02-24
Prefix agent/broker reply-to address with "activemq.management"

k-wall:


lulf:
This needs to be fixed in standard controller as well which checks queue message counts as part of checking if an address is drained on a broker. Would that complicate cleanup of those queues?

k-wall:
> This needs to be fixed in standard controller as well which checks queue message counts as part of checking if an address is drained on a broker. Would that complicate cleanup of those queues?

Yes, indeed, ProtonRequestClient would need the same 'fix'.   Agent's synch process will take care of deleting the old response queues.  Attempts to delete live response queues will fail as they have a receiver link attached which Artemis disallows.  You actually already see Agent sync trying and failing to delete dynamic response queues today, so in this respect, it is no worse.

k-wall:
I tried making changes to StandardController.  I wasn't happy with the results.  The complexity was sending the request for the removal of your response queue when you still have a receiving link attached to the queue.   I tried a couple of approaches (sending a one shot request, closing the receiving link first), but the code was displeasing.   Also the risk of queue cleanup going wrong and queues accumulating was troubling.   This would create a bigger problem than the one I am trying to solve.

On balance I think leaving standard-controller as it stands is better.  If a broker that is being drained is full and the depth query fails, the upshot is the BrokerStatusCollector assumes it is not ready for removal.  This is safe.  Once the application drains the queue a but, the depth queries will start to work again.  We can live with the noise in the logs.  If  ARTEMIS-2262 gets implemented, we can revisit.

I do think the changes for agent are worth having.  Currently the user's ability to control addresses when the broker is full will be impaired by the defect.

k-wall:
The agent now uses Artemis management to create a non-durable, purge on no consumers, exclusive reply-to queue named `activemq.management.tmpreply.<uuid>`.  Old response queues will be cleaned up either by the action of the address synch (which deletes for queues so named with 0 consumers attached), or Broker restart.  The `purgeOnNoConsumers` is belt and braces. - it simple ensures unconsumed management responses are purged as soon as possible.

I updated the original issue #2366 with a description of the issue from the messaging tenant perspective.  We ought to add a system test ensuring the system behaviours problem under the broker full boundary condition. This should be easy now we have the convenient ability to control the global-max-size.

lulf:
@enmasse-ci run tests

PR #2369, opened 2019-02-24
Fix #1279 : JSON patch / JSON merge patch support (for address spaces)

### LINKED ISSUE #1279: Add support for PATCH to api server
Allowing PATCH/json-patch support will allow 'oc edit addressspace foo' to work, which will be a common operation if changing plans etc. Arguably we might want to support changing plans before we add this support.

k-wall:


k-wall:
How does this approach to adding patch support look?   I supported both JSON merge patch (RFC7386) (which seems to be what is required to support `oc edit`) and JSON patch (RFC6902).
Currently the Swagger API is announcing only the latter (I need to look how best to support both).

Once the approach looks reasonable, I'll expand to cover addresses and messaging users(?).

api-server/src/main/java/io/enmasse/api/v1/http/HttpAddressSpaceService.java
```diff
@@ -119,6 +124,65 @@ public Response createAddressSpace(@Context SecurityContext securityContext, @Co
...
+
+
+        return doPatch(securityContext, namespace, addressSpaceName,
+                jsonNode -> patch.apply(jsonNode));
+    }
+
+    private Response doPatch(@Context SecurityContext securityContext, @PathParam("namespace") String namespace, @PathParam("addressSpace") String addressSpaceName,
+                             CheckedFunction<JsonNode, JsonNode, JsonPatchException> patcher) throws Exception {
+        return doRequest("Error patching address space " + addressSpaceName, () -> {
+            verifyAuthorized(securityContext, namespace, ResourceVerb.update);
```
lulf: This should use the verb patch i think

k-wall: > This should use the verb patch i think

I don't understand the comment. Could you clarify? This private method is a shared utility method that is shared by the two public methods that are annotated @PATCH.

lulf: I mean ResourceVerb.patch instead of ResourceVerb.update. Kubernetes uses patch in RBAC to authorize the patch operation.

See https://kubernetes.io/docs/reference/access-authn-authz/rbac/

k-wall: Ah, I see, yes I'll fix that.

api-server/src/main/java/io/enmasse/api/v1/http/HttpAddressSpaceService.java
```diff
@@ -207,6 +206,10 @@ private AddressSpace setAddressSpaceDefaults(SecurityContext securityContext, St
             }
         } else {
             validateChanges(existing, addressSpace);
+
+            // Question - not sure of the intent here. This prevents the removal of any annotation (or label) which
```
k-wall: @lulf the existing code we have here prevents any label or annotation from being removed (either with patch or replace), even on that the user defines himself from his own domain.  This seems quite restrictive.   Addresses are similarly restrictive.

api-server/src/main/resources/swagger.json
```diff
@@ -733,6 +733,48 @@
...
             },
+            "patch": {
+                "tags": [
+                    "enmasse_v1beta1",
+                    "addresses"
+                ],
+                "description": "patches (RFC6902) the specified Address",
+                "operationId": "patchesEnmasseV1alpha1NamespacedAddress",
+                "consumes": [
+                    "application/json-patch+json"
```
k-wall: This version of Swagger doesn't permit the definition of requests that support many payload types.  I made an arbitrary choice of RFC6092  I read that OpenAPI v3 brings this ability.

api-server/src/main/java/io/enmasse/api/v1/http/HttpAddressSpaceService.java
lulf: Sorry @k-wall, resolved this by mistake. I agree we should relax the restriction to only cover our own annotations

lulf: Sorry @k-wall, resolved this by mistake. I agree we should relax the restriction to only cover our own annotations

k-wall:
> +
> +            // Question - not sure of the intent here. This prevents the removal of any annotation (or label) which
> Sorry @k-wall, resolved this by mistake. I agree we should relax the restriction to only cover our own annotations

Thanks.  As this relaxation of the annotation validation is broader concern than PATCH support, I think it would be best to commit this PR as is (with my questions removed) and raise a support PR to cover the extra work.

api-server/src/main/java/io/enmasse/api/v1/http/HttpAddressServiceBase.java
```diff
@@ -282,6 +284,33 @@ protected Response internalDeleteAddresses(SecurityContext securityContext, Stri
...
+            verifyAuthorized(securityContext, namespace, ResourceVerb.patch);
+
+            Optional<Address> existing = apiHelper.getAddress(namespace, addressSpace, addressName);
+
+            if (!existing.isPresent()) {
+                return Response.status(404).entity(Status.notFound("Address", addressName)).build();
+            }
+            Address existingAddress = existing.get();
+
+            ObjectMapper mapper = new ObjectMapper();
```
lulf: Nitpick ,but maybe make this static final and just reuse it?

api-server/src/main/resources/swagger.json
```diff
@@ -234,6 +234,48 @@
                     }
                 }
             },
+            "patch": {
+                "tags": [
+                    "enmasse_v1beta1",
+                    "addressspaces"
+                ],
+                "description": "patches (RFC6902) the specified AddressSpace",
+                "operationId": "patchesEnmasseV1alpha1NamespacedAddressSpace",
```
lulf: Just a note here that it should probably be be V1beta1 in the name, but I understand this is an issue with the existing operationIds as well.

lulf:
@enmasse-ci run tests

api-server/src/main/resources/swagger.json
k-wall: @lulf  I'll fix all of these in one go, under a separate issue.

api-server/src/main/java/io/enmasse/api/v1/http/HttpAddressServiceBase.java
k-wall: Fixed.

PR #2379, opened 2019-02-27
Go native makefile and build fixes

lulf:


lulf:
@enmasse-ci run tests

ctron:

ctron:
I absolutely like the cleanup.

I would suggest to also kick out: https://github.com/EnMasseProject/enmasse/blob/15916e49c76d4ab384c6776675458dfe3dcac996/.travis.yml#L81as it should no longer be needed in Travis.

The only problem now is, that you may have some messed up GOPATH in your local system, which creates a different result than when running on a CI system. Yes that is what a CI system is for, but to my understanding GOPATH should project specific, and not system wide. Yes, you have the same problem with maven dependencies, but then they are much more stable than go stuff. Anyway, there is a CI system for finding such issues. You only search a bit longer why it fails in CI but works locally.

What I don't like in this PR is the collapsing of the .PHONY target, as all those one-liner targets make it hard to find changed and cause merge conflicts. I think splitting them up make the Makefile a bit easier to handle.

lulf:
> I would suggest to also kick out: https://github.com/EnMasseProject/enmasse/blob/15916e49c76d4ab384c6776675458dfe3dcac996/.travis.yml#L81as it should no longer be needed in Travis.

I failed to notice that, thanks.

> The only problem now is, that you may have some messed up GOPATH in your local system, which creates a different result than when running on a CI system. Yes that is what a CI system is for, but to my understanding GOPATH should project specific, and not system wide.

According to https://github.com/golang/go/wiki/GOPATH#use-a-single-gopath there seems to be a convention to use a single GOPATH for a system. Looking at a project like operator-sdk, there doesn't seem to be a precedence in modifying GOPATH to perform a build.

I think trying to make CI identical to local dev is a good target, but I also think that we should setup CI to work as a local dev environment rather than the other way around.

I will revert the changes to the .PHONY target (and group things a bit more logically maybe).

Also I will look at failing jenkins build :)

ctron:
> According to https://github.com/golang/go/wiki/GOPATH#use-a-single-gopath there seems to be a convention to use a single GOPATH for a system.

And then you have a tutorial which suggest otherwise :wink: -> https://github.com/golang/go/wiki/GithubCodeLayout#setup-the-workspace

And you also have the problem that a single GOPATH can only hold one version of a dependency, and you spend hours finding out why your application doesn't work, just because there was a different version in your GOPATH already, … So I would strongly suggest to not use a single GOPATH.

lulf:
> And then you have a tutorial which suggest otherwise wink -> https://github.com/golang/go/wiki/GithubCodeLayout#setup-the-workspace

I don't see how this mentions or advocates the use of multiple GOPATH's? It actually creates a single GOPATH at `workspace/go` and clones multiple projects beneath the gopath. Am I missing something?

lulf:
> And you also have the problem that a single GOPATH can only hold one version of a dependency, and you spend hours finding out why your application doesn't work, just because there was a different version in your GOPATH already, … So I would strongly suggest to not use a single GOPATH.

Aren't the dependencies anyway in `vendor/` usually?

lulf:

ctron:
> > And then you have a tutorial which suggest otherwise wink -> https://github.com/golang/go/wiki/GithubCodeLayout#setup-the-workspace
> 
> I don't see how this mentions or advocates the use of multiple GOPATH's? It actually creates a single GOPATH at `workspace/go` and clones multiple projects beneath the gopath. Am I missing something?

~~~
cd ~/workspace # Standard location for Eclipse workspace
mkdir mygo # Create your Go workspace
export GOPATH=~/workspace/mygo # GOPATH = Go workspace
cd $GOPATH
~~~

lulf:
How is this related to setting the GOPATH in a projects build system or not. In this example, the user is setting the GOPATH and is checkout out the projects. The projects assume that GOPATH is set so that they can build. The project build systems don't modify GOPATH to point somewhere else.

Isn't this the same assumption that this PR will make for enmasse?

ctron:
> Aren't the dependencies anyway in `vendor/` usually?

Only if you opt-in to that idea (which you should). The default experience of go is to not use vendor folders, but to use "go get". And you can find a bunch tools that don't use the vendor folder.

But that will all be better with go modules :roll_eyes:

ctron:
> The project build systems don't modify GOPATH to point somewhere else.

Which brings us back to the experience the user should have with building EnMasse. EnMasse uses a Makefile, which most go projects don't. And if you check out EnMasse now in a different directory structure than a "go path" compatible one, which Java developers have no idea about, then you can't build EnMasse by simply running "make" …

lulf:
>  Only if you opt-in to that idea (which you should). The default experience of go is to not use vendor folders, but to use "go get". And you can find a bunch tools that don't use the vendor folder.

I consider this a non-issue then, as enmasse uses vendoring

> Which brings us back to the experience the user should have with building EnMasse. EnMasse uses a Makefile, which most go projects don't. And if you check out EnMasse now in a different directory structure than a "go path" compatible one, which Java developers have no idea about, then you can't build EnMasse by simply running "make" …

I think the moment we decided to put Go code in EnMasse, we should adopt the `go` development process, mainly because it is the most 'picky' about the location of the source code.

As we gradually migrate more components to Go, I think this will be less controversial.

Having these workarounds to avoid the reality that we're a go project can be a lot harder to maintain than just stating "to build enmasse, clone the repo into $GOPATH/src/enmasseproject/enmasse" IMO.

Maybe the best way to resolve this disagreement is to have a vote and stick with that decision.

@robshelly @k-wall @rgodfrey @vbusch @kornys @famartinrh please state your position!

lulf:
@enmasse-ci run tests testcase=**SmokeTest

lulf:

lulf:

lulf:

lulf:

lulf:
Some other time...

PR #2386, opened 2019-02-28
Fixes for running systemtests on different node

kornys:


kornys:
@enmasse-ci run tests

.jenkins/jenkins-functions.groovy
```diff
@@ -98,7 +98,6 @@ def postAction(String coresDir, String artifactDir) {
     if (status == null) {
         currentBuild.result = 'SUCCESS'
     }
-    tearDownOpenshift()
```
lulf: We don't need this for PRs etc?

kornys: nope, since openstack server is deleted immediately after test run

PR #2389, opened 2019-03-01
Fix #2305: Bump keycloak from 3.4.3.Final to 4.8.3.Final

### LINKED ISSUE #2305: Upgrade keycloak to 4.8.3
Release notes: https://www.keycloak.org/docs/latest/release_notes/index.html
Upgrade instructions: https://www.keycloak.org/docs/4.8/upgrading/

k-wall:


k-wall:
With this change, keycloak realm creation requests made by keycloak-controller were failing with a 400 response from the keycloak server.  There was no error logged server, even at trace.  I found that when Keycloak Client API + RESTEasy serialise the Realm request, it is serialised fully, including null values for config items that have no value.    I noticed if I hand crafted a minimal request (like the UI does) all is well.  

 bc59951 reconfigures to ResteasyJackson2Provider to use `ObjectMapper.setSerializationInclusion(JsonInclude.Include.NON_NULL)`.  This is a workaround.

An an experiment, I rolled back to Keycloak 3.4.3 and examined the requests that it sends.  Its realm requests are 'maximal' (with nulls for unset fields), this leads me to believe that it is likely the issue is Keycloak server side.  I need to raise an issue against this project after putting together a mimimal reproduction.

k-wall:
@enmasse-ci run tests

k-wall:
@enmasse-ci run tests testcase=UserApiTest

PR #2397, opened 2019-03-04
Fix reference to address space plan in agent

lulf:
Should fix console tests


kornys:

PR #2407, opened 2019-03-05
Add IoTConfig CRD and controller

ctron:


pkg/util/install/utils.go
```diff
@@ -0,0 +1,174 @@
...
+			"app":               app,
+			"name":              name,
+		},
+	}
+
+	if deployment.Spec.Template.ObjectMeta.Labels == nil {
+		deployment.Spec.Template.ObjectMeta.Labels = make(map[string]string)
+	}
+
+	deployment.Spec.Template.ObjectMeta.Labels["enmasse-component"] = component
```
lulf: I think just 'component' is sufficient. If you have other things using that label, you can further distinguish using app=enmasse.

ctron: True. Changed and pushed.

PR #2410, opened 2019-03-06
Upgrade to Hono 0.9

ctron:


iot/iot-application-base/pom.xml
```diff
@@ -0,0 +1,38 @@
...
+
+    <dependency>
+      <groupId>io.netty</groupId>
+      <artifactId>netty-tcnative</artifactId>
+    </dependency>
+
+    <dependency>
+      <groupId>io.micrometer</groupId>
+      <artifactId>micrometer-registry-prometheus</artifactId>
+    </dependency>
```
lulf: Should these be added to the toplevel pom dependencyManagement as well? CC @k-wall

ctron: Currently it looks like no one else is using that. And we had some issues updating from 1.0.x to 1.1.x in Hono. So it might be a good idea to keep those separate in any case.

But it would also be OK to pull them up.

PR #2409, opened 2019-03-05
Re-add surefire provider to exclude tests

lulf:

PR #2411, opened 2019-03-06
Use surefire provider

kornys:


lulf:
@kornys i added it back in #2409

kornys:
@lulf ah I see

PR #2415, opened 2019-03-06
Authentication service type and operator

lulf:


lulf:
@enmasse-ci run tests

lulf:
@rgodfrey, could you have a look at these docs, especially the latter:

* https://github.com/lulf/enmasse/blob/51148c0b9f8e7e15a07bc55db8bdd706c391f576/documentation/modules/con-authentication-services.adoc
* https://github.com/lulf/enmasse/blob/51148c0b9f8e7e15a07bc55db8bdd706c391f576/documentation/modules/con-external-authentication-service-api.adoc

rgodfrey:
> @rgodfrey, could you have a look at these docs, especially the latter:
> 
> * https://github.com/lulf/enmasse/blob/51148c0b9f8e7e15a07bc55db8bdd706c391f576/documentation/modules/con-authentication-services.adoc
> * https://github.com/lulf/enmasse/blob/51148c0b9f8e7e15a07bc55db8bdd706c391f576/documentation/modules/con-external-authentication-service-api.adoc

@lulf There are a couple of places where the example yaml / json is broken (missing " or : ) ... also I'd probably avoid mentioning the view permission, as I don't think we use it anywhere and it doesn't make sense outside of consoles / custom resources.  Otherwise looks good.

documentation/modules/con-authentication-services.adoc
```diff
@@ -0,0 +1,83 @@
...
+      type: persistent-claim
+      size: 5Gi
+    datasource: <6>
+      type: postgresql
+      host: example.com
+      port: 5432
+      database: authdb
+----
+<1> The type can be one of `none`, `standard` and `external`.
+<2> (Optional) The secret must contain a field `admin.username` for the user and `admin.password` for the password of the {AuthProductName} admin user. If not specified, a random password will be generated and stored in a secret.
```
jenmalloy: ...contain the `admin.username` field for the user...

documentation/modules/con-external-authentication-service-api.adoc
```diff
@@ -0,0 +1,62 @@
...
+authentication and authorization information. The authentication service is queried by the address
+space components such as the router and broker, whenever a new connection is established to the
+messaging endpoints.
+
+== Authentication
+
+The requested identity of the client can be read from the SASL handshake `username`. The
+implementation can then authenticate the user.
+
+The authenticated identity should be returned in the `authenticated-identity` map with the following
```
jenmalloy: Change "should" to "will"

documentation/modules/con-external-authentication-service-api.adoc
```diff
@@ -0,0 +1,62 @@
...
+        "sub: "myid",
+        "preferred_username": "myuser"
+    }
+}
+----
+
+== Authorization
+
+Authorization is a capability that can be requested by the client using the `ADDRESS-AUTHZ`
+connection capability. If this is set on the connection, the server should respond with this
```
jenmalloy: Can "should" be deleted and still retain the meaning?

documentation/modules/con-external-authentication-service-api.adoc
```diff
@@ -0,0 +1,62 @@
...
+capability in the offered capabilities, and add the authorization information to the connection
+properties.
+
+The authorization information is contained within a map of address to list of operations allowed on
```
jenmalloy: ...is stored within a map that correlates the address to a list of operations allowed on that address.

lulf:
@jenmalloy Update the docs

pkg/controller/authenticationservice/authenticationservice_controller.go
```diff
@@ -0,0 +1,444 @@
...
+	return r.updateStatus(ctx, authservice, func(status *adminv1beta1.AuthenticationServiceStatus) error {
+		status.Host = authservice.Spec.External.Host
+		status.Port = authservice.Spec.External.Port
+		status.CaCertSecret = authservice.Spec.External.CaCertSecret
+		status.ClientCertSecret = authservice.Spec.External.ClientCertSecret
+		return nil
+	})
+}
+
+func equalSecretRef(a *corev1.SecretReference, b *corev1.SecretReference) bool {
```
ctron: You might want to have a look at `reflect.DeepEqual()`

pkg/controller/authenticationservice/authenticationservice_controller.go
```diff
@@ -0,0 +1,444 @@
...
+	newStatus := adminv1beta1.AuthenticationServiceStatus{}
+	updateFn(&newStatus)
+
+	if authservice.Status.Host != newStatus.Host ||
+		authservice.Status.Port != newStatus.Port ||
+		!equalSecretRef(authservice.Status.CaCertSecret, newStatus.CaCertSecret) ||
+		!equalSecretRef(authservice.Status.ClientCertSecret, newStatus.ClientCertSecret) {
+
+		authservice.Status = newStatus
+		err := r.client.Update(context.TODO(), authservice)
```
ctron: I think you should take the context provided to the function.

pkg/controller/authenticationservice/authenticationservice_controller.go
```diff
@@ -0,0 +1,444 @@
...
+			return reconcile.Result{}, err
+		}
+	}
+	return reconcile.Result{}, nil
+}
+
+func (r *ReconcileAuthenticationService) keycloakOauthUri(ctx context.Context, authservice *adminv1beta1.AuthenticationService) (string, error) {
+
+	route := &routev1.Route{}
+	err := r.client.Get(context.TODO(), types.NamespacedName{Name: authservice.Name, Namespace: authservice.Namespace}, route)
```
ctron: Same issue with the context, … also at a few other locations.

pkg/controller/authenticationservice/none.go
```diff
@@ -0,0 +1,82 @@
...
+	install.ApplyServiceDefaults(service, "none-authservice", authservice.Name)
+	if service.Annotations == nil {
+		service.Annotations = make(map[string]string)
+	}
+	service.Annotations["service.alpha.openshift.io/serving-cert-secret-name"] = "none-authservice-cert"
+	service.Spec.Ports = []corev1.ServicePort{
+		{
+			Port:     5671,
+			Protocol: corev1.ProtocolTCP,
+			TargetPort: intstr.IntOrString{
```
ctron: There are two functions for creating that `intstr.From<Type>`

pkg/controller/authenticationservice/standard.go
```diff
@@ -0,0 +1,216 @@
...
+		return err
+	}
+	return nil
+}
+
+func applyStandardAuthServiceDeployment(authservice *adminv1beta1.AuthenticationService, deployment *appsv1.Deployment) error {
+
+	install.ApplyDeploymentDefaults(deployment, "standard-authservice", authservice.Name)
+
+	install.ApplyInitContainerWithError(deployment, "keycloak-plugin", func(container *corev1.Container) error {
```
ctron: Ignoring the error will cause a warning in some IDEs. However there is a variant of this function `ApplyInitContainer` which does not require an error.

pkg/util/images/mapping.go
```diff
@@ -44,7 +44,7 @@ func loadImageMapFrom(fileName string) (map[string]string, error) {
 
 }
 
-const defaultImageMapFileName = "/etc/operatorImageMap.yaml"
+const defaultImageMapFileName = "operatorImageMap.yaml"
```
ctron: I am not sure why you need this change? My assumption would be that it breaks the current deployment.

pkg/util/math.go
```diff
@@ -19,3 +20,12 @@ func Max(x, y int64) int64 {
 func MaxDuration(x, y time.Duration) time.Duration {
 	return time.Duration(Max(int64(x), int64(y)))
 }
+
+func GenerateAlphaNumeric(len int) string {
```
ctron: That already exists in `util.GeneratePassword()`.

pkg/util/math.go
```diff
@@ -19,3 +20,12 @@ func Max(x, y int64) int64 {
...
+	bytes := make([]byte, len)
+	for i := 0; i < len; i++ {
+		bytes[i] = byte(65 + rand.Intn(25))
+
+	}
+	return string(bytes)
```
ctron: This might fail due to the fact that you are not ensuring proper UTF-8 with the random bytes. But I would recommend anyway to drop this in favor of `util.GeneratePassword()`.

pkg/util/images/mapping.go
lulf: In the controller unit tests, it is required to use this map as it will try to lookup the image. It was also convenient when testing it locally not having to set the env.

How about setting the env to the desired location in the Dockerfile? This way we have the location only known in one place.

ctron: So the argument is to have it more convenient when debugging, rather than having a predictable default when running properly?

For the tests we would manually load this from a different location. You can simple override the loaded map in the unit tests.

lulf: > So the argument is to have it more convenient when debugging, rather than having a predictable default when running properly?

I think its good practice/not controversial to not have hardcoded paths in the application code, and instead make it part of the deployment configuration, i.e. env variables or command line arguments in the Kubernetes deployment. Ideally I think it should be a 'required' command line/env argument.

ctron: I would prefer it if the container starts up with sane defaults, without the need to provide additional configuration for the default use case. If that means baking in a an environment variable into the container image than this will work as well.

Grepping through the source code of kubernetes and client-go, I see the pattern of having a bunch of override capabilities (like env vars of command line arguments), and then falling back to a hardcoded value. Which is exactly what we have here as well. But again, my focus is on having a seamless default startup.

lulf: Updated it to set ENMASSE_IMAGE_MAP_FILE in the Dockerfile

pkg/util/math.go
lulf: Done

pkg/util/math.go
lulf: Done

pkg/controller/authenticationservice/standard.go
lulf: Fixed

lulf:
@enmasse-ci run tests testcase=standard.SmokeTest

lulf:
@enmasse-ci run tests testcase=standard.SmokeTest

lulf:
@enmasse-ci run tests

lulf:
@enmasse-ci run tests

lulf:
@enmasse-ci run tests

lulf:
@enmasse-ci run tests testcase=web.FirefoxWebConsoleTest#testCreateDeleteQueue

lulf:
@enmasse-ci run tests testcase=web.FirefoxWebConsoleTest#testCreateDeleteQueue

lulf:
@enmasse-ci run tests

lulf:
@enmasse-ci run tests testcase=CommonTest

lulf:
@enmasse-ci run tests testcase=brokered.auth.AuthenticationTest

PR #2416, opened 2019-03-06
 Add the next IoT component to the EnMasse Controller 

ctron:
This PR builds on top of PR #2407 … adding the next IoT component


cmd/enmasse-controller-manager/main.go
```diff
@@ -58,8 +60,13 @@ func main() {
 
 	// register APIs
 
+	if err := api.Install(scheme.Scheme); err != nil {
+		log.Error(err, "Failed to register OpenShift schema")
```
lulf: s/OpenShift/Kubernetes

pkg/controller/iotconfig/auth_service.go
```diff
@@ -0,0 +1,226 @@
...
+	rc.ProcessSimple(func() error {
+		return r.processConfigMap(ctx, nameAuthService+"-config", config, r.reconcileAuthServiceConfigMap)
+	})
+
+	return rc.Result()
+}
+
+func (r *ReconcileIoTConfig) reconcileAuthServiceDeployment(config *iotv1alpha1.IoTConfig, deployment *appsv1.Deployment) error {
+
+	install.ApplyDeploymentDefaults(deployment, "iot", appAuthService, deployment.Name)
```
lulf: It would be nice to keep app: enmasse for all components, this makes it easier to find out which components are deployed by enmasse when you're looking at the whole cluster. Maybe you can add another label to be able to filter 'iot' components?

pkg/controller/iotconfig/auth_service.go
```diff
@@ -0,0 +1,226 @@
...
+	service.Spec.Ports[0].Port = 5671
+	service.Spec.Ports[0].TargetPort = intstr.FromInt(5671)
+	service.Spec.Ports[0].Protocol = corev1.ProtocolTCP
+
+	if service.Annotations == nil {
+		service.Annotations = make(map[string]string)
+	}
+
+	// FIXME: remove OpenShift specific feature
+	service.Annotations["service.alpha.openshift.io/serving-cert-secret-name"] = nameAuthService + "-tls"
```
lulf: FWIW, one approach I'm taking in the authservice controller is to allow a secret with the certificate to be specified in the CRD. On Kubernetes, this is the only way to get the cert, whereas on OpenShift, we can 'default' to this annotation.

pkg/util/openshift.go
```diff
@@ -0,0 +1,31 @@
...
+	if openshift == nil {
+		b := detectOpenshift()
+		openshift = &b
+	}
+	return *openshift
+}
+
+func detectOpenshift() bool {
+
+	value, ok := os.LookupEnv("ENMASSE_OPENSHIFT")
```
lulf: We should detect it by querying the API, like in https://github.com/EnMasseProject/enmasse/blob/master/keycloak-controller/src/main/java/io/enmasse/keycloak/controller/KeycloakController.java#L170 .

templates/enmasse-controller-manager/020-ClusterRole-enmasse-controller-manager.yaml
```diff
@@ -0,0 +1,28 @@
...
+rules:
+  - apiGroups: [ "" ]
+    resources: [ "configmaps", "secrets", "pods", "services", "persistentvolumeclaims" ]
+    verbs: [ "get", "list", "watch", "update", "create", "patch" ]
+  - apiGroups: [ "apps" ]
+    resources: [ "deployments" ]
+    verbs: [ "get", "list", "watch", "update", "create", "patch" ]
+  - apiGroups: [ "route.openshift.io" ]
+    resources: [ "routes" ]
+    verbs: [ "get", "list", "watch", "update", "create", "patch" ]
```
lulf: We cannot grant the operator cluster-wide privileges to create configmaps, secrets, pods, services, persistentvolumeclaims, deployments, routes etc. as this is too permissive. The operator will have to manage these resource only for its own namespace and should be in the namspaced Role only (as I see you have added).

We can grant clusterwide permissions for our own resources as you have done below.

templates/enmasse-controller-manager/020-ClusterRole-enmasse-controller-manager.yaml
```diff
@@ -0,0 +1,28 @@
...
+  - apiGroups: [ "enmasse.io" ]
+    resources: [ "addressspaces", "addresses" ]
+    verbs: [ "get", "list", "watch", "update", "create", "patch" ]
+  - apiGroups: [ "admin.enmasse.io" ]
+    resources: [ "authenticationservice" ]
+    verbs: [ "get", "list", "watch", "update", "create", "patch" ]
```
lulf: For authenticationservice we really only need access to our own namespace, so this can be removed.

templates/enmasse-controller-manager/020-Role-enmasse-controller-manager.yaml
```diff
@@ -5,9 +5,15 @@ metadata:
...
+  - apiGroups: [ "enmasse.io" ]
+    resources: [ "addressspaces", "addresses" ]
+    verbs: [ "get", "list", "watch", "update", "create", "patch" ]
   - apiGroups: [ "admin.enmasse.io" ]
     resources: [ "authenticationservice" ]
     verbs: [ "get", "list", "watch", "update", "create", "patch" ]
   - apiGroups: [ "iot.enmasse.io" ]
-    resources: [ "iotconfig" ]
+    resources: [ "iotconfigs", "iotprojects" ]
     verbs: [ "get", "list", "watch", "update", "create", "patch" ]
```
lulf: I don't see any point in providing namespace access to 'enmasse.io' and 'iot.enmasse.io' resources, as they are already granted by the cluster role.

pkg/controller/iotconfig/auth_service.go
ctron: the problem with that is, that "app" has a special meaning in the web console UI, and so you can group apps, by their different functionality. So instead of having one big blob of services, you have them grouped.

cmd/enmasse-controller-manager/main.go
ctron: Nope, as this actually IS the OpenShift schema.

pkg/controller/iotconfig/auth_service.go
ctron: Yes, my idea was to fix this by generating a certificate in the controller, but simply providing this might be a good alternative for that.

pkg/util/openshift.go
ctron: I wanted to ask you about, sounds like a good idea.

cmd/enmasse-controller-manager/main.go
lulf: Sorry, I didn't read the whole context of the change!

templates/enmasse-controller-manager/020-ClusterRole-enmasse-controller-manager.yaml
ctron: That might be an issue with the operator framework then … from what I saw you need to configure this (namedspaced vs non-namespaced) for the whole controller … so we might need to split that up into two applications then.

pkg/controller/iotconfig/auth_service.go
lulf: In this case though, I think it's more important to adhere to existing conventions, i.e. tag all enmasse resources with 'app', because it has implications for those expecting this label to exist.

templates/enmasse-controller-manager/020-Role-enmasse-controller-manager.yaml
ctron: True.

templates/enmasse-controller-manager/020-ClusterRole-enmasse-controller-manager.yaml
ctron: Done.

templates/enmasse-controller-manager/020-ClusterRole-enmasse-controller-manager.yaml
lulf: I think you can configure to watch local namepace, and watch resources at the cluster level individually. Many users will be upset by this level of privilege requirements, so we cannot do cluster-wide watching on resources we don't own ourselves.

pkg/controller/iotconfig/auth_service.go
ctron: > FWIW, if you want to influence UI on OpenShift 3.x, you can set annotations on the service to provide some level of grouping.

Do you have a pointer on this?

lulf: *digs up enmasse 0.15.3*:

https://github.com/EnMasseProject/enmasse/blob/release-0.15.3/templates/include/messaging-service.jsonnet#L78

https://docs.openshift.com/container-platform/3.5/dev_guide/templates.html#other-recommendations

ctron: Dependencies is something different however.

lulf: Yeah I was hoping there was some additional annotations you could use. In any case I think its more important to make `oc`/`kubectl` experience as good as possible, and be able to query for all enmasse components (which is a superset of the 'iot' group). 

I wouldn't worry to much about the UI anyway since it will change.

ctron: This is how it looks in the console (single app):

image

Multiple apps:

image

ctron: Done

pkg/util/openshift.go
ctron: Done.

pkg/util/install/utils.go
```diff
@@ -6,40 +6,51 @@
...
+// Apply some default service values
+func ApplyServiceDefaults(service *corev1.Service, component string, name string) {
+
+	ApplyDefaultLabels(&service.ObjectMeta, component, name)
+	service.Spec.Selector = createDefaultLabels(nil, component, name)
+
+}
+
+// Apply some default deployment values
+func ApplyDeploymentDefaults(deployment *appsv1.Deployment, component string, name string) {
```
lulf: It would be nice to keep the existing components vs app distinction, because 'component' could be more generic than the 'name'. I.e. app: enmasse, component: router, name: router-xyz

pkg/controller/iotconfig/auth_service.go
ctron: > On Kubernetes, this is the only way to get the cert, …

I think one could also use the operator on plain k8s … https://github.com/openshift/service-ca-operator

Would it be sufficient to provide the name of the secret in the CRD? Or do you want to have the full content?

templates/enmasse-controller-manager/020-ClusterRole-enmasse-controller-manager.yaml
ctron: > I think you can configure to watch local namepace, and watch resources at the cluster level for a particular type.

I fail to see that option, do you have a pointer for me?

pkg/util/install/utils.go
lulf: It allows allow you to collect which resources (services, deployments, volumes etc.) belong to a particular component

ctron: I don't really understand what you want me to change here? Do you want the "app" parameter back, always setting it to "enmasse"?

templates/enmasse-controller-manager/020-ClusterRole-enmasse-controller-manager.yaml
lulf: I thought i saw it in the docs somewhere, but now that I'm looking for it I can't find it. Looking at the code, it seems it is not possible through the controller-runtime abstraction but potentially when using informers directly?

pkg/controller/iotconfig/auth_service.go
lulf: I've used the `SecretReference` type for this information. I.e.

```
type AuthenticationServiceSpecNone struct {
  CertificateSecret *corev1.SecretReference       `json:"certificateSecret"`
  ...
}
```

pkg/util/install/utils.go
lulf: Ah, sorry. Again I didn't see the full context and somehow I thought component was the app parameter based on the usage. Please ignore my comment!

pkg/controller/iotconfig/auth_service.go
ctron: This allows using a separate namespace, do we want to support that?

templates/enmasse-controller-manager/020-ClusterRole-enmasse-controller-manager.yaml
ctron: So instead of re-implementing all the existing code, I would say that we simple split it into two application instances, running with different service accounts. We already can do this with the module flags.

lulf: Agreed, we should do this short term. However I think we need to fix this longer term, hopefully there will be support in controller-runtime.

ctron: I split off a second instance, re-using the image, but creating a second service account.

While the service account will be created in the YAML part, the deployment of the operator is being done by the controller-manager. Otherwise the controller-manager would require the permissions to create cluster roles and cluster role bindings, which would somehow seem to contradict the while split.

pkg/controller/iotconfig/auth_service.go
ctron: As discussed, we will do this is a follow up PR.

templates/enmasse-controller-manager/020-ClusterRole-enmasse-controller-manager.yaml
lulf: Not required for this PR, but I think we should deploy the operator explicitly through in a iot/operator/deployment-iot-operator.yaml as well. This makes it clearer that you can opt-in the whole iot-thing. We can then in the first instance also create an enmasse-iot operator in operator hub until the global namespace issue is resolved.

I will follow up this in a subsequent PR, so lets just leave this as is for now.

PR #2419, opened 2019-03-06
Refactored metrics and alerts

robshelly:
* Changed misleading metric names
* Added Prometheus rules for aggregated metrics
* Improved alerts for critical components
* Refactored Grafana dashboard to reflect changes


lulf:
@robshelly Seems to be a unit test that expects 5 metrics, so you just need to update that to expect 4 metrics instead

PR #2437, opened 2019-03-12
Drop deployment of iot-operator by controller

ctron:
Drop IoT operator deployment by the enmasse-controller-manager and do a "manual" deployment.


lulf:
@ctron I've made an identical change in #2415, so If you don't mind I can merge it there.

lulf:
ah I’ll rebase instead

PR #2440, opened 2019-03-12
Unit tests for plan builders

kornys:


api-model/src/main/java/io/enmasse/admin/model/v1/AddressSpacePlanSpec.java
```diff
@@ -81,6 +81,6 @@ public void setAddressPlans(List<String> addressPlans) {
     }
 
     public List<String> getAddressPlans() {
-        return Collections.unmodifiableList(addressPlans);
+        return addressPlans;
```
lulf: This should not be modifyable

kornys: this cause NPE

kornys: @lulf that is the reson why I remove that

lulf:
@enmasse-ci run tests testcase="plans.PlansTest"

lulf:
@enmasse-ci run tests

PR #2443, opened 2019-03-13
Deploy IoT operator on Travis test

ctron:
This PR deploys the enmasse-controller-manager and the iot-operator in an additional travis job. This can be the basis for further running IoT system tests on each PR, without make the Travis build run longer.

~~**Note:** This PR builds on top of #2439 and thus already contains its content. So this PR must be rebased before it can be merged.~~


ctron:
@famartinrh You might want to have a look at this as well, this is the initial system test for IoT, it simply checks of the components got deployed.

kornys:

systemtests/src/test/java/io/enmasse/systemtest/iot/SimpleK8sDeployTest.java
```diff
@@ -0,0 +1,92 @@
...
+
+        final int rc = p.waitFor();
+        if ( rc != 0) {
+            throw new RuntimeException(String.format("Failed to run: %s -> %s", command, rc));
+        }
+    }
+
+    @BeforeAll
+    protected static void setup () throws Exception {
+        execute(new String[] {
```
kornys: @ctron why do you use execution like this? We have executor in systemtest which can work with oc and kubectl

ctron: I didn't know and find … I will migrate that then.

kornys: Or better is it posible to domit via fabric8 kubernetes client.

ctron: True, can you apply a JSON file with it?

ctron: So applying a JSON file doesn't seem to work, as it must have a Java implementation of the JSON classes … 

I would like to simply load that using `kubectl` … however the only `execute` method I found was tied to `minikube`.

ctron: Ok, found something. Replaced it with `KubeCMDClient`.

ctron:
Rebased with master. This one should be good to go now.

systemtests/scripts/run_test_kubernetes.sh
```diff
@@ -40,13 +40,21 @@ mkdir -p standard-authservice-cert/
...
 kubectl create -f ${ENMASSE_DIR}/install/bundles/enmasse
 kubectl create -f ${ENMASSE_DIR}/install/components/none-authservice
 kubectl create -f ${ENMASSE_DIR}/install/components/standard-authservice
 kubectl create -f ${ENMASSE_DIR}/install/components/example-plans
 kubectl create -f ${ENMASSE_DIR}/install/components/example-roles
 
+kubectl create -f ${ENMASSE_DIR}/install/components/enmasse-controller-manager
+kubectl create -f ${ENMASSE_DIR}/install/components/iot/api
+kubectl create -f ${ENMASSE_DIR}/install/components/iot/common
+kubectl create -f ${ENMASSE_DIR}/install/components/iot/operator
```
lulf: Maybe not deploy IoT component unless running IoT tests?

systemtests/src/test/java/io/enmasse/systemtest/iot/SimpleK8sDeployTest.java
```diff
@@ -0,0 +1,80 @@
...
+    static {
+        IOT_LABELS= new HashMap<> ();
+        IOT_LABELS.put("component", "iot");
+    }
+
+    private Kubernetes client = Kubernetes.getInstance();
+
+    @BeforeAll
+    protected static void setup () throws Exception {
+        KubeCMDClient.createFromFile(NAMESPACE, Paths.get("templates/iot/k8s-systemtests-config.json"));
```
lulf: Maybe this is the plan, but it should ideally be created using api-model classes instead of putting yaml/json in the systemtests.

systemtests/scripts/run_test_kubernetes.sh
ctron: That would add another condition in those scripts, if you are fine having this, then I can add this. However this only makes sense to me if we stick to the second Travis job.

systemtests/src/test/java/io/enmasse/systemtest/iot/SimpleK8sDeployTest.java
ctron: So right now we still don't have this API model. And it should be easy to replace this later on.

ctron:
Currently most Travis builds run around 42 minutes. There is a limit of 50 minutes per job. Adding a minute of reserve, this would allow us to have 7 minutes for the IoT part, including the deployment of the extra components. Assuming that nothing more gets added by other components in the future.

Also, at the moment it takes ~42 minutes to validate a PR. Assuming that we have a second, or maybe a third job, we could parallelize more tests, and be quicker in validating PRs.

Altogether I think this is the right time to split off jobs and run them in parallel.

lulf:
I agree, the remaining time is a bit little. Hopefully we can improve the speed a bit in the future to allow running it in a single job so that we can avoid PR builds being blocked.

In my view then what remains for this  PR is my in-code comments and its good to go.

systemtests/src/test/java/io/enmasse/systemtest/iot/SimpleK8sDeployTest.java
famartinrh: I'm working in doing related IoT tests using api-model classes, I'll be creating IoTConfig in api-model soon, so I can also change this when the moment arrives

systemtests/scripts/run_test_kubernetes.sh
lulf: That makes sense, thanks

ctron:
I added an env-var to the script to explicitly deploy IoT components. But I will double check it that works in the travis build.

systemtests/src/test/java/io/enmasse/systemtest/iot/SimpleK8sDeployTest.java
lulf: Sounds good to me, thanks @famartinrh

PR #2447, opened 2019-03-14
IoTProject in api-model

famartinrh:
This PR adds more classes and builder annotations to IoTProject. The classes are inspired from go code of iot-operator.


famartinrh:
I think we can implement pojo classes for IoTConfig in a future PR, do you agree @ctron ??

api-model/src/main/java/io/enmasse/iot/model/v1/ProvidedDownstreamStrategy.java
```diff
@@ -0,0 +1,91 @@
...
+                prefix = "Doneable",
+                value = "done"
+                )
+        )
+//FIXME: implement missing fields and remove ignore annotation
+@JsonIgnoreProperties(ignoreUnknown = true)
+@JsonInclude(NON_NULL)
+public class ProvidedDownstreamStrategy {
+
+	private String namespace;
```
ctron: I think there is a formatting issue in this file. If you want an Eclipse formatting profile, then I could provide you with one.

famartinrh: I had set up some formatting settings manually, but probably I miss some, so please pass me your Eclipse formatting profile please

PR #2451, opened 2019-03-15
Drop "provided" strategy

ctron:
I would like to drop the provided strategy. At least for now. We don't really support it anyway.


lulf:
I think we've decided we want to provide provided (hah) strategy somehow, so closing this PR for now.

ctron:
Yes, agreed. We need to manage to provide the provided one alongside with the managed one :grin:

PR #2454, opened 2019-03-15
Support jvm options

lulf:


lulf:
@enmasse-ci run tests

PR #2457, opened 2019-03-18
Move reference of infra config from annotation to spec schema

lulf:
This allows the existence of the infraconfig field to be validated by OpenAPI schema
validation, as well as allowing the Operator Catalog UI to reference the
field.


lulf:
@enmasse-ci run tests testcase=plans.PlansTest

templates/crds/addressspaceplans.crd.yaml
```diff
@@ -32,6 +32,7 @@ spec:
             - addressSpaceType
             - resourceLimits 
             - addressPlans
+            - infraConfig
```
k-wall: Should this be `infraConfigRef`?

templates/crds/addressspaceplans.crd.yaml
```diff
@@ -43,6 +44,13 @@ spec:
               type: string
             addressSpaceType:
               type: string
+            infraConfigRef:
```
k-wall: Here we declare an object, but later, if the example plans for instance, it is used as if it were a string.  How does this work?

lulf: Fixed

kornys:
@lulf this will require update in systemtests as well, right?

lulf:
@kornys It will anyway fall back to the annotation, but I have updated the systemtest code after the merge of using api-model

PR #2459, opened 2019-03-18
Derive keycloak JVM memory settings from container memory

lulf:
This will only happen if JVM settings are not overriden explicitly.

### DIFF
--- a/pkg/controller/authenticationservice/standard.go
+++ b/pkg/controller/authenticationservice/standard.go
@@ -6,6 +6,7 @@ package authenticationservice
 
 import (
 	"context"
+	"fmt"
 	adminv1beta1 "github.com/enmasseproject/enmasse/pkg/apis/admin/v1beta1"
 	"github.com/enmasseproject/enmasse/pkg/util"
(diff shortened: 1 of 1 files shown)


lulf:

pkg/controller/authenticationservice/standard.go
```diff
@@ -111,6 +112,9 @@ func applyStandardAuthServiceDeployment(authservice *adminv1beta1.Authentication
 		jvmOptions := "-Dvertx.cacheDirBase=/tmp -Djboss.bind.address=0.0.0.0 -Djava.net.preferIPv4Stack=true -Duser.timezone=UTC"
 		if authservice.Spec.Standard.JvmOptions != nil {
 			jvmOptions += " " + *authservice.Spec.Standard.JvmOptions
+		} else if qty, ok := authservice.Spec.Standard.Resources.Requests["memory"]; ok {
+			maxHeap := qty.ScaledValue(resource.Mega)
```
k-wall: Poor variable name.  It is not (JVM) Heap, it is the container's memory.

lulf: Fixed

PR #2464, opened 2019-03-18
Api model iot config

famartinrh:
This PR adds IoTConfig classes to api-model, this is one of the last pieces to implement iot tests in systemtests


famartinrh:
@enmasse-ci run tests testcase=**.SimpleK8sDeployTest

api-model/src/main/java/io/enmasse/iot/model/v1/ImageOverride.java
```diff
@@ -0,0 +1,48 @@
+/*
+ * Copyright 2018, EnMasse authors.
```
ctron: That would be 2019 … unless you do time travel :wink: 

Same for all the others. Unless they are really from 2018, then it would be "2018-2019".

api-model/src/main/java/io/enmasse/iot/model/v1/ImageOverride.java
```diff
@@ -0,0 +1,48 @@
...
+        editableEnabled = false,
+        generateBuilderPackage = false,
+        builderPackage = "io.fabric8.kubernetes.api.builder",
+        inline = @Inline(
+                type = Doneable.class,
+                prefix = "Doneable",
+                value = "done"
+                )
+        )
+//FIXME: implement missing fields and remove ignore annotation
```
ctron: I think by the end of this PR, the FIXME should be gone. Same for all the others.

systemtests/src/test/java/io/enmasse/systemtest/iot/SimpleK8sDeployTest.java
```diff
@@ -47,15 +52,43 @@
...
+                .withServiceSecretNames(secrets)
+                .endSecretCertificatesStrategy()
+                .endInterServiceCertificates()
+                .endSpec()
+                .build();
+
+        configTempFile = File.createTempFile("iot-config", "json");
+        new ObjectMapper().writeValue(configTempFile, config);
+
+        KubeCMDClient.createFromFile(NAMESPACE, Paths.get(configTempFile.toURI()));
```
ctron: I think you should be able to use the kubernetes client, to create it from the Java world, instead of calling `kubectl`.

systemtests/src/test/java/io/enmasse/systemtest/iot/SimpleK8sDeployTest.java
```diff
@@ -47,15 +52,43 @@
...
     }
 
     @AfterAll
     protected static void cleanup() throws Exception {
+        if(configTempFile!=null && configTempFile.exists()) {
```
ctron: Same here, use the plain Java client to delete that resource.

api-model/src/test/java/io/enmasse/iot/model/v1/IoTConfigTest.java
```diff
@@ -0,0 +1,53 @@
...
+                config.getSpec().getInterServiceCertificates().getSecretCertificatesStrategy().getServiceSecretNames());
+        assertNotNull(config.getSpec().getImageOverrides());
+        ImageOverride image = config.getSpec().getImageOverrides().get("image1");
+        assertNotNull(image);
+        assertEquals("img", image.getName());
+        assertEquals("Always", image.getImagePullPolicy());
+        assertNull(config.getSpec().getEnableDefaultRoutes());
+    }
+
+}
```
ctron: I would like to see some parse testing here as well. So that we can drop in a few JSON files, and see that they still parse, after making modifications to the JSON model. I think you can already take the one file from the system tests as a starting point.

systemtests/src/test/java/io/enmasse/systemtest/iot/SimpleK8sDeployTest.java
famartinrh: in systemtests we use or command line clients or we use directly the api-server.
Do you mean using the api-server approach or an approach similar to this?
https://github.com/EnMasseProject/enmasse/blob/c4061432b33ad2945e98fb6d6b7c94b6d947835d/iot/iot-tenant-service/src/main/java/io/enmasse/iot/model/IoTProjects.java#L42-L43

kornys: @famartinrh so that's not true, ve use kubeCmdClient in only few operations (mainly for testing), calling api-server just for enmasse specific stuff, but Jens talking about using fabric8 kube client, in systemtests we use this client and we have interface for it Kubernetes.java, and since it is singleton you can get it everywhere by Kubernetes.getInstance()

api-model/src/test/java/io/enmasse/iot/model/v1/IoTConfigTest.java
famartinrh: added

PR #2465, opened 2019-03-18
Replace javascript authservice with a java one

lulf:
This removes some special handling of none-authservice in the makefiles,
and allowing us to build the whole project only with maven and go. The
implementation is supposed to be a drop-in replacement of the existing.


lulf:
@enmasse-ci run tests testcase=standard.auth.AuthenticationTest

lulf:
@enmasse-ci run tests testcase=standard.auth.AuthenticationTest

lulf:
@enmasse-ci run tests testcase=standard.auth.AuthenticationTest

PR #2471, opened 2019-03-19
Upgrade fixes

lulf:
@ctron Wdyt about not reassigning the label selector if it is already set? If you try to the label selectors, the Kubernetes API will complain about it being immutable, so this prevents an upgrade where labels change to work for instance.

@k-wall Various fixes to handle upgrades, deleting keycloak-controller, better error handling in address space controller

### DIFF
--- a/pkg/util/install/utils.go
+++ b/pkg/util/install/utils.go
@@ -37,7 +37,9 @@ func ApplyDefaultLabels(meta *v1.ObjectMeta, component string, name string) {
 func ApplyServiceDefaults(service *corev1.Service, component string, name string) {
 
 	ApplyDefaultLabels(&service.ObjectMeta, component, name)
-	service.Spec.Selector = CreateDefaultLabels(nil, component, name)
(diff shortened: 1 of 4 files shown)


pkg/util/install/utils.go
```diff
@@ -37,7 +37,9 @@ func ApplyDefaultLabels(meta *v1.ObjectMeta, component string, name string) {
 func ApplyServiceDefaults(service *corev1.Service, component string, name string) {
 
 	ApplyDefaultLabels(&service.ObjectMeta, component, name)
-	service.Spec.Selector = CreateDefaultLabels(nil, component, name)
+	if service.Spec.Selector == nil {
```
ctron: I think that is a good idea. From what I saw in other examples, the way to detect if a resource is newly created was:

~~~go
if existing.CreationTimestamp.IsZero() {
  // do new stuff
}
~~~

lulf: @ctron Thanks, updated PR with that.

PR #2475, opened 2019-03-19
Debug collecting router state

kornys:
@lulf I see stat standard job fail on timeout because collecting router state stuck without error message. So I will try to investigate that


kornys:
@enmasse-ci run tests testcase=standard.**

kornys:
@lulf so I checked log and after few tests run on pods stopped work, on qdrouters and also on api-server, so it can be caused by new version of fabric8 client.

kornys:
@enmasse-ci run tests testcase=standard.**

Issue #2408, opened 2019-03-05
Global Console - phase 1


k-wall:
This design proposal is moved to: #2477

**Overview**

Currently in EnMasse a management console is co-deployed with each addresspace.  This gives the user the ability to manage the contents of the addressspace, but there is no UI that gives the user a view of the whole system nor the ability to add or delete addressspaces.  This is a gap.

We desire a Global Console that has the ability to view all addressspace that exist cluster wide, gives the ability to create new addressspaces, as well as update and delete existing ones.  In the longer term, it is envisaged that this new console will incorporate the address managing capabilities provided by the existing Console (and other features), but the first phase the global console will simple link to the existing consoles provided by the addressspace.

In addition it is desired that the Global Console is extensible to manage other Custom Resources too.  The use-case here is to manage the Custom Resources that correspond to other messaging infrastructure for instance Apache Strimzi.    Th architecture will allow the view and the control of the Custom Resource to be plugged in.  This will allow the UI to present a multi-stage create workflow targeted to the Resource in question.

In the long term, the Global Console will also present key statistics about each managed resource. This will allow the user to understand a glance the state of the whole system.  The statistics being presented will vary depending on the resource in question.   Statistic won't be included in the first phase.

**Design**

There will be a Global Console UI served from a new Global Console Server component serving the static content.

Authentication will be delegated to OAUTH.  In OpenShift this will be automatically configured at deployment to use the built OAUTH infrastructure.  On Kubernettes, the user will need to integrate their Kubernettes with a OAUTH provider.   

The UI will use the Kubernettes API to list and control the resources that it needs to expose for Management.  This part of the UI will be extensible so that the ability to control other resources can be enabled.  In the first the phase the two resource types exposed for management will be EnMasse's standard addressspace and brokered addressspace.   The UI will send the Bearer token in order that the server side may authorise the requests. 

Initially the UI will periodically poll the API for the list of resources (standard addressspace and brokered addressspace).  Later this will evolve to utilise watch for a more responsive UI with fewer overheads.

There will be support for a multi page create workflow that will allow the user to create new instances of managed resources.  The HTML and JS supporting this workflow will be implementations of a pluggable API exposed by the Console itself.

It will be possible for the user to click through from the the managed resource to the existing Console of that resource.  To enable this, the resource will expose a console URL.  EnMasse's Console will be improved to allow back navigation.   Authentication will be seem-less as the existing OAUTH session will be valid for the old Console too.

[...]

PR #2477, opened 2019-03-19
Global Console Design - Phase 1

k-wall:
The design had outgrown #2408 and the discussion unwieldy. I have pulled everything together into a proposal.


documentation/design/proposals/global-console-phase1.adoc
```diff
@@ -0,0 +1,159 @@
...
+The Operator will use the information in the Console CR to create and keep synchronised the following resources.
+
+- oauthclient (OpenShift)
+- route (OpenShift only)
+- service (with serving-cert-secret-name on OpenShift).
+- serviceaccount
+- configmap instance 
+- deployment 
+
+On OpenShift, for the oauthclient, if the `oauthSecretRef` references an oauthclient instance that does not exist, the operator will create it. If it already exists, it will simply use it.  The Operator will ensure that the oauthclient's redirect urls are synchronized with the routes that exist to the service at that moment.
```
lulf: Implementation note: this can probably share the existing oauth client used by the per-addressspace console? This is created by the operator today

k-wall: Yes, that's my thinking.   In fact, I think they *have* to share the oauthclient otherwise there won't be a seamless SSO experience. The oauthclient will need to be updated each time an addressspace comes or goes with the redirect url of the address space's console route.

When the user clicks through from global console to an addressspace console, there will be a OAuth2 dosey doe with the authentication server but the user won't need to authenticate.

k-wall:
@lulf @vbusch updated to reflect the implementation as stands.  As I feel this design will need to change again for the next phase, I haven't polished the document.  I have just made the minimal changes.

documentation/design/console/global-console.adoc
```diff
@@ -0,0 +1,134 @@
...
+
+The Console is EnMasse comprises two separate web-applications that are hyperlinked together.
+
+* Global Console - that allows the viewing and management of address spaces.
+* Address Space Console - that allows the viewing and management of:
+ * addresses within an address space, and
+ * messaging connections made by applications within that address space.
+ * statistics
+
+In the long term, the desire is that current responsibilities of the Address Space Console will be migrated into the Global Console and the separate Address Space Consoles will be retired completed.
```
vbusch: Why wouldn't we plug in the Address Space Console, the same as we would for Strimzi?

documentation/design/console/global-console.adoc
```diff
@@ -0,0 +1,134 @@
...
+The Global Console is a Pattern Fly 4/React-JS application.  It is served by Oauth-proxy (https://github.com/openshift/oauth-proxy or https://github.com/pusher/oauth2_proxy).
+
+Authentication is delegated to the Oauth-proxy component.
+
+- For OpenShift, OpenShift Oauth-Proxy is configured to authenticate against the OAuth2 server of OpenShift itself.
+- For Kubernetes, Pusher Oauth-Proxy is configued with Open ID Connect (OIDC).  This needs to be the same OIDC instance that Kubernetes is configured to use itself.
+
+The UI will use the Kubernetes API to list and control the resources that it needs to expose for Management. This part of the UI will be extensible so that the ability to control other resources can be enabled. In the first the phase the two resource types exposed for management will be EnMasse's standard addressspace and brokered addressspace.
+
+Initially the UI will periodically poll the API for the list of resources (standard address space and brokered address space). Later this will evolve to utilise watch for a more responsive UI with fewer overheads.
```
vbusch: We will also need a console server in order to move to a watch.  
Currently in openshift, we are fetching the namespaces that the user has permissions to.  Then fetching the addressspaces for each namespace.  This would require a watch for each project + 1 for the project list.  Chrome has a limit of 30 websockets per host.
It would be best to use one websocket, which watches a console server.

documentation/design/console/global-console.adoc
```diff
@@ -0,0 +1,134 @@
...
+
+There is support for a multi page create workflow that will allow the user to create new instances of managed resources. The HTML and JS supporting this workflow will be implementations of a pluggable API exposed by the Console itself.
+
+It will be possible for the user to click through from the the managed resource to the existing Console of that resource. To enable this, the resource will expose a console URL. EnMasse's Console will be improved to allow back navigation. Authentication will be seem-less as the existing OAUTH session will be valid for the old Console too.
+
+The mast head of the application will present a summary of the number of resources of each type current visible to the user, cluster-wide. In the first version, it will present only the number of brokered and standard addressspaces, separately.
```
vbusch: It has been decided to remove this from the mast head.  Once we have more custom resources that we support, it might be added back as tabs.

documentation/design/console/global-console.adoc
k-wall: Agreed.

documentation/design/console/global-console.adoc
k-wall: I'd always imagined the address space console being subsumed into the global console.  I don't see the point of maintain a separate component.  I agree that for other components, a link to another console might be the way.  I think there will be time for discussion when we look at the next phase.

documentation/design/console/global-console.adoc
k-wall: Fixed.

PR #2481, opened 2019-03-20
Fix go make target so make build_go works

lulf:


k-wall:
The PR failed CI owing to this.  Should the make target be creating the directory?

for i in enmasse-controller-manager iot/qdr-proxy-configurator iot/iot-gc; do make -C $i build; done
make[1]: Entering directory `/home/travis/gopath/src/github.com/enmasseproject/enmasse/enmasse-controller-manager'
echo "" > "build/operatorImageMap.yaml"
/bin/sh: 1: cannot create build/operatorImageMap.yaml: Directory nonexistent
make[1]: *** [build/operatorImageMap.yaml] Error 2
make[1]: Leaving directory `/home/travis/gopath/src/github.com/enmasseproject/enmasse/enmasse-controller-manager'

lulf:
@k-wall Juuuuust updated the PR :)

lulf:
Ok, still not working, but i'm looking into it

k-wall:
I think there are still problems. `make clean` from the top level isn't cleaning the build directories.  If you manually deleted one, the build fails.

Makefile
```diff
@@ -1,14 +1,35 @@
...
+	mqtt-gateway \
+	mqtt-lwt \
+	service-broker \
+	iot/iot-tenant-service \
+	iot/iot-auth-service \
+	iot/iot-device-registry \
+	iot/iot-http-adapter \
+	iot/iot-mqtt-adapter \
+
+FULL_BUILD       = true
```
ctron: Shouldn't that be `?=` instead of `=`?

lulf: I don't think it makes sense to make it configurable.  FULL_BUILD is basically the way to prevent the per module Makefile targets to trigger when running make targets in the toplevel Makefile (its ugly i know). See Makefile.java.mk, where its used to prevent mvn clean install to run for instance

PR #2483, opened 2019-03-20
Redeploy selenium app if deployment fail

kornys:


kornys:

kornys:
@lulf this should avoid random fail results during tests with selenium

PR #2490, opened 2019-03-21
Add support for exporting endpoint info

### LINKED ISSUE #2446: Instruct url info to be put into configmap
**Overview**

When following a gitops-model, it is beneficial to get a deterministic URL for endpoints. Although we cannot control the service endpoint names in EnMasse (2 address spaces could end up creating conflicting ones), it would be useful if a tenant can instruct a configmap to be created with the endpoint information (url, ports, CA).

**Design**

The `AddressSpace` resource would be extended to support the following syntax:

```
apiVersion: enmasse.io/v1beta1
kind: AddressSpace
metadata:
  name: myspace
spec:
  endpoints:
  - name: messaging
    service: messaging
    exports/bindings:
      - kind: ConfigMap
        name: myconfig // Requires the 'system:serviceaccount:enmasse-infra:address-space-controller' service account RBAC rule to be created by application owner.
        namespace: mynamespace // Defaults to namespace of address space
      - kind: Secret
        name: mysecret
      - kind: Service
        name: myservice
```

The Secret/ConfigMap will contain the the following fields:

```
data:
  service.host: messaging-....svc
  service.port.amqp: 5672
  service.port.amqps: 5671
  // Following is only populated if endpoint is exposed
  external.host: ...
  external.port.amqps: 443
```

Service format:
```
kind: Service
...
spec:
  type: ExternalName
  # Points to the existing service in the enmasse-infra namespace
  externalName: messaging-....enmasse-infra.svc.cluster.local
  ports:
  - port: 5672
  - port: 5671
```

**Tasklist**

lulf:
This feature allows exporting endpoint information into a configmap,
service, or secret.

Fixes #2446


address-space-controller/src/main/java/io/enmasse/controller/ExportsController.java
```diff
@@ -0,0 +1,172 @@
...
+                .withType("ExternalName")
+                .withExternalName(endpointStatus.getServiceHost())
+                .withPorts(ServiceHelper.toServicePortList(endpointStatus.getServicePorts()))
+                .endSpec()
+                .build();
+
+        Service existing = client.services().inNamespace(addressSpace.getMetadata().getNamespace()).withName(name).get();
+        if (existing != null &&
+                (!endpointStatus.getServiceHost().equals(existing.getSpec().getExternalName())
+                        || !ServiceHelper.toServicePortList(endpointStatus.getServicePorts()).equals(existing.getSpec().getPorts()))) {
```
k-wall: Is a list comparison what we want here?  Shouldn't this be a set?

address-space-controller/src/main/java/io/enmasse/controller/ExportsController.java
```diff
@@ -0,0 +1,172 @@
...
+    @Override
+    public AddressSpace reconcile(AddressSpace addressSpace) {
+        Map<String, List<ExportSpec>> exportsMap = new HashMap<>();
+        for (EndpointSpec endpointSpec : addressSpace.getSpec().getEndpoints()) {
+            if (endpointSpec.getExports() != null) {
+                exportsMap.put(endpointSpec.getName(), endpointSpec.getExports());
+            }
+        }
+
+        for (EndpointStatus endpointStatus : addressSpace.getStatus().getEndpointStatuses()) {
```
k-wall: The issue talked about the export including the CA, but I don't see that in the implementation.

address-space-controller/src/main/java/io/enmasse/controller/ExportsController.java
```diff
@@ -0,0 +1,172 @@
...
+        }
+    }
+
+    private static Map<String, String> buildExportMap(EndpointStatus endpointStatus) {
+        Map<String, String> map = new HashMap<>();
+        map.put("service.host", endpointStatus.getServiceHost());
+        for (Map.Entry<String, Integer> portEntry : endpointStatus.getServicePorts().entrySet()) {
+            map.put("service.port." + portEntry.getKey(), String.valueOf(portEntry.getValue()));
+        }
+        map.put("external.host", endpointStatus.getExternalHost());
```
k-wall: Can't external host be null? If so, should we omit from the map?

address-space-controller/src/main/java/io/enmasse/controller/ExportsController.java
lulf: Fixed

address-space-controller/src/main/java/io/enmasse/controller/ExportsController.java
lulf: Fixed.

address-space-controller/src/main/java/io/enmasse/controller/ExportsController.java
lulf: Done.

k-wall:
@lulf  what do I actually need to do to enable to use a configmap expose?  I see // Requires the 'system:serviceaccount:enmasse-infra:address-space-controller' service account RBAC rule to be created by application owner in the issue and I see `"io.fabric8.kubernetes.client.KubernetesClientException: Failure executing: POST at: https://172.30.0.1/api/v1/namespaces/enmasse-infra/configmaps. Message: Forbidden!Configured service account doesn't have access. Service account may have been revoked. configmaps "myexpot" is forbidden: cannot set blockOwnerDeletion if an ownerReference refers to a resource you can't set finalizers on: no RBAC policy matched, <nil>."` if I do nothing.  What actually does a user need to do?

lulf:
@k-wall I was able to export using the following resources:

```
kind: Deployment
apiVersion: apps/v1
metadata:
  name: myapp
spec:
  replicas: 1
  selector:
    matchLabels:
      name: myapp
  template:
    metadata:
      labels:
        name: myapp
    spec:
      containers:
      - name: image
        image: busybox
        env:
        - name: HOST
          valueFrom:
            configMapKeyRef:
              name: myconfig
              key: service.host
        command:
        - sleep
        args:
          - "3600"

kind: RoleBinding
metadata:
  name: rbac-binding
roleRef:
  kind: Role
  name: rbac
subjects:
- kind: ServiceAccount
  name: address-space-controller
  namespace: enmasse-infra

kind: Role
metadata:
  name: rbac
rules:
  - apiGroups: [ "" ]
    resources: [ "configmaps" ]
    verbs: [ "get", "create", "update", "list", "watch" ]

kind: AddressSpace
metadata:
    name: space1
spec:
    type: standard
    plan: standard-small
    endpoints:
      - name: messaging
        service: messaging
        exports:
          - kind: ConfigMap
            name: myconfig
      - name: mqtt
        service: mqtt
      - name: console
        service: console
```

PR #2492, opened 2019-03-21
Add iot config to enmasse operator

lulf:


lulf:
Looking into the test failure...

lulf:
@ctron So, it had to add a small fix to the run_test_kubernetes.sh script so that it actually replaced the enmasse-infra namespace in for the iot resources. With that, the iot-operator is running, and the iotconfig is created, but it doesn't look like anything gets created. 

The enmasse-operator is trying to reconcile the IotConfig but it prints:

"Config was not found"

PR #2496, opened 2019-03-22
enforce python2 to avoid errors when default is python3

jbtrystram:
When python3 is set as default python executable the build fails because this script use some py2 syntax.
This ensure the script is ran with the appropriate python version.


lulf:

lulf:

PR #2497, opened 2019-03-22
Update 0.27.0 changelog

lulf:


k-wall:
I think I would call out the Artemis and Keycloak upgrades.   I would also mention PATCH support.

lulf:
Agreed. It’s a bit hard to work through all the commits to remember what’s changed. We should get better at using milestones for issues to track that

PR #2501, opened 2019-03-22
Fix issue where route was created by when endpoint was not exposed

lulf:
Remove special constructors for EndpointSpec. This removes compatibilty
with the old `servicePort` field, that has been deprecated for several
releases (and I'm not sure anyone really uses).

This fixes a case where the route would get created even if the endpoint
was not exposed.


lulf:
@enmasse-ci run tests

PR #2502, opened 2019-03-23
Update router image to 1.6.0

lulf:
Remove router-metrics component as router now has native support for
prometheus. Setup localhost http listener for retrieving prometheus
metrics.


lulf:
@enmasse-ci run tests

lulf:
@enmasse-ci run tests

Makefile
```diff
@@ -17,7 +17,6 @@ DOCKER_DIRS = \
 	standard-controller \
 	keycloak-plugin \
 	router \
```
k-wall: Yippee!

lulf:
@enmasse-ci run tests

address-space-controller/src/main/resources/templates/standard-space-infra.yaml
```diff
@@ -101,6 +101,14 @@ objects:
         linkCapacity: ${ROUTER_LINK_CAPACITY}
       }
 
+      listener {
+        host: 127.0.0.1
+        port: 8080
+        authenticatePeer: no
+        metrics: true
+        http: true
```
grs: You can explicitly disable websockets here (websockets: false) and the console (httpRootDir: invalid) so that only metrics can be accessed. As you are only exposing this over localhost that may not be a big concern (but how is promethues tehn able to access it?).

Note also on a different point, if desired there is also now an http based healthz endpoint (automatically enabled but can be disabled (healthz: false). This avoids the noise in the logs caused by a socket to the amqp ports that never sends the amqp protocol header.

lulf: Ah, thanks! Yeah I forgot prometheus needs to scrape these externally, so we need to expose it. I will update these settings along with the new healthz endpoint.

PR #2507, opened 2019-03-24
Fix creating tck resources

kornys:
@lulf this PR fixes problem with creating resources for tck tests and also removes categorize_docker_logs which caused the mv logging errors on the end of pipelines


kornys:
@enmasse-ci run tests testcase=brokered.SmokeTest profile=systemtests-smoke

kornys:
@enmasse-ci run tests testcase=standard.SmokeTest profile=systemtests-smoke

systemtests/scripts/run_test_kubernetes.sh
```diff
@@ -106,7 +106,6 @@ if [[ -z "$DISABLE_LOG_SYNC" ]]; then
     #stop docker logging
     echo "process for syncing docker logs with PID: ${LOGS_PID} will be killed"
     kill ${LOGS_PID} || true
-    categorize_docker_logs "${DOCKER_LOG_DIR}" || true
```
lulf: Is this not working anymore?

systemtests/scripts/test_func.sh
```diff
@@ -125,11 +125,19 @@ function teardown_test() {
...
+apiVersion: enmasse.io/v1beta1
+kind: AddressSpace
+metadata:
+    name: ${ADDRESS_SPACE_NAME}
+    namespace: ${NAMESPACE}
+spec:
+    type: brokered
+    plan: brokered-single-broker
+    authenticationService:
+        type: standard
```
lulf: No need to set type when name is set.

systemtests/scripts/run_test_kubernetes.sh
kornys: @lulf ut was set implemented in past for categorize logs based on address space (namespace), but we changed deployment to one infra project so address space name is not in pod name so it does not work

PR #2516, opened 2019-03-25
add script to delete hanging openshift project

dejanb:


scripts/delete_openshift_project.sh
```diff
@@ -0,0 +1,12 @@
...
+# Deletes Openshift project stuck in the finalize state
+# Usage: delete_openshift_project enmasse-infra ec2-x-x-x-x.zone.compute.amazonaws.com:8443
+
+TOKEN=$(oc whoami -t)
+# This doesn't work with ec2-x-x-x-x.zone.compute.amazonaws.com format of host names
+#ENDPOINT=$(oc config current-context | cut -d/ -f2 | tr - .)
+ENDPOINT="$2"
+PROJECT="$1"
+
+kubectl get namespace "$PROJECT" -o json | jq 'del(.spec.finalizers[] | select("kubernetes"))' | curl --insecure -k -H "Content-Type: application/json" -H "Authorization: Bearer $TOKEN" -X PUT --data-binary @- https://$ENDPOINT/api/v1/namespaces/$PROJECT/finalize
```
kornys: on server where is openshift installed this will fail because kubectl is not installed

dejanb: `kubectl` is available in the same dir as `oc` in the download archive, so it should work.I'm not sure if it work with minishift though. Thought it'd be useful as a util script, if not I can put it somewhere else.

kornys: don't understand which download archive you mean?

lulf:
Whats the intended use? Probably not for systemtests? 

If it’s a convenience script because EnMasse is not properly uninstalled causing project delete to hang, that’s a bug we should fix. 

How can we reproduce this @dejanb?

scripts/delete_openshift_project.sh
dejanb: I meant on Openshift client tools download

dejanb:
@lulf No, it's not for the system tests, just a small util script to clean things up when things go wrong. I had it like that yesterday. Will see to find more details next time.

lulf:
@dejanb ok, thanks. Since this is strictly not an enmasse-related script I don't think it should live in the repo. And if there is an actual issue with enmasse project not getting deleted/hanging, that is a serious issue for some users that I'd like to understand more of.

If you encounter the issue, please record the steps you took before trying to delete. The 'tested' process is documented here: https://enmasse.io/documentation/master/openshift/#uninstalling-messaging

dejanb:
No worries, just thought it might be useful for others in such situations

ctron:
Hey @dejanb  … https://github.com/ctron/kill-kube-ns … I made this a while back based on kubernetes/kubernetes#60807 … Looks like this works for most people.

dejanb:
@ctron yeah, that's what I started with and just adapted it to work with Openshift (use token and endpoint if possible). We can modify that original script and incorporate some of this changes

PR #2520, opened 2019-03-25
Split docker build from default target

lulf:
* Require docker_build to be explicitly invoked
* Use podman on travis and parallelize build

This avoids a deadlock where Golang docker images depended on java build
to complete, and java build depended on golang build to complete. Make
docker_build a standalone target also allows for parallelizing
docker_build invocations, which should work well with non-daemon
container tools such as podman

### DIFF
--- a/.jenkins/Jenkinsfile
+++ b/.jenkins/Jenkinsfile
@@ -71,6 +71,13 @@ pipeline {
                 }
             }
         }
+        stage('build docker image') {
+            steps {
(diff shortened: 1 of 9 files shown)


lulf:
@enmasse-ci run tests

kornys:
@lulf what about rename docker_build to container_build?

lulf:
@kornys, yes but for a later PR

kornys:
Ok

PR #2528, opened 2019-03-26
changes to IoTProject managed strategy

famartinrh:


ctron:
When you rebase on master, then the test failure should be fixed.

famartinrh:
This changes will be added in PR #2517

PR #2532, opened 2019-03-26
Perform liveness check on statically configured listener

lulf:
This makes sure the router is not killed if it is not configured within
liveness probe interval. Only readiness check should check that
router is configured. This bug was introduced when switching to http
based probes.


lulf:
@enmasse-ci run tests testcase=standard.SmokeTest

PR #2536, opened 2019-03-27
Allow configuring idle timeout and thread settings

lulf:
Broker and router settings related to AMQP heart beat and threads used
to handle connections can now be tuned in the standard infra config


lulf:
@enmasse-ci run tests

PR #2537, opened 2019-03-27
Fix IoT system tests

ctron:
So it turns out the system tests should have have worked. However the failure to detect "non-ready" containers in the `TestUtils` reported "OK" for the test, although the test was not ok.

The cause for the test to fail was that the hostnames in the plain k8s inter service certificates where generated with `enmasse-infra` instead of `enmasseci`, which is required for running on travis.

### DIFF
--- a/systemtests/scripts/run_test_kubernetes.sh
+++ b/systemtests/scripts/run_test_kubernetes.sh
@@ -8,6 +8,8 @@ TEST_PROFILE=$2
 TESTCASE=$3
 failure=0
 
+: ${KUBE_OPERATION:=create}
+
 BASE_DIR="${CURDIR}/../../"
 API_URL=$(kubectl config view --minify | grep server | cut -f 2- -d ":" | tr -d " ")
 API_TOKEN=$(kubectl describe secret $(kubectl get serviceaccount default -o jsonpath='{.secrets[0].name}') | grep -E '^token' | cut -f2 -d':' | tr -d " ")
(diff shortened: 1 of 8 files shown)


systemtests/scripts/run_test_kubernetes.sh
```diff
@@ -32,10 +34,16 @@ mkdir -p api-server-cert/
 openssl req -new -x509 -batch -nodes -days 11000 -subj "/O=io.enmasse/CN=api-server.${KUBERNETES_NAMESPACE}.svc.cluster.local" -out api-server-cert/tls.crt -keyout api-server-cert/tls.key
 kubectl create secret tls api-server-cert --cert=api-server-cert/tls.crt --key=api-server-cert/tls.key
 
+if [ "$ONLY_PULL_ONCE" = "true" ]; then
+	# disable loading images always when running on travis
+	# we don't get refresh images, so we can speed this up
+	sed -i "s/imagePullPolicy: Always/imagePullPolicy: IfNotPresent/g" ${ENMASSE_DIR}/install/*/*/*.yaml
```
lulf: Perhaps instead set  IMAGE_PULL_POLICY=IfNotPresent .travis.yml

ctron: Will that override the makefile?

ctron: Ok, I changes this to "common.sh" when it is a PR

PR #2540, opened 2019-03-28
Remove mqtt adapter tls setting from example config

dejanb:


ctron:
You actually need this, as the MQTT endpoint requires a dedicated certificate. You can't use the routes, but you also shouldn't use the internal service certificate for external services either.

But it is indeed missing in the current documentation.

lulf:
Close?

ctron:
> Close?

I think so, @dejanb what do you think?

PR #2545, opened 2019-03-29
Allow installing iot modules from ansible

lulf:


kornys:
@lulf should be in systemtests/ansible/inventory as well, WDYT?

lulf:
@kornys default is false, so no need to set it. I don't think we should install it by default as we want to test without it as well

kornys:
@lulf but from my POV I think it should be there because if I will test it manualy I can easily change that var in inventory

PR #2546, opened 2019-03-29
Prefix qdr-proxy-configurator with "iot-"

ctron:


ctron:
Yes, that may be an idea … I have to rebase the PR anyway.

PR #2547, opened 2019-03-29
Embed the hono-legal package into the fat jar

ctron:


ctron:
@enmasse-ci run tests testcase=standard.SmokeTest profile=systemtests-smoke

ctron:
We don't need the actual legal content. Maven has all the dependencies already.

PR #2557, opened 2019-03-30
Make api-server startup handle case if request fails and assume Kuber…

lulf:
…netes instead.

This is causing it to fail the test for inclusion in operator hub: https://github.com/operator-framework/community-operators/pull/234

### DIFF
--- a/standard-controller/src/main/java/io/enmasse/controller/standard/StandardController.java
+++ b/standard-controller/src/main/java/io/enmasse/controller/standard/StandardController.java
@@ -7,6 +7,8 @@ package io.enmasse.controller.standard;
 import io.enmasse.k8s.api.*;
 import io.enmasse.metrics.api.Metrics;
 import io.enmasse.model.CustomResourceDefinitions;
+import io.fabric8.kubernetes.client.DefaultKubernetesClient;
+import io.fabric8.kubernetes.client.NamespacedKubernetesClient;
 import io.fabric8.openshift.client.DefaultOpenShiftClient;
 import io.fabric8.openshift.client.NamespacedOpenShiftClient;
 import io.vertx.core.Vertx;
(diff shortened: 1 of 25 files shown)


SamiSousa:
This may not be a suitable workaround, since the scorecard proxy could be running on OpenShift as well. Maybe there is another way to check if the distribution is OpenShift, or perhaps this can be parameterized with a flag (eg. `--openshift=true`)

lulf:
@SamiSousa it needs to be checked at the runtime since want to avoid duplicating yaml for kubernetes and openshift. However, assuming it is failing due to starting up earlier than the scorecard, I think perhaps a better fix is to repeat the check N times and with a timeout if it gets connection refused

lulf:

lulf:
So... long story short is that there was another automatic API call made against OpenShift for OAUTH, which shouldn't really be necessary. I've updated this PR to change how we default to only adapt to OpenShift client when necessary. This fixes the travis build in https://github.com/operator-framework/community-operators/pull/234 .

lulf:

lulf:

lulf:
@enmasse-ci run tests profile=systemtests-smoke

lulf:
@enmasse-ci run tests profile=systemtests-smoke testcase=SmokeTest

lulf:
@enmasse-ci run tests profile=systemtests-smoke testcase=SmokeTest

lulf:
@enmasse-ci run tests

standard-controller/src/main/java/io/enmasse/controller/standard/StandardController.java
```diff
@@ -124,16 +126,16 @@ public void stop() {
...
             } finally {
-                openShiftClient.close();
+                kubeClient.close();
                 log.info("StandardController stopped");
             }
         }
     }
 
-    private static boolean isOpenShift(NamespacedOpenShiftClient client) {
+    private static boolean isOpenShift(NamespacedKubernetesClient client) {
```
k-wall: There's a version of this method in src/main/java/io/enmasse/api/server/ApiServer.java that this PR changes into a retry loop.  If it is important to retry there, why not here too?

lulf: I will remove the retry from the api-server because it is no longer needed

lulf: I removed this check altogether, as the openshift-specific behavior is never used in standard-controller

PR #2563, opened 2019-04-01
Improve IoT adapter password changing logic

dejanb:
Currently the logic sets a new password on every iteration until IoT project becomes ready

### DIFF
--- a/pkg/controller/iotproject/iotproject_controller.go
+++ b/pkg/controller/iotproject/iotproject_controller.go
@@ -142,7 +142,6 @@ func (r *ReconcileIoTProject) updateProjectStatusError(ctx context.Context, requ
 
 	newProject := project.DeepCopy()
 	newProject.Status.IsReady = false
-	newProject.Status.DownstreamEndpoint = nil
 
 	return r.client.Update(ctx, newProject)
 }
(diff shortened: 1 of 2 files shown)


ctron:
I am not sure this works … because "IsReady" will be updated after this function was running …

ctron:
… but looking at the code, I think the credentials should initialize the password the with one from the status section, instead of leaving it out:

https://github.com/EnMasseProject/enmasse/blob/f8665ce243b235a70e504cbc3a38143ca08aef8b/pkg/controller/iotproject/managed.go#L84-L86

I that would be copied from the status section, then it would only be set if the status section was not written. However that also means that the password may only be assigned to the status section if the update was successful. So it needs to be moved out of `reconcileAdapterMessagingUser`.

ctron:
We discussed this in detail during a call. The solution from @dejanb works, and is pretty close to what we want. We only figured out there is a remaining loophole when the project becomes "unready" and the recovers.

In order to properly we fix this, we decided to always store the endpoint information, if the IoT project is ready or not. This will make it much easier fixing the cause of this.

dejanb:
@ctron Implemented what we discussed and simplified a code a bit. It works as expected on my end

PR #2564, opened 2019-04-01
Dont change persistence settings of standard authservice

lulf:
Neither data source or persistence of standard authservice will be
changed even if custom resource is changed, in case applied by mistake. Changing these settings are not supported and should be prevented by a validation admission webhook in the future.


lulf:

PR #2566, opened 2019-04-01
Postgresql fix

lulf:
With these changes I'm able to use the postgresql datasource @kornys


lulf:

PR #2571, opened 2019-04-02
Replace envsubst with sed

lulf:


lulf:
@enmasse-ci run tests

k-wall:
Out of curiosity, why replace envsubst?

vbusch:
> Out of curiosity, why replace envsubst?

@k-wall might be that sed is usually available whereas envsubst requires you to install gettext

PR #2580, opened 2019-04-04
Add support for specifying router policy

lulf:
* Allows configuring router policy


lulf:

lulf:
@enmasse-ci run tests testcase=standard.infra.InfraTest

k-wall:
Does this feature have a way to differentiate between user connections (i.e. connections from a user's applications) and connections made internally within EnMasse (e.g. connections from Brokers, MQTT components).   My expectation is that these limits should be applied to the former.

grs:
As specified this policy is applied to all connections. There appears to be no way to associate different policy with different listeners. To apply different policy to internal connections there would need to be a different vhost that they all used or a distinct user group defined.

k-wall:
From the perspective of a user of EnMasse, I would expect the policy controls exposed by EnMasse to be applied to my application traffic.   As a user I should not need to factor in the internal overheads of EnMasse itself.

lulf:
@grs So for internal connections we have 2 options I guess:

1. Change all components to specify an 'internal' vhost. However I guess we cannot enforce external applications to not use this vhost
2. Create a user group defined for internal components. How is the client identified when using TLS client certs? 

I think I would prefer option 2 if we can get that to work.

grs:
If using SASL EXTERNAL, the certificate's subject is used as the identity for the connection.

lulf:
@enmasse-ci run tests testcase=standard.infra.InfraTest

lulf:
@k-wall I think we should go with this scheme for now, and improve it once the router supports per-listener policies.

lulf:
@enmasse-ci run tests testcase=standard.infra.InfraTest

PR #2581, opened 2019-04-04
Global console

k-wall:


k-wall:
@lulf This is still a work in progress with a list of todos in a comment at the top of consoleservice_controller.go.   If you could give feedback on , the CRD, and the touch points with the AdressSpaceController, that would be very helpful.

kornys:
@k-wall is this PR ready for initial testing?

templates/enmasse-operator/050-Deployment-enmasse-operator.yaml
```diff
@@ -73,6 +73,8 @@ spec:
           value: ${KEYCLOAK_IMAGE}
         - name: KEYCLOAK_PLUGIN_IMAGE
           value: ${KEYCLOAK_PLUGIN_IMAGE}
```
lulf: You need to add all the console images as envs to the operator so that it can be overridden from env

templates/enmasse-operator/050-Deployment-enmasse-operator.yaml
```diff
@@ -73,6 +73,8 @@ spec:
...
+        - name: CONTROLLER_ENABLE_CONSOLE_SERVICE
+          value: "true"
```
lulf: What happens if you set this to false? Will the per-addressspace console still function?

lulf:
@k-wall I think the CRDs, address-space-controller parts looked fine.

templates/enmasse-operator/050-Deployment-enmasse-operator.yaml
k-wall: No, the addresspace console will require OAuth too.   I think it would be best if on OpenShift, the route to the console didn't get created if the console service is absent.

k-wall:
@enmasse-ci run tests

k-wall:
@enmasse-ci run tests

k-wall:
@enmasse-ci run tests

k-wall:
@enmasse-ci run tests

k-wall:
@enmasse-ci run tests

k-wall:
@enmasse-ci run tests

k-wall:
@enmasse-ci run tests testcase=**.SessionTest profile=systemtests

lulf:
@vbusch @k-wall  A couple of notes on the UI:

* The address space plans should be ordered by .spec.displayOrder. 
* When creating the address space, the word 'Unavailable' glares out to me. I wonder if instead we should adopt a 'phase' field as for the Address to provide a better message. I.e. the following state transitions: 'Provisioning' -> 'Active' <-> 'Failed'

kornys:

kornys:
@enmasse-ci run tests testcase=common.console.**

PR #2597, opened 2019-04-08
Fix check image upgraded

kornys:


kornys:

PR #2602, opened 2019-04-09
Use new Java base image for IoT

ctron:


ctron:
Updated the images to use the new base image

PR #2607, opened 2019-04-09
Make create cert scripts work on OSX

dejanb:


dejanb:
Can we remove this last sed as "-z" option is non-standard and it's not doing much functionally?

PR #2611, opened 2019-04-10
Test IoTProject creation as part of smoketests

ctron:
I think it makes sense to always test if an IoTProject can be created.


ctron:

ctron:
> LGTM, is the test failing for valid reasons?

No, it should not fail. I am trying to figure out what goes wrong here.

And I also think that the cleanup logic doesn't work as it should. As deleting the IoTProject will delete the AddressSpace and therefore the exception later on occurs if I interpret the logs correctly.

lulf:
@ctron closable?

PR #2613, opened 2019-04-10
fix iot guide with new template files path

jbtrystram:
Files were moved recently but the guide was not updated

### DIFF
--- a/iot/examples/README.md
+++ b/iot/examples/README.md
@@ -10,14 +10,14 @@
 * Install EnMasse with IoT services
 
   ```
-  oc apply -f templates/build/enmasse-latest/install/bundles/enmasse
+  oc apply -f install/bundles/enmasse
   
-  oc apply -f templates/build/enmasse-latest/install/components/example-authservices/standard-authservice.yaml
-  oc apply -f templates/build/enmasse-latest/install/components/example-roles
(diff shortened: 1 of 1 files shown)


iot/examples/README.md
```diff
@@ -17,7 +17,7 @@
   oc apply -f templates/build/enmasse-latest/install/components/example-plans
   
   oc apply -f templates/build/enmasse-latest/install/preview-bundles/iot
-  oc apply -f iot/examples/iot-config.yaml
+  oc apply -f templates/iot/examples/iot-config.yaml
```
dejanb: The patch should be "install/components/iot/examples/" for all these resources

jbtrystram: why was it moved there then ? 
https://github.com/EnMasseProject/enmasse/blob/master/templates/iot/examples/iot-config.yaml

jbtrystram:
@dejanb WDT ?

dejanb:
@jbtrystram you don't need "templates/build/enmasse-latest/", you should execute commands from that folder (as that's how release will look like and what doc should cover)

iot/examples/README.md
dejanb: That's the source of templates ...

PR #2616, opened 2019-04-10
Sort plans by display order

lulf:


lulf:
Good point @vbusch. Somehow it feels strange to add the displayOrder to this schema, but perhaps instead the API can return the plans in the displayOrder, and the UI would not do any sorting at all?

lulf:
@vbusch I've modified the PR to do the sorting server side based on displayOrder. I have also added another field that exists already in the plans, `displayName`, which I think could be useful for the UI (haven't change anything there as it would require keeping track of both name and displayName in the client I think).

console/console-init/ui/src/components/MessagingInstances/MessagingInstance/Enmasse/EnmasseAddressSpaces.js
```diff
@@ -23,7 +23,7 @@ const translateAddressSpaces = addressSpaces => {
     return translation;
 }
 
-const getPlanNames = plans => plans.spec.plans.map(plan => plan.name).sort();
+const getPlanNames = plans => plans.spec.plans.map(plan => plan.name);
```
vbusch: This could get the displayName now.

lulf: When creating the address space, you still need to input the name and not the displayName in the .spec.plan field, so one would have to remember both.

PR #2623, opened 2019-04-11
Add external IoT services for Kubernetes deployment

dejanb:


dejanb:
I think in the long run we should use ExposeSpec to align better with EnMasse services. But this should do for the time being

PR #2628, opened 2019-04-11
Make phase forward compatible

lulf:
* Allows downgrade to previous version
* Allow unknown fields to all model objects in order to allow future
extensions without breaking forward compatiblity.


lulf:

PR #2630, opened 2019-04-11
seamless sso between global console and address space consoles

k-wall:
This demonstrates the sort of change that would be required to have a seamless SSO experience between the new console and the existing ones.

This currently this POC PR hardcodes a key used to cipher the access token - ultimately secret this would need to be shared between all console and admin/agent instances. Key rotation is also an issue to be resolved.

For the cookie sharing to work, either all consoles would need to be served from the same subdomain with the cookie scoped to that subdomain), or all consoles would need to be served from the same host (with addresspace consoles given unique paths). Currently this PR hardcodes .nip.io as a cookie domain.


k-wall:
For a solution utilising a fixed cookie key, minimum would be:

1. Add cookie-key secret ref to console service CRD.  Have operator create the secret containing a random key.
2. Add cookie-domain to CRD
3. Inject cookie-key and cookie-domain to oauth-proxy configuration.
4. Have agent read the cookie-key secret.  Use this key to decrypt cookie.

To allow key rotation:

List above + the following tasks:

1. Have operator periodically refresh the cookie secret and trigger a rolling restart of the global-console.
2. Agent would need to watch the cookie secret so it knows when the secret changes. It could try cookie secret and its immediate predecessor to avoid the race of someone authenticating whilst the key is changed.

To allow to global console and address space console to use the same host.  (Avoids the need for global console and address space cosonles to share the same domain).

For this, the address space console would need to switch to using path base routed.  The global console would be served from `/` and the address space consoles from a path (such as `/addressspace/<addressspacename>`).

1. Refactor the address space controller to change the route.
2. There a number of places with console server where it assumes the application is hosted at `/`.  These would need to be refactored.

k-wall:
This change allows the user to configure SSO between global console and address space console.  To enable the feature, the user need to configure the `ssoCookieDomain` so that the cookie will be visible to the address space consoles too.  At which point SSO will be enabled.  If testing with `nip.ip`, set the cookie domain to `.nip.ip`.

lulf:
@enmasse-ci run tests

k-wall:
We could autoenable the computation of `ssoCookieDomain. This would give a slightly nicer out of the box feel.  The operator could calculate the common subdomain suffix of all the console routes and set ssoCookieDomain accordingly.  I would suggest that the algorithm should refuse to set a cookie domain that corresponded to a bare TLD.  This would work reasonably for a `company.com`, or in a development situation `nip.io`.  The disadvantage is that for domain names formed with a country code at the TLD position, it could enable sharing of a cookie across say, across, `co.uk`.   If we implemented this, I would suggest adding a annotation as a switch to turn the feature off.

k-wall:
We could autoenable the computation of `ssoCookieDomain` which would give a slightly nicer out of the box feel.  The operator could calculate the subdomain suffix of all the console routes and set ssoCookieDomain.  The algorithm would refused to set a cookie domain that corresponded to a TLD This would work reasonably for a company.com, or in a development situation nip.io.  It would give less desirable results for domains such a co.uk

lulf:
@enmasse-ci run tests

k-wall:
@enmasse-ci run tests testcase=**ConsoleTest

k-wall:
@enmasse-ci run tests testcase=ServiceCatalogWebTest

k-wall:
@lulf Can you re-review this PR please?  The ability to auto set `ssoCookieDomain` is added, unless disabled by setting the `"enmasse.io/disable-autocompute-sso-cookie-domain` to true.

The GlobalConsoleTest are written is such a way that they don't need to be changed.  They pass for me.  ServiceCatalogWebTest is running now.

PR #2632, opened 2019-04-11
Allow using schema if some parts are missing

lulf:
This makes it less error prone to define invalid plans. For instance, authentication services will still be loaded if you have invalid plans, and plans that are valid will still be available.


lulf:
@enmasse-ci run tests

lulf:
Fixes #2361

PR #2633, opened 2019-04-12
Don't fail deleting an item that is already gone

ctron:
Currently when the configurator tries do delete an item, it may happen that the item is not there. In this case it currently fails to resync the configuration, as deleting returns "not found" as an error.


ctron:

ctron:

PR #2648, opened 2019-04-14
Fixes for console tests

kornys:
fixes plans tests and webconsoletests


lulf:
@kornys Not sure if its the correct fix to assume WebDriverException means unauthorized :) But if you think its ok we can merge it

kornys:
@lulf I think it is fine for now :) anyway these tests will be deleted when new console will be completed :)

PR #2672, opened 2019-04-16
Added link in about box to the docs

vbusch:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -47,7 +47,9 @@ endif
 ifneq ($(strip $(GLOBAL_CONSOLE_DISPLAY_NAME)),)
         MAVEN_ARGS+="-Dapplication.globalconsole.display.name=$(GLOBAL_CONSOLE_DISPLAY_NAME)"
 endif
-
+ifneq ($(strip $(PROJECT_DOCS_URL)),)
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -47,7 +47,9 @@ endif
 ifneq ($(strip $(GLOBAL_CONSOLE_DISPLAY_NAME)),)
         MAVEN_ARGS+="-Dapplication.globalconsole.display.name=$(GLOBAL_CONSOLE_DISPLAY_NAME)"
 endif
-
+ifneq ($(strip $(PROJECT_DOCS_URL)),)
```
k-wall: I don't think we want to reply on arguments to the build to customise labels.  Better to include these in pom.properties.   That way derivative versions can specify their own values and subject them to source control.

pom.xml
```diff
@@ -983,7 +983,11 @@
 					    <name>applicationGlobalConsoleDisplayName</name>
 					    <value>${application.globalconsole.display.name}</value>
 					</parameter>
-                                        <parameter>
+					<parameter>
```
k-wall: I'm not sure why you pass this parameter here.  This plugin is concerned with the production of the license.html file.

k-wall:
This change isn't working for me.  The about box contains nothing more than the EnMasse Version line.

PR #2682, opened 2019-04-16
Ensure old authentication services are removed

lulf:
* Ensure old authentication services are removed from the cached map
* Unify implementation of Keycloak interaction in KeycloakUserApi and RealmController


k-wall:
qq.  What use-case cause there to be old authentication services to be removed?  Is this an upgrade scenario?

lulf:
@k-wall its for the case where you

* create authservice a
* create and delete address spaces using it
* remove authservice a 
* create authservice b
* create and delete address spaces

Problem is that 'a' will still have a handle in api-server for instance, so when you query for all messaging users, it will try to contact it if it is not removed. The other bug was that it could be null if it was queried before it was up, which would end up putting a null in the keycloakMap. This is also fixed in this PR.

k8s-api/src/main/java/io/enmasse/k8s/api/CachingSchemaProvider.java
```diff
@@ -8,23 +8,32 @@
...
     public Schema getSchema() {
         return schema;
     }
 
+    public void registerListener(SchemaListener listener) {
+        listeners.add(listener);
+    }
+
     @Override
     public void onUpdate(List<Schema> items) {
```
k-wall: This code doesn't read well. It seems arbitrary to ignore all but the first entry in the list.

keycloak-user-api/src/main/java/io/enmasse/user/keycloak/KeycloakUserApi.java
```diff
@@ -47,16 +48,24 @@
...
-        this(keycloakFactory, clock, Duration.ZERO);
-    }
-
     public KeycloakUserApi(KeycloakFactory keycloakFactory, Clock clock, Duration apiTimeout) {
         this.keycloakFactory = keycloakFactory;
         this.clock = clock;
         this.apiTimeout = apiTimeout;
     }
 
+    public synchronized void pruneAuthenticationServices(List<AuthenticationService> items) {
```
k-wall: I found the API less than obvious.  On my first reading of the use, I was expecting the items list to be list that is being pruned rather than those that are to be retained.

k8s-api/src/main/java/io/enmasse/k8s/api/CachingSchemaProvider.java
lulf: I assume you mean the existing code. Yes, it will take the first since there is only 1 Schema. The SchemaApi/KubeSchemaApi should instead take the SchemaListener interface rather than Watcher<Schema> . I will change this at a later point to avoid too big a code change.

keycloak-user-api/src/main/java/io/enmasse/user/keycloak/KeycloakUserApi.java
lulf: Fixed

k8s-api/src/main/java/io/enmasse/k8s/api/CachingSchemaProvider.java
k-wall: Yeah, I saw it was existing code that cached _the_ schema from the list.  Happy to defer a refactor until a better time.

keycloak-user-api/src/main/java/io/enmasse/user/keycloak/KeycloakUserApi.java
k-wall: That does the job.

PR #2686, opened 2019-04-16
Collect pod logs from failed test

kornys:


kornys:
@lulf could you please add that to downstream release branch?

PR #2691, opened 2019-04-17
Allow the refresh rate to be set in the container

vbusch:
Getting the refresh rate from container environment variable.

TODO: set the variable on the consoleservice CR.  Currently the controller is setting it as a container environment variable

### DIFF
--- a/pkg/controller/consoleservice/consoleservice_controller.go
+++ b/pkg/controller/consoleservice/consoleservice_controller.go
@@ -573,7 +573,7 @@ func applyDeployment(consoleservice *v1beta1.ConsoleService, deployment *appsv1.
 
 		if consoleservice.ObjectMeta.GetAnnotations() != nil {
 			install.ApplyEnv(container, "ITEM_REFRESH_RATE", func(envvar *corev1.EnvVar) {
-				envvar.Value = consoleservice.ObjectMeta.GetAnnotations()["refreshRate"]
+				envvar.Value = consoleservice.ObjectMeta.GetAnnotations()["enmasse.io/console-refresh-rate"]
 			})
 		}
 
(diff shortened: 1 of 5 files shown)


k-wall:
Looks okay, I'm not sure this deserves its own entry in the CRD, especially as our intention is to remove the need for the periodic refresh with watch.  Perhaps we can just define an annotation?

vbusch:
@enmasse-ci run tests testcase=WebConsoleTest

vbusch:
@k-wall had to rebase it again.  
The refresh rate is now being read from the annotations: refreshRate

pkg/controller/consoleservice/consoleservice_controller.go
```diff
@@ -571,6 +571,12 @@ func applyDeployment(consoleservice *v1beta1.ConsoleService, deployment *appsv1.
 
 		install.ApplyEnvSimple(container, "OPENSHIFT_AVAILABLE", strconv.FormatBool(util.IsOpenshift()))
 
+		if consoleservice.ObjectMeta.GetAnnotations() != nil {
+			install.ApplyEnv(container, "ITEM_REFRESH_RATE", func(envvar *corev1.EnvVar) {
+				envvar.Value = consoleservice.ObjectMeta.GetAnnotations()["refreshRate"]
```
k-wall: EnMasse internal annotations should be prefixed with `enmasse.io/` to keep them clear of the user's domain.   I infer from AnnotationKeys.java that our preferred convention is hyphen separated rather than camelCase, so I would go `enmasse.io/console-refresh-rate`.

vbusch:
@enmasse-ci run tests testcase=WebConsoleTest

PR #2692, opened 2019-04-17
Enable/disable docs depending on section enabler

lulf:


documentation/assemblies/assembly-installing.adoc
```diff
@@ -5,29 +5,46 @@
...
+ifdef::OperatorSource[]
 include::../modules/proc-installing-using-olm.adoc[leveloffset=+1]
+endif::OperatorSource[]
 
 endif::[]
 ifeval::["{cmdcli}" == "oc"]
+
+ifdef::OperatorSource[]
 {ProductName} can be installed by applying the YAML files using the `oc` command-line tool, by using
 the Operator Marketplace, or by running the link:https://www.ansible.com/[Ansible^] playbooks.
```
jenmalloy: @lulf, But this sentence needs to omit the Operator Marketplace part for downstream, correct?

lulf: Yes, maybe we should leave that sentence in but not provide instructions?

jenmalloy: If the user isn't able to install that way yet, I think it needs to be deleted. @ncbaratta thoughts?

ncbaratta: I agree with @jenmalloy here.  I don't want us to mention something that the users don't have access to yet.  I'd comment it out for now and we can uncomment when it's available.

lulf: Well then the current PR is fine I think. In the downstream you don’t set the OperatorSource enabler and it will not show up

lulf: Well then the current PR is fine I think. In the downstream you don’t set the OperatorSource enabler and it will not show up

lulf: Upstream we set this enabler so that it shows up in the doc

PR #2698, opened 2019-04-18
fixes for http adapter tests

famartinrh:


famartinrh:

PR #2702, opened 2019-04-23
connect to mqtt adapter retrying

famartinrh:
this tryies to solve instability in MqttAdapterTest


famartinrh:

PR #2704, opened 2019-04-23
install connection event listeners before opening the connection

k-wall:
This change should fix an occasional test fail seen during system test runs.  It affects only the client (browser) side.  I theory is that the test failure is a race condition likely on slow hardware, where the server side manages to close the connection before the client's thread of control has had chance to install the connection listeners.


kornys:
@enmasse-ci run tests testcase=standard.web.FirefoxWebConsoleTest#testCannotOpenConsolePage

PR #2706, opened 2019-04-23
Make sure CRDs are ordered before other resources

lulf:
* Also remove number prefixing when creating OLM package


templates/Makefile
```diff
@@ -119,9 +120,17 @@ $(IOT_BUNDLE): replace_images
 OLM=$(PACKAGE_INSTALL_DIR)/olm/enmasse
 $(OLM): replace_images
 	mkdir -p $(OLM)
-	for i in crds csv; do \
+	for i in csv; do \
 		cp $(BUILDDIR)/replaced/$$i/*.yaml $(OLM)/; \
 	done
+	for i in crds; do \
```
vbusch: This patch is deleting the 7 templates/crds/*.yaml files.  What would this be iterating over?

Issue #2306, opened 2019-02-01
Design for address space bridging


lulf:
**Overview**

A commonly requested feature is the ability to bridge address spaces. I.e. have clients in address space A be able to send/receive from addreesses in Address space B.

**Design**

TODO

**Tasklist**

TODO

rgodfrey:
## Overview

This document describes alternate mechanisms for exposing addresses within one addresspace such that they can be reached
from a second address space.

## Design

An address space view consists of a restriction over the available addresses within an address space, associating
permissions (send and/or receive) with each address within the view.  The address space owner can then define which
other address spaces (managed by the same enmasse installation) will have access to the view.

### Option a)
```
apiVersion: admin.enmasse.io/v1beta1
kind: AddressSpaceView
metadata:
  name: myspace.myview
spec:
  addresses:
    - name: foo
      operations: [ send, receive ]
    - name: bar*
      operations: [ receive ]
  addressspaces: [ namespace1.A, namespace2.B ] 
    
```

### Option b)

Separate view definition from view exposure

```
apiVersion: admin.enmasse.io/v1beta1
kind: AddressSpaceView
metadata:
  name: myspace.myview
spec:
  addresses:
    - name: foo
      operations: [ send, receive ]
    - name: bar*
      operations: [ receive ]
     
```

and ...

```
apiVersion: admin.enmasse.io/v1beta1
kind: AddressSpaceViewVisibility
metadata:
  name: myspace.myview
spec:
  addressspaces: [ namespace1.A, namespace2.B ]     
```

### Option c)

Add the view definitions into the address space

```
apiVersion: enmasse.io/v1beta1
kind: AddressSpace
metadata:
  name: myspace
spec:
  type: standard
  plan: standard-unlimited
  views:
    - name: myview
      addresses:
        - name: foo
          operations: [ send, receive ]
        - name: bar*
          operations: [ receive ]
    - name: myview2
      addresses:
        - name: foo
          operations: [ receive ]          
 ```
 
 while still separating the visibility
 
 ```
 apiVersion: admin.enmasse.io/v1beta1
 kind: AddressSpaceViewVisibility
 metadata:
   name: myspace.myview
 spec:
   addressspaces: [ namespace1.A, namespace2.B ]     
 ```

**Comment by lulf** [comment]
An alternative d) for consideration (not prefix view with address space, makes it possible to 'reuse' view):

```
apiVersion: enmasse.io/v1beta1
kind: AddressSpace
[...]

PR #2710, opened 2019-04-24
*DRAFT* Initial proposal of Kafka address space

lulf:


documentation/design/proposals/kafka-address-space.adoc
```diff
@@ -0,0 +1,177 @@
...
+        type: persistent-claim
+        size: 100Gi
+        deleteClaim: false
+    entityOperator:
+      topicOperator: {}
+      userOperator: {}
+----
+
+Creating this resource will not create a Kafka cluster, but it will prepare the `address-space-controller` with the configuration needed to create one when a tenant creates an `AddressSpace` of the `kafka` type. Moreover, settings that are applied on a per-address-space basis will be ignored and set based on the `AddressSpace` definition instead. The following settings are ignored/overridden:
+
```
k-wall: This mean that all Kafka address spaces share the same Kakfa cluster, right?

k-wall:
How would this proposal live alongside #2306 (bridging)

documentation/design/proposals/kafka-address-space.adoc
lulf: No, that will come at a later point. I will try to clarify this in the doc, but in the same way we don't create the router and broker when you create the *InfraConfig, we would not create the Kafka cluster when you create the KafkaConfig. The cluster will be created when the AddressSpace is created.

I think longer term, sharing a cluster would be ideal, but it requires modifications to Kafka itself or some proxy in the middle to make it transparent.

lulf:
@k-wall In the first version I think bridging should not be supported. However, we would need to support it at some point, and I imagine it would be in a similar form as for the brokered space, by deploying an amqp-kafka bridge.

documentation/design/proposals/kafka-address-space.adoc
```diff
@@ -0,0 +1,177 @@
...
+[source,yaml,options="nowrap",subs="attributes"]
+----
+apiVersion: enmasse.io/v1beta1
+kind: AddressSpace
+metadata:
+  name: myspace
+spec:
+  type: kafka
+  plan: kafka-small
+  endpoints:
```
ctron: Would that include an AMQP 1.0 endpoint?

documentation/design/proposals/kafka-address-space.adoc
```diff
@@ -0,0 +1,177 @@
...
+[source,yaml,options="nowrap",subs="attributes"]
+----
+apiVersion: enmasse.io/v1beta1
+kind: Address
+metadata:
+  name: myspace.mytopic
+spec:
+  type: topic
+  plan: kafka-topic-small
+  properties:
```
ctron: What I don't like with the "bag of properties" concept here, is that you no real way to validate that. I would suggest to follow the pattern of Kubernetes instead: create dedicated types for the different choices: see https://github.com/kubernetes/api/blob/master/core/v1/types.go#L49

lulf: @ctron, I’m not sure I understand the reasoning behind distinct types leading to better validation.   

The proposal is not describing a “bag of properties” without a schema. Each field within the properties object may have different and possibly nested types.

This should provide the same level of validation of each property as you would from having separate objects for different address types.

Moreover, any such object would need to allow any property to be omitted and have a default in order to simplify configuration. In addition, the implementation would need to allow unknown properties in order to handle forward and backward compatibility.

With the above requirements I think either direction will give the same benefits. 

A challenge in having separate fields would be that we have the same types (I.e topic) defines in multiple address space types, so i don’t think distinct properties fields for each type makes sense.

If this makes sense to you, I can update the proposal to clarify this bit.

documentation/design/proposals/kafka-address-space.adoc
lulf: In the first iteration I think AMQP 1.0 is out of scope. I would like to see this be part of the next iteration, and this design doc should updated when preparing for that.

documentation/design/proposals/kafka-address-space.adoc
lulf: So I think the conclusion is to use the `properties` field name, but provide OpenAPI spec to ensure that only a valid set of properties are specified together. I will update this proposal to reflect that, does that sound ok @ctron ?

lulf:
@k-wall Do you want to use this document as a basis for the 'new' design or should I just close this PR?

k-wall:
We should keep the branch so that we can reuse your work later.  It'd be nice to close the PR to keep the PR queue short/clean.

PR #2727, opened 2019-04-24
AuthServiceTest - extra test-only logging to help diagnose test failure

k-wall:
I don't yet understand this test failure (AuthServiceTest#testStandardAuthServiceWithDB).  The enmasse-operator pod log is missing for the time when the test failed so I can't confirm that the operator actioned the auth service.  This extra logging should help if we see the issue again.


k-wall:
@enmasse-ci run tests testcase=AuthServiceTest

PR #2736, opened 2019-04-24
Correct console operator error handling

k-wall:
Correct defect that would prevent the creation  of OAuthClient object if the route did not yet have an ingress.


k-wall:

PR #2737, opened 2019-04-24
On Darwin changing 'base64 -w0' into 'base64 -b0'.

buuhsmead:
On Darwin base64 has no option named '-w0' but uses the option '-b0'.

This PR checks if it is or is not on Darwin and uses the current base64 option according to the used platform.


lulf:
Is there an alternative way we can make it work on Darwin, @robshelly?

robshelly:
This has now been resolved in the [this](https://github.com/EnMasseProject/enmasse/pull/2743/commits/f179abcb46ee1df1f423279e34d5fe7db88c1670) commit.

As of that commit, the Ansible playbook creates the default configuration for Alertmanger during installation, which is already Base64 encoded. If users want to create a their own configuration the documentation [describe this process ](https://enmasse.io/documentation/master/openshift/#config-alerts-messaging) and provides a sample config file.

PR #2738, opened 2019-04-25
Fix readback of authorization rules

lulf:
Authorization rules containing a '_' character will not be read back
correctly when doing a GET for a MessagingUser. This change adds a unit
test to verify the issue and a fix that ensures the entire address is
preserved when decoding the group name from Keycloak.

### DIFF
--- a/keycloak-user-api/src/main/java/io/enmasse/user/keycloak/KeycloakUserApi.java
+++ b/keycloak-user-api/src/main/java/io/enmasse/user/keycloak/KeycloakUserApi.java
@@ -586,7 +586,7 @@ public class KeycloakUserApi implements UserApi, RealmApi {
             } else if (groupRep.getName().contains("_")) {
                 String[] parts = groupRep.getName().split("_", 2);
                 Operation operation = Operation.valueOf(parts[0]);
-                String address = parts[1];
+                String address = decodePart(parts[1]);
                 operationsByAddress.computeIfAbsent(address, k -> new LinkedHashSet<>())
                         .add(operation);
             }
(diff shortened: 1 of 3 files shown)


keycloak-user-api/src/main/java/io/enmasse/user/keycloak/KeycloakUserApi.java
```diff
@@ -586,7 +586,7 @@ static User buildUser(UserRepresentation userRep, List<GroupRepresentation> grou
             } else if (groupRep.getName().contains("_")) {
                 String[] parts = groupRep.getName().split("_");
```
ctron: I think it would be simpler to use as limit of 2 here:

~~~java
String[] parts = groupRep.getName().split("_", 2);
~~~

lulf: Fixed

keycloak-user-api/src/main/java/io/enmasse/user/keycloak/KeycloakUserApi.java
```diff
@@ -584,9 +584,9 @@ static User buildUser(UserRepresentation userRep, List<GroupRepresentation> grou
             } else if (groupRep.getName().startsWith("monitor")) {
                 globalOperations.add(Operation.view);
             } else if (groupRep.getName().contains("_")) {
-                String[] parts = groupRep.getName().split("_");
+                String[] parts = groupRep.getName().split("_", 2);
                 Operation operation = Operation.valueOf(parts[0]);
-                String address = decodePart(parts[1]);
+                String address = parts[1];
```
ctron: I think you still need to decode the part though (see failing test):

```suggestion
                String address = decodePart(parts[1]);
```

PR #2741, opened 2019-04-25
Further console operator improvements

k-wall:
We saw a case in an OCP4 test environment where the sso-cookie secret didn't get automatically generated.  This changes the reconcile algorithm so that if that situation were to occur, the secret would get generated on the next pass.

Also changed the reconciler so that if it itself updates the consoleservice CR it suppresses the requeue to prevent a needless iteration (that usually resulted in errors in the logs).


k-wall:

PR #2742, opened 2019-04-25
Adding IoT MQTT examples to the docs

dejanb:


templates/iot/examples/k8s-tls/create
```diff
@@ -36,8 +37,13 @@ echo $CA_SUBJECT
...
 function createServiceKeyAndCert() {
-  openssl req -config "$SCRIPTDIR/ca.cnf" -nodes -newkey rsa:4096 -keyout "$OUTDIR/$1-key.pem" -days 365 -subj "/O=EnMasse/OU=IoT/CN=$1" | \
-  SAN="DNS.1:${1}.svc,DNS.2:${1}.${NAMESPACE}.svc,DNS.3:localhost" openssl x509 -req -extfile "$SCRIPTDIR/ca.cnf" -extensions "san_ext" -out "$OUTDIR/$1.pem" -days 365 -CA "$OUTDIR/ca-cert.pem" -CAkey "$OUTDIR/ca-key.pem" -CAcreateserial
+  DNSHOSTNAME=${CLUSTER}
+  if [ ${CLUSTER} != "localhost" ]; then
+    DNSHOSTNAME=${1}-${NAMESPACE}.${CLUSTER}
+  fi
+  echo $DNSHOSTNAME
+  openssl req -config "$SCRIPTDIR/ca.cnf" -nodes -newkey rsa:4096 -keyout "$OUTDIR/$1-key.pem" -days 365 -subj "/O=EnMasse/OU=IoT/CN=$DNSHOSTNAME" | \
+  SAN="DNS.1:${1}.svc,DNS.2:${1}.${NAMESPACE}.svc,DNS.3:$DNSHOSTNAME" openssl x509 -req -extfile "$SCRIPTDIR/ca.cnf" -extensions "san_ext" -out "$OUTDIR/$1.pem" -days 365 -CA "$OUTDIR/ca-cert.pem" -CAkey "$OUTDIR/ca-key.pem" -CAcreateserial
```
ctron: I think we should leave the CN and DNS entry for localhost in place. Simply add another `DNS.4` entry with `$DNSHOSTNAME`

documentation/modules/proc-iot-installing-services.adoc
```diff
@@ -30,10 +30,20 @@ ifeval::["{cmdcli}" == "oc"]
...
 ./install/components/iot/examples/k8s-tls/create
 {cmdcli} create secret tls iot-mqtt-adapter-tls --key=install/components/iot/examples/k8s-tls/build/iot-mqtt-adapter-key.pem --cert=install/components/iot/examples/k8s-tls/build/iot-mqtt-adapter-fullchain.pem
-{cmdcli} create secret generic iot-service-ca --from-file=service-ca.crt=install/components/iot/examples/k8s-tls/build/trusted-certs.pem
 ----
 endif::[]
 
++
+[NOTE]
+====
+If your cluster is not running on the `localhost`, you need to specify cluster IP address when creating certificates. For example:
```
ctron: You actually need to specify the hostname, but you can use "nip.io" if you don't have one, with your IP address.

I know that, in this case, it only is a small difference, but putting your actual IP here wouldn't work if I understood it right.

documentation/modules/proc-iot-api.adoc
```diff
@@ -68,4 +72,9 @@ mvn spring-boot:run -Drun.arguments=--hono.client.host=$MESSAGING_HOST,--hono.cl
 curl --insecure -X POST -i -u sensor1@myapp.iot:hono-secret -H 'Content-Type: application/json' --data-binary '{"temp": 5}' https://{http-adapter}/{api}
 ----
 
-// TODO cover using MQTT here
+. Send {api} message using MQTT:
```
jenmalloy:
```suggestion
. Send an {api} message using MQTT:
```

documentation/modules/proc-iot-installing-services.adoc
jenmalloy:
```suggestion
If your cluster is not running on the `localhost`, you need to specify the cluster IP address when creating certificates. For example:
```

documentation/modules/proc-iot-api.adoc
```diff
@@ -2,9 +2,13 @@
...
+:mqtt-adapter: $(oc -n enmasse-infra get routes iot-mqtt-adapter --template='{{ .spec.host }}')
+:mqtt-port: 443
 endif::[]
 ifeval::["{cmdcli}" == "kubectl"]
 :http-adapter: $(kubectl -n enmasse-infra get service iot-http-adapter-external -o jsonpath={.status.loadBalancer.ingress[0].hostname}):30443
+:mqtt-adapter: $(kubectl -n enmasse-infra get service iot-mqtt-adapter-external -o jsonpath={.status.loadBalancer.ingress[0].hostname})
+:mqtt-port: 30883
 endif::[]
 
 = Sending and receiving {api} data
```
jenmalloy: Is `{api}` meant as a literal value in this section? Could it be changed to "API" instead?

templates/iot/examples/k8s-tls/create
dejanb: @ctron should be fixed now

documentation/modules/proc-iot-api.adoc
dejanb: So, {api} will be resolved to either "telemetry" or "event", like 
https://enmasse.io/documentation/master/openshift/#iot-telemetry-messaging-iot
The idea is not to duplicate docs as the topic is pretty much the same

documentation/modules/proc-iot-api.adoc
dejanb: Removed {api} from the text, it will remain just "a message"

documentation/modules/proc-iot-api.adoc
jenmalloy: @dejanb, ah, got it. (Sorry, I think I figured this out during my last review and then forgot again)

PR #2746, opened 2019-04-27
Switch to using quay.io for images

lulf:


k-wall:
Looks reasonable to me. Should the other EnMasse *-*-base projects (router-docker-base) be updated too?

lulf:
@k-wall router-docker-base, java-docker-base and nodejs-docker-base are already updated :) 

Pushed a fix to fix broker startup wrt. new java options.

lulf:

PR #2752, opened 2019-04-29
Fix npm shrinkwraps

k-wall:
No functional change.


k-wall:
@vbusch and @jbtrystram can you tell me if the package-locks are stable for you with this PR?  I've tested this on two machines and it seems good to me.

k-wall:
@vbusch does f16a060 help?

lulf:
Also works for me, thanks @k-wall

PR #2754, opened 2019-04-30
Updated restarting pod alert trigger expression

robshelly:


robshelly:
@lulf That's a good point. I hadn't considered that it would still take longer to trigger on quick restarts.

What do you think if labelling the quick restarts with a `severity: critical` versus `severity: warning` on the slower ones? That way we can easily use inhibit rules to prevent two alerts triggering in the case of quick restarts.

PR #2761, opened 2019-05-02
Clean up logging, at least a bit

ctron:


ctron:

PR #2757, opened 2019-05-01
Extra logging/defensive code to help understand problem during CI run

k-wall:
I am seeing an odd problem during CI of MsgPatternsInternalTest#testMqttMessage.  I am hoping this will help me understand why.

PR #2765, opened 2019-05-02
Wait for client isRunning=false in MsgCliApiClient.getClientInfo

famartinrh:


k-wall:
@famartinrh The time calculations look muddled.  150000 *seconds* = 1 day.   Do we really want to hang the tests for a 1 day if something goes wrong and never completes?

famartinrh:
@k-wall you are totally right, this was a mistake, currently I'm changing this timeout and merging with your PR #2757 and both things should be a valid workaround

PR #2766, opened 2019-05-02
Fix vertx/vertx web resource leak

k-wall:


k-wall:
> I tested this locally and still fails, but anyway these changes are improvements, so merge it anyway.
> I think merging changes from #2757 is a good idea, don't know which other options we have

Unfortunately, I don't see this failing anywhere apart from CI, so it is difficult for me to debug.   Just out of interest @famartinrh if you run the tests separately i.e. `SYSTEMTEST_ARGS="io.enmasse.systemtest.brokered.clients.mqtt.MsgPatternsInternalTest"` (io.enmasse.systemtest.standard.clients.mqtt.MsgPatternsInternalTest) do they fail too?  Knowing this would help understand if the problem is triggered by some remaining state from earlier invocations

k-wall:
> I tested this locally and still fails, but anyway these changes are improvements, so merge it anyway.

Unfortunately, I don't see this failing anywhere apart from CI, so it is difficult for me to debug.   Just out of interest @famartinrh if you run the tests on there own, i.e. SYSTEMTEST_ARGS="io.enmasse.systemtest.brokered.clients.mqtt.MsgPatternsInternalTest" does it fail too?  

> I think merging changes from #2757 is a good idea, don't know which other options we have

Does #2757 work around the problem for you?

famartinrh:
@k-wall running tests separately they pass, I can reproduce the failure using this arguments `SYSTEMTEST_ARGS="**.MsgPatternsInternalTest#testMqttMessage"` or this `SYSTEMTEST_ARGS="**.MsgPatternsInternalTest" `

k-wall:
> @k-wall running tests separately they pass, I can reproduce the failure using this arguments `SYSTEMTEST_ARGS="**.MsgPatternsInternalTest#testMqttMessage"` or this `SYSTEMTEST_ARGS="**.MsgPatternsInternalTest"`

interesting... so it must be some state left behind by previous tests.   I did wonder if it could be something like the client's HTTP kept alive connections to pods that get recycled?  But the closure of the resources should guard against this......

If you could run your failing case with -Djavax.net.debug=all that might shed some light.

PR #2772, opened 2019-05-03
Use dispatch feature of setting vhost policies per listener

lulf:
This change sets the vhost policies only on public listeners so that
internal components are not restricted.


lulf:
@enmasse-ci run tests profile=standard-shared testcase=InfraTest

lulf:
@grs I'm having some issues making this PR run. I get the following error from the router when starting:

```
2019-05-06 07:45:46.160934 +0000 MAIN (critical) Router start-up failed: Python: PolicyError: Policy 'public' is invalid: Policy vhost 'public' allows connections by default but default sett
ings are not defined
```

The policy file looks as follows:

```
[["vhost",{"hostname":"public","maxConnectionsPerUser":500,"allowUnknownUser":true,"groups":{"public":{"allowDynamicSource":true,"sources":"*","maxSenders":300,"allowAnonymousSender":true,"remoteHosts":"*","targets":"*"}},"maxConnectionsPerHost":500,"maxConnections":1000}]]s
```

The public listeners look like this:

```
listener {                                                                                                                                                        
  ...
  policyVhost: public                                                                                                                                                                         
}      
```

At first sight, it seems maybe I still need to use '$default', but the question is then how do I prevent it from being used by other listeners?

grs:
Set the 'group' name to $default. We can't use named users groups as our users are not defined in the router config.

Note that you are not preventing other listeners from using this virtual host by specifying policyVHost, you are just preventing users connecting to *this* listener from using any other vhost. I think that is fine though. However if you want to actively prevent components connected to the internal listeners from using this vhost, you would need to setup a separate internal vhost and restrict those listeners to that vhost.

lulf:
@enmasse-ci run tests profile=standard-shared testcase=InfraTest

PR #2782, opened 2019-05-07
Use ENMASSE_OPENSHIFT as a short-circuit to determine the presence of OpenShift

k-wall:
The Go path already uses this environment variable so it seems reasonable for Java to do the same.

### DIFF
--- a/address-space-controller/pom.xml
+++ b/address-space-controller/pom.xml
@@ -20,6 +20,10 @@
     </dependency>
     <dependency>
       <groupId>io.enmasse</groupId>
+      <artifactId>api-common</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>io.enmasse</groupId>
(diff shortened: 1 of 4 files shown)


k-wall:
> LGTM
> 
> An additional place where we are using a check for openshift is in the console, OPENSHIFT_AVAILABLE:
> 
> * pkg/controller/consoleservice/consoleservice_controller.go
> * console/console-init/oauth-proxy/bin/init.sh
> 
> For consistency, could rename the variable to ENMASSE_OPENSHIFT

I think the use-cases for the two environment variables are different.  `ENMASSE_OPENSHIFT` is a way for the deployer to tell the EnMasse software it is running under OpenShift, so don't bother use heuristics to detect it yourself.  `OPENSHIFT_AVAILABLE` is just a Console concern and is a mechanism to get knowledge of the presence of OpenShift over to the client side.   I think conflating the two would be a mistake.

PR #2786, opened 2019-05-08
test for multiple iot projects

famartinrh:
Adds new test `MultipleProjectsTest` which for now runs a simple workload over http-adapter of 2 iot projects concurrently.
The idea with this test and more similar tests to come is to verify the behaviour of the platform in a more real case because the iot infrastructure is shared between tenants


famartinrh:
closing because this work is continued in #2816

PR #2787, opened 2019-05-08
suggest infinispan DR in IoT Readme

jbtrystram:
While we'll still working on the device registry and writing the docs , here is a quick guide to use the infinispan DR easily.

Inspiration is from https://github.com/infinispan-demos/infinispan-kubernetes

### DIFF
--- a/templates/iot/examples/infinispan-service.yml
+++ b/templates/iot/examples/infinispan-service.yml
@@ -1,0 +1,80 @@
+apiVersion: v1
+kind: Service
+metadata:
+  annotations:
+    description: Hot Rod's port.
+  labels:
+    application: infinispan-server
+  name: infinispan-server-hotrod
+spec:
+  ports:
+  - port: 11222
+    targetPort: 11222
+  selector:
+deploymentConfig: infinispan-server
+---
+apiVersion: apps/v1beta1
+kind: StatefulSet
+metadata:
+  labels:
+    application: infinispan-server
+  name: infinispan-server
+spec:
+  replicas: 1
+  serviceName: infinispan-server-headless
+  template:
+    metadata:
+      labels:
+        application: infinispan-server
+        deploymentConfig: infinispan-server
+      name: infinispan-server
+    spec:
+      containers:
+      - args:
+        - cloud
+        - -Djboss.default.jgroups.stack=kubernetes
+        env:
+        - name: KUBERNETES_LABELS
+          value: application=infinispan-server
+        - name: KUBERNETES_NAMESPACE
+          valueFrom:
+            fieldRef:
+              fieldPath: metadata.namespace
+        image: jboss/infinispan-server:9.4.6.Final
+        livenessProbe:
+          exec:
+            command:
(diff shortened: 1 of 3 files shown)


templates/iot/examples/infinispan-service.yml
```diff
@@ -0,0 +1,80 @@
...
+          failureThreshold: 5
+          initialDelaySeconds: 10
+          periodSeconds: 10
+          successThreshold: 1
+          timeoutSeconds: 80
+        resources:
+          requests:
+            cpu: "0.5"
+            memory: 512Mi
+terminationGracePeriodSeconds: 120
```
ctron: I think this last line should be indented.

jbtrystram: fixed :)

templates/iot/examples/infinispan-service.yml
```diff
@@ -0,0 +1,80 @@
...
+    description: Hot Rod's port.
+  labels:
+    application: infinispan-server
+  name: infinispan-server-hotrod
+spec:
+  ports:
+  - port: 11222
+    targetPort: 11222
+  selector:
+deploymentConfig: infinispan-server
```
ctron: I guess there is an indent problem as well

jbtrystram: yep. it's odd because I tested it and all went fine. fixed

templates/iot/examples/infinispan-service.yml
```diff
@@ -0,0 +1,80 @@
+apiVersion: v1
+kind: Service
+metadata:
+  annotations:
+    description: Hot Rod's port.
+  labels:
+    application: infinispan-server
+  name: infinispan-server-hotrod
```
ctron: This doesn't seem to math with the "example" IoTConfig, which defines `example-infinispan` instead of `infinispan-server-hotrod`.

jbtrystram: yes you are right, the value in the _example_ file is a remnant from the service exposed by the infinispan operator which I used during development.
thanks for catching this

ctron:
I think we need the following in addition:

~~~yaml
---
apiVersion: v1
kind: ServiceAccount
metadata:
  name: infinispan
  labels:
    app: enmasse
    enmasse-component: iot
---
kind: Role
metadata:
  labels:
    app: enmasse
    enmasse-component: iot
  name: iot.enmasse.io:infinispan
rules:
  - apiGroups: [ "" ]
    resources: [ "pods" ]
    verbs: [ "get", "list", "watch" ]
---
kind: RoleBinding
metadata:
  name: infinispan
  labels:
    app: enmasse
    enmasse-component: iot
roleRef:
  kind: Role
  name: iot.enmasse.io:infinispan
subjects:
- kind: ServiceAccount
  name: infinispan
  namespace: ${NAMESPACE}
~~~

And then set the `serviceAccount` in the `StatefulSet`

~~~
.spec.template.spec.serviceAccount: infinispan
~~~

PR #2788, opened 2019-05-08
Docs: IoT docs copyedits

jenmalloy:


documentation/modules/proc-iot-creating-project.adoc
```diff
@@ -7,7 +12,7 @@ ifeval::["{cmdcli}" == "oc"]
 . Change to a non-admin user:
```
jenmalloy: @dejanb Seems like the command is missing here? Or maybe the colon should just be a period here (if it depends what the exact command will be)

dejanb: @jenmalloy You're right ... it should be the same as Step 1 in https://enmasse.io/documentation/master/openshift/#create-address-space-cli-messaging-gs (I can't link to it in github properly).

jenmalloy:
@dejanb, Could you review the latest changes please? Thanks!

documentation/modules/proc-iot-api.adoc
```diff
@@ -13,17 +18,22 @@ endif::[]
 
 = Sending and receiving {api} data
 
-.Procedure
+== Starting the {api} consumer
 
-== Starting {api} consumer
+.Prerequisites
```
bhardesty: Just a minor suggestion for prereqs: I typically state them as present conditions (so, "The IoT services are installed", "An IoT project is created", etc.)

However, if the other prereqs in this doc are phrased a certain way, I'd just match it here for consistency.

documentation/modules/proc-iot-api.adoc
```diff
@@ -13,17 +18,22 @@ endif::[]
...
 
-. In Hono project run:
+. In the Hono project, run:
 +
 [options="nowrap",subs="attributes"]
 ----
 cd cli
 ----
-. At least once run:
+. At least once, run:
```
bhardesty: Maybe something like, "Run the following command at least once:"

Still, it's not clear what "at least once" means. How do I know if I need to run it more than once?

documentation/modules/proc-iot-creating-project.adoc
```diff
@@ -1,13 +1,28 @@
...
+//
+// assembly-iot-guide.adoc
+// assembly-IoT.adoc
+
 [id='iot-creating-project-{context}']
 = Creating an IoT project
 
+After installing the IoT services, you create an IoT project.
+
+.Prerequisite
```
bhardesty: ".Prerequisites" should be plural even if there's only one prereq.

PR #2789, opened 2019-05-08
Fix #1280: implement watch (work in progress)

### LINKED ISSUE #1280: Add support for watching addressspaces and addresses to api server

k-wall:
Uses @lulf's idea to proxy a watch of the config map backing the address spaces/addresses.
Tabluar watches (i.e oc get addressspace -o wide -w) not showing correct columns yet.


k-wall:
Re. the comment about tabular watches is spurious.   I see now that when watching a resource with `--output=wide` the client does not make a request with Accept "application/json;as=Table".   It seems the list of columns that are displayed in the watch/wide is determined differently.  I see that for CRDs allow the list of additional columns to be configured https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/#additional-printer-columns.  Aside: Is `additional-printer-columns` supported by API extensions that are not CRDs??

k-wall:
Closing PR - likely we will take a different approach - #3231

PR #2796, opened 2019-05-09
Prefixed metric names with application name

robshelly:


metrics-api/src/main/java/io/enmasse/metrics/api/PrometheusMetricsFormatter.java
```diff
@@ -4,19 +4,37 @@
...
+import org.slf4j.LoggerFactory;
+
 
 public class PrometheusMetricsFormatter implements MetricsFormatter {
+    private static final Logger log = LoggerFactory.getLogger(PrometheusMetricsFormatter.class.getName());
     public String format(List<Metric> metricList) {
         StringBuilder sb = new StringBuilder();
+        String appPrefix = "";
+
+        try {
```
k-wall: Use try-with-resources in order to automatically close the stream

metrics-api/src/main/resources/pom.properties
```diff
@@ -0,0 +1 @@
+appPrefix=${application.bundle.prefix}
```
k-wall: I'd suggest using a filename other than `pom.properties`.  Maybe just metrics.properties.

metrics-api/src/main/java/io/enmasse/metrics/api/PrometheusMetricsFormatter.java
```diff
@@ -4,19 +4,37 @@
...
+            java.io.InputStream inputStream = this.getClass().getResourceAsStream("/pom.properties");
+            java.util.Properties properties = new Properties();
+
+            properties.load(inputStream);
+            appPrefix = properties.getProperty("appPrefix");
+        } catch (IOException e) {
+            log.info("Error reading app prefix {}", e);
```
k-wall: I think failing early is better than continuing in an unexpected state.  I'd just throw an exception and let the program die.

metrics-api/src/main/java/io/enmasse/metrics/api/PrometheusMetricsFormatter.java
```diff
@@ -4,19 +4,37 @@
...
+
 
 public class PrometheusMetricsFormatter implements MetricsFormatter {
+    private static final Logger log = LoggerFactory.getLogger(PrometheusMetricsFormatter.class.getName());
     public String format(List<Metric> metricList) {
         StringBuilder sb = new StringBuilder();
+        String appPrefix = "";
+
+        try {
+            java.io.InputStream inputStream = this.getClass().getResourceAsStream("/pom.properties");
```
k-wall: getResourceAsStream returns null if the file is not found.  You should throw an exception if null is returned.

PR #2813, opened 2019-05-13
fix infinite loop MsgcliApiClient

famartinrh:


famartinrh:
@enmasse-ci run tests testcase=**.MsgPatternsInternalTest

famartinrh:
@enmasse-ci run tests testcase=**.MsgPatternsInternalTest

famartinrh:
@k-wall you are totally right, this recursion is unnecessary. I'm adding another commit for it.
The infinite loop was actually fixed in my second commit, because we looped and did retryies but the CompletableFuture holding the response wasn't re-initialized causing the infinite loop. This changes are made to hunt that issue that happened during weekend

famartinrh:
@enmasse-ci run tests testcase=**.MsgPatternsInternalTest

PR #2814, opened 2019-05-14
Fix global console tests selector

kornys:


kornys:
@enmasse-ci run tests testcase=common.console.**

PR #2816, opened 2019-05-14
Test of multiple iotprojects

famartinrh:
Adds a few fixes for iot tests with the new usage of fabric8 clients in systemtests
Also adds new test MultipleProjectsTest which for now runs a simple workload over http-adapter of 2 iot projects concurrently.
The idea with this test and more similar tests to come is to verify the behaviour of the platform in a more real case because the iot infrastructure is shared between tenants


famartinrh:

Issue #2821, opened 2019-05-16
Operator install stays in unknown state due to incorrect YAML in docs


garethahealy:
**Describe the bug**
Following the install for the operator doesn't work. The Subscription has an invalid source value. Currently points to _enmasse-operators_, but should be pointing to the named CatalogSourceConfig, which is: _installed-enmasse-operators_ 

**To Reproduce**
Following docs @ https://github.com/EnMasseProject/enmasse/blob/e44c8c596d187171e580746d1caba84a3ce13aa9/documentation/modules/proc-installing-using-olm.adoc

**Expected behavior**
Operator installs successfully following docs

k-wall:
Fixed by #2822

PR #2822, opened 2019-05-16
[2821] Fixed docs YAML using incorrect source name for operator install

garethahealy:
https://github.com/EnMasseProject/enmasse/issues/2821

### DIFF
--- a/documentation/modules/proc-installing-using-olm.adoc
+++ b/documentation/modules/proc-installing-using-olm.adoc
@@ -23,7 +23,7 @@ cat <<EOF | {cmdcli} apply -f -
 apiVersion: operators.coreos.com/v1
 kind: OperatorSource
 metadata:
-  name: enmasse-operators
+  name: {OperatorSourceName}
   namespace: {OperatorMarketplaceNamespace}
 spec:
   type: appregistry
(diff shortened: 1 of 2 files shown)


garethahealy:
@k-wall @lulf ; is this OK to be merged?

jenmalloy:
I approve from a documentation standpoint, but I'll let @k-wall approve the technical aspect first

documentation/modules/proc-installing-using-olm.adoc
```diff
@@ -65,7 +65,7 @@ metadata:
 spec:
   channel: alpha
   name: {OperatorName}
-  source: {OperatorSourceName}
+  source: {OperatorSourceConfigName}
```
ctron: To me the change looks fine. However I think that for the `name` attribute in the other CR instances (`CatalogSourceConfig` and `OperatorSource), we should use the parameter as well.

And the more I look at this, the less I think this will work downstream. At least the publisher and display names should be replaced as well. But more important, the app registry will not contain downstream images. And the operator package name will be different as well.

For this PR I think it is fine, but I think we need some more work in general.

garethahealy: @ctron ; is this OK to be merged then? PR says you've requested changes but at same time say its fine.

ctron: I said it is fine in general, but taking a look at the `name` parameters, don't you think they should be changed as well?

garethahealy: rebased and updated name params to use attributes.

PR #2829, opened 2019-05-20
Add missing serialVersionID, when implementing Serializable

ctron:


jbtrystram:
I added the `Protostream` annotations on those classes so i don't it's still needed to expand Serializable... Tired so many things to make this work, i lost track on some of them :laughing: 
I'll double check and update this.

jbtrystram:
@ctron #2835 should fix that ;)

PR #2833, opened 2019-05-20
fix bug in restart during messaging test

famartinrh:


famartinrh:
@enmasse-ci run tests testcase=**.CommonTest

PR #2835, opened 2019-05-20
remove uneeded serializable in Infinispan Device Registry

jbtrystram:
using the protostream annotations and unsing Infinispan's ProtoStreamMarashaller does the serialization work, there is no need to extend serializable.


jbtrystram:

PR #2844, opened 2019-05-24
Fix the QDR image name, comes now from quay.io

ctron:


ctron:

PR #2846, opened 2019-05-28
Correct parsing of group header

lulf:
The api-server does not properly parse the X-Remote-Group header when it contains multiple comma-separated values. This prevents the api-server to work OOTB with OLM on certain
installations such as minikube, possibly also OCP4.


lulf:
@k-wall can this be put into 0.28.1 ? I think it will potentially solve some operator hub /open shift 4 issues

PR #2857, opened 2019-05-31
Handle the upgrade of pre-0.27 authentication services within the authentication service operator.

k-wall:
This change enhances the authentication service operator so that it extracts resources/claim/secret details from the existing keycloak deployment and creates the appropriate `authenticationservice` record.  On OpenShift, if a `postgresql` deploymentconfig exists and the database details tally with those of keycloak deployment a postgresql type authenticationservice record is created instead.


k-wall:
I still need to adjust ansible not to create authenticationservice record if keycloak deployment already exists (for the upgrade using Ansible case).

Issue #2825, opened 2019-05-16
DashboardURL on ServiceInstances created 1.0 do not work in 1.1


philbrookes:
**Describe the bug**
Any ServiceInstances created in AMQ Online 1.0 will have dashboard URLs that do not work once having upgraded to AMQ Online 1.1.

**To Reproduce**
- Install AMQ Online 1.0
- Deploy AMQ Online Standard from the catalog
- Check Dashboard URL works normally
- Upgrade to AMQ Online 1.1 with [these instructions](https://access.redhat.com/documentation/en-us/red_hat_amq/7.2/html-single/installing_and_managing_amq_online_on_openshift_container_platform/index#upgrading-enmasse-using-ansible-messaging)
- Check Dashboard URL is no longer valid.

**Expected behavior**
The dashboard URL should still work.

**Screenshots**
N/A

**Additional context**
N/A

rgodfrey:
@philbrookes when you say "Dashboard URL is no longer valid" are you saying you get a 404, or that you can no longer log in, or some other symptom?

vbusch:
Reproduced with amq-online 1.0, and upgrading to 1.1
SSO redirection is failing: 

https://172.30.149.188:8443/auth/realms/amq-online-infra-my-space/protocol/openid-connect/auth?client_id=enmasse-console&state=0612acaa-6576-4096-b613-6b2e349b6f85&redirect_uri=https%3A%2F%2Fconsole-o66h8qvqqa-amq-online-infra.127.0.0.1.nip.io%2F%3Fauth_callback%3D1&scope=openid&response_type=code&kc_idp_hint=openshift-v3
/auth/realms/amq-online-infra-my-space/protocol/openid-connect/auth?client_id=enmasse-console&state=fc8e1b99-ad3b-4a59-bd4a-6c0b8a100f73&redirect_uri=https%3A%2F%2Fconsole-o66h8qvqqa-amq-online-infra.127.0.0.1.nip.io%2F%3Fauth_callback%3D1&scope=openid&response_type=code&kc_idp_hint=openshift-v3
![console_url](https://user-images.githubusercontent.com/1995405/58603348-68e6be80-825e-11e9-8eb5-fed78fb4c52b.png)

It is fixed upstream.  Tested with 0.26 and upgrading to latest.

vbusch:
Existing address spaces are not updated during the upgrade process.  The address space configurations should be reviewed, and then updated manually.   Added documentation for how to do this.

[...]

PR #2859, opened 2019-05-31
Added documentation for upgrading address spaces

vbusch:
fixes: #2825 Address spaces are not upgraded as part of the upgrade process.  Adding documentation for how to do it manually.

### DIFF
--- a/documentation/assemblies/assembly-service-admin-guide.adoc
+++ b/documentation/assemblies/assembly-service-admin-guide.adoc
@@ -15,6 +15,8 @@ include::assembly-configuring.adoc[leveloffset=+1]
 
 include::assembly-upgrading.adoc[leveloffset=+1]
 
+include::../modules/proc-upgrading-address-spaces.adoc[leveloffset=+1]
+
 include::../modules/proc-uninstalling.adoc[leveloffset=+1]
 
 include::assembly-monitoring.adoc[leveloffset=+1]
--- a/documentation/assemblies/assembly-upgrading.adoc
+++ b/documentation/assemblies/assembly-upgrading.adoc
@@ -10,6 +10,8 @@ configuration of the new version will trigger the upgrade to start.
 
 Upgrading {ProductName} is done by applying the YAML files for the new version.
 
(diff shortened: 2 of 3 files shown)


k-wall:
thanks @vbusch  for the address space upgrade investigation.  The user having to trigger it by amending the infraconfigs' version just seems weird to me.  Did we really design it that way?  To me the implication is that the infraconfig plans should always be replaced as part of the upgrade, but this would be odd as the infraconfig belongs to the user. They may have tailored it.  Also what would it mean if, after the upgrade, I didn't update the infraconfig's version and then created new addressspaces.  The 'old' infraconfig would applied to the new addresspace.  If I then went on to update the infraconfig, changes would be ignored by the controller because the version didn't match that of the running code..   One thing I don't understand at the moment is what circumstances we actually materialise the version field in the infraconfig.  The infraconfig from the install is versionless.

lulf:
@k-wall @vbusch allow me to shed some light on this :) In 0.26 and older, the 'version' field was mandatory, and upgrading required applying new infra configs with the updated version. The problem with this, as you point out Keith, is that the infra configs are user owned and it is odd to having to reapply it for every version.

So, in 0.27 and newer, the version field was made optional, so that users would not have to maintain the version number.  If version is not set, KubeSchemaApi.assembleSchema() will set the version of the parsed infra config to the same version as the controller before updating the 'current' schema.

The reconciliation loop will compare the infra config in the address space annotation with the 'desired' infra config to decide if changes should be applied (see CreateController).

Why upgrade did not happen automatically I'm not sure, but hopefully the above should explain a little why things are the way they are.

vbusch:
@lulf thank you, that explanation was really helpful.
For upstream, sounds like we won't need this if we assume the users have a version newer then 0.26

k-wall:
@vbusch if `io.enmasse.k8s.api.KubeSchemaApi#assembleSchema` were changed to unconditionally replace the infraconfig's version with the runtime version, wouldn't that give the correct behaviour?   Whenever the version changes, surely we want to give the addressspacecontroller the chance to rewrite the deployments etc.

vbusch:
@k-wall When upgrading from 0.26 to 0.27, the version is removed.  Which allows the upgrade from 0.27 to 0.28 to work.  When going from 0.26 -> 0.28 removing the check as you suggested fixes the issue as well.
Closing this, since docs upstream are not required.

PR #2883, opened 2019-06-07
Docs for issue-2873: Upgrading router pods

jenmalloy:


documentation/assemblies/assembly-upgrading.adoc
```diff
@@ -1,20 +1,23 @@
...
 // assembly-service-admin-guide.adoc
+
 [id='upgrading-{context}']
 = Upgrading {ProductName}
 
 {ProductName} supports upgrades between minor versions using cloud native tools and the same
-mechanism used to apply configuration changes.  When upgrading, the updated infrastructure
-configuration of the new version will trigger the upgrade to start.
+mechanism used to apply configuration changes. When upgrading, the updated infrastructure
+configuration of the new version triggers the upgrade process to begin.
```
k-wall: First sentence is not too clear.  Is the first clause worthwhile?  It is the same tooling we introduced the reader to in the install guide. With the second clause we are trying to recommend that the user selects the same upgrade mechanism (ansible or bundle) as they did for the install, but I don't think this comes across clearly.

Second sentence is no longer correct.  It should say "applying the configuration change automatically triggers the upgrade process to start".  It is no longer trigger by an update to the infraconfig.

documentation/modules/proc-upgrading-router-pods.adoc
```diff
@@ -0,0 +1,33 @@
+// Module included in the following assemblies:
+//
+// assembly-upgrading.adoc
+
+[id='post-upgrade-step-router-pod-{context}']
+= Upgrading router pods running a previous version of the image
+
+When upgrading, sometimes a router pod is not upgraded. The affected router pod can be identified and restarted to fix this issue.
```
k-wall: Perhaps "When upgrading, sometimes a router pod is not automatically restarted by {KubePlatform}.  If this happens it is important to manually restart the affected pod in order to ensure the correct operation of the system"

documentation/assemblies/assembly-upgrading.adoc
jenmalloy: @k-wall, Does the release template method still work, or does that need to be removed?

jenmalloy:
@k-wall, Can you please take a look at the changes?

documentation/assemblies/assembly-upgrading.adoc
k-wall: Are you referring to the bundles? If so, yes, they still work/supported.

PR #2884, opened 2019-06-08
Increase timeout for qdstat in test

vbusch:
Attempt to prevent random failures, by giving qstat more time to complete.


vbusch:
@k-wall updated the patch to make the change in KubeCMDClient.

PR #2885, opened 2019-06-08
Retry the dockerclient response if it is still running (test only change)

k-wall:
Aids an investigation into a Jenkins problem


k-wall:
Unintentionally duplicates @famartinrh work #2765

PR #2888, opened 2019-06-09
Make standard controller check for topic reg with subserv before considering address ready.

k-wall:


kornys:
@enmasse-ci run tests testcase=standard.mqtt.**

k-wall:
@enmasse-ci run tests testcase=TopicTest

vbusch:
@rgodfrey we would still like to get your opinion on this fix.  We are merging it in now, so we can create new images with it for the ci.

lulf:
The fix looks ok, and the reasoning behind the original issue looks good. 

I noticed there is some code that could be shared between the subserv collector and the router collector. Moreover, kube is queried twice for the router pods instead of querying once and passing the list of pods to both status checks.

 In general I don’t like that we have a subserv dependency, but I guess the lifetime of both standard controller and subserv may be so that this fix is acceptable short term, and we can rethink MQTT in the longer term...

k-wall:
> The fix looks ok, and the reasoning behind the original issue looks good.
> 
> I noticed there is some code that could be shared between the subserv collector and the router collector. Moreover, kube is queried twice for the router pods instead of querying once and passing the list of pods to both status checks.
> 
> In general I don’t like that we have a subserv dependency, but I guess the lifetime of both standard controller and subserv may be so that this fix is acceptable short term, and we can rethink MQTT in the longer term...

@lulf thanks for the comments. I agree with all that you said.  With regard to the code duplication, I needed to make the changes in a minimally invasive way for branch to minimise the risk of regression.  I Intend to refactor on trunk to remove duplication and prevent the requery to kube.     I too find the additional standard-controller => subserv horrid.  I think the real task here is to rearchitect subserv away.

PR #2891, opened 2019-06-10
Prevent sporadic failures from CommonTest#testMessagePersistent

k-wall:
On Jenkins, we occasional saw this test fail with an SSLHandshakeException.  I think what is happening is the service has not yet realised that the pod has become ready.

(test change only)


k-wall:
@enmasse-ci run tests testcase=CommonTest#testMessagePersistent

PR #2892, opened 2019-06-10
Internal clients able to run on multinode cluster

kornys:


kornys:
@enmasse-ci run tests testcase=MsgPatternsInternalTest

PR #2900, opened 2019-06-12
Cleaned up use of IMAGE_PULL_POLICY

vbusch:


address-space-controller/src/main/resources/templates/standard-space-infra.yaml
```diff
@@ -462,8 +462,6 @@ objects:
             value: ${BROKER_PLUGIN_IMAGE}
           - name: TOPIC_FORWARDER_IMAGE
             value: ${TOPIC_FORWARDER_IMAGE}
-          - name: ${IMAGE_PULL_POLICY}
```
k-wall: I think the defect is that "- name: ${IMAGE_PULL_POLICY}" should read "- name: IMAGE_PULL_POLICY".  The purpose is to propagate the IMAGE_PULL_POLICY down to the resources created from the templates.

address-space-controller/src/main/resources/templates/standard-space-infra.yaml
```diff
@@ -545,7 +543,7 @@ objects:
                 key: cookie-secret
                 optional: true
           image: ${AGENT_IMAGE}
-          imagePullPolicy: ${env.IMAGE_PULL_POLICY}
```
k-wall: I think your change here is incorrect.

k-wall:
@vbusch can you review please?

vbusch:
LGTM

PR #2926, opened 2019-06-21
Handle apiVersion correctly in upgrade tests

kornys:


kornys:
@enmasse-ci run tests profile=upgrade upgrade_from=0.27.0

PR #2936, opened 2019-06-25
Fix a possible NPE, having an empty spec section

ctron:
I think we should also backport this.


ctron:
For Go I think that the empty "spec" section will result in a non-null spec section, as there is no pointer but a struct. So that would align with this PR I think.

Of course we can also fix the rest of the Java code, to tolerate a `null` value.

But I think we should fix the current NPE, which isn't handled properly at all.

k-wall:
@ctron I think we should take the change as it stands for backport into 0.28.2 (minimising the risk), and have a bigger change on master separately.

PR #2946, opened 2019-06-27
Upd OLM csv

k-wall:
Update the olm clusterserviceversion
- Fix missing role (relates to changes made in #2857 but not reflected here)
- remove consoleservice from the example.  the consoleservice operator spontaneously creates the CR, so creating it explicity is supefluous.
- remove addressspaceschema from example. you can't create these, standard and brokered are baked in

### DIFF
--- a/templates/csv/enmasse.clusterserviceversion.yaml
+++ b/templates/csv/enmasse.clusterserviceversion.yaml
@@ -146,26 +146,10 @@ metadata:
           }
         },
         {
-          "apiVersion": "admin.enmasse.io/v1beta1",
-          "kind": "ConsoleService",
-          "metadata": {
-            "name": "console"
-          },
-          "spec": {}
-        },
-        {
           "apiVersion": "iot.enmasse.io/v1alpha1",
           "kind": "IoTConfig",
           "metadata": {
             "name": "default"
-          },
-          "spec": {}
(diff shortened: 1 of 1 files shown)


lulf:
@k-wall I believe the operator courier tool will fail if not all crds have examples associated with them . It’s a bit strange for addressspacesxhema I know

k-wall:
> @k-wall I believe the operator-courier tool will fail (at least it used to) if not all crds have examples associated with them . It’s a bit strange for addressspacesxhema I know

@lulf ah, I see.  there ought to be a way to mark the CR in such a way that it is allowed to be omitted.  It is confusing to see addressspaceschema there

lulf:
@k-wall agreed. Could be worth raising an issue for this with operator framework (https://github.com/operator-framework/operator-lifecycle-manager).

I’m wondering maybe we should look if there is a more kubernetes-natural way of providing the schema info without this read only global resource that I feel doesn’t fit...

k-wall:
@lulf with `operator-courier 2.1.2` I don't see an an error when I run verify if I omit `consoleservice` and  `addressspaceschema`.   I decided to make this PR just address the missing role, which I'd like to include 0.28.2.  Later for 0.29 I see it is possible to exclude the others again.

PR #2948, opened 2019-06-27
Go 1.12 and updated go types

secondsun:
Fixes https://github.com/EnMasseProject/enmasse/issues/2942

# Motivation
The AeroGear team is using enmasse/AMQ Online in an operator and our operator is written in go.  This PR updates the address object to have the Status field.

#How to test
Write a simple operator, go dep add the branch the PR is coming from, fetch an address and make sure its status matches the status of the object in k8s.

I couldn't find any integration tests in the project, but I did test the code locally with my operator.

Issue #2486, opened 2019-03-21
Building with Go 1.12 fails with "go tool vet" directly is unsupported; use "go vet"


k-wall:
As noted in Go's [release notes](https://golang.org/doc/go1.12#tools), go tool vet is no longer supported

```
cd /Users/keith/src/forks/enmasse/go/src/github.com/enmasseproject/enmasse && go tool vet cmd pkg
vet: invoking "go tool vet" directly is unsupported; use "go vet"
make[1]: *** [test_go_vet] Error 1
```

We want to keep compatibility with Go 1.10 and Go 1.11 so we need the build system to work with all three.

It appears with Go 1.12 the vet arguments are different.   It appear to scan for packages and finds system libraries.

```
cd /Users/keith/src/forks/enmasse/go/src/github.com/enmasseproject/enmasse && go vet cmd
# cmd/internal/bio
/usr/local/Cellar/go/1.12/libexec/src/cmd/internal/bio/buf.go:45:18: method Seek(offset int64, whence int) int64 should have signature Seek(int64, int) (int64, error)
/usr/local/Cellar/go/1.12/libexec/src/cmd/internal/bio/buf.go:57:18: method Seek(offset int64, whence int) int64 should have signature Seek(int64, int) (int64, error)
# cmd/link
/usr/local/Cellar/go/1.12/libexec/src/cmd/link/link_test.go:15:2: struct field tag "\n\tLondon. Michaelmas term lately over, and the Lord Chancellor sitting in Lincolnâs Inn Hall. Imp

```

ctron:
I can run `go vet` with Go 1.11.2 … And I think we could get away focusing on Go 1.11.x

jbtrystram:
I just ran into this as well : looks like Go 1.12 is default in fedora 30.

Also, Go vet is not happy with the `pkg` argument : 
```
[enmasse] cd /home/jbtrystram/go/src/github.com/enmasseproject/enmasse/go/src/github.com/enmasseproject/enmasse && go vet pkg
can't load package: package pkg: cannot find package "pkg" in any of:
	/usr/lib/golang/src/pkg (from $GOROOT)
	/home/jbtrystram/go/src/pkg (from $GOPATH)
```

secondsun:
@jbtrystram `go vet ./pkg/...` makes it happy

secondsun:
PR with fix : https://github.com/EnMasseProject/enmasse/pull/2951

PR #2951, opened 2019-06-28
Go 1.12 update

secondsun:
A breakup of https://github.com/EnMasseProject/enmasse/pull/2948/ that only includes the fix for https://github.com/EnMasseProject/enmasse/issues/2486

# How to Test 
Ensure you have go 1.12+ installed and run `make`.  Without this PR the vet tool will fail to run, with the vet tool it will run.


k-wall:
I think this change would mean the minimum Go version would change to 1.12.   See any problems with updating Go now, or any CI implications,  @ctron and @kornys?

This ought to be reflected in the HACKING.md file in the project's root.

ctron:
I think it is ok to move along to Go 1.12. However we should make that a conscious decision, as this also impacts the downstream release. As we do need Go 1.12 support there as well

But I think it makes sens doing the upgrade. And I think we could target that for 0.29. 

@kornys That would mean that we need Go 1.12 images for building.

kornys:
@ctron thb I don't know which version of ho is currently installed on CI images

secondsun:
@k-wall I just tried the command on go 1.10 and it works just fine.  This PR only makes the project buildable in go 1.12

ctron:
I think we should plan to migrate in the next version.

PR #2967, opened 2019-07-04
Use re-create deployment strategy for file based device registry

ctron:
This also includes a few formatting changes automatically applied by `go fmt`.

### DIFF
--- a/pkg/controller/iotconfig/file_device_registry.go
+++ b/pkg/controller/iotconfig/file_device_registry.go
@@ -49,7 +49,7 @@ func (r *ReconcileIoTConfig) processFileDeviceRegistry(ctx context.Context, conf
 
 	if !util.IsOpenshift() {
 		rc.ProcessSimple(func() error {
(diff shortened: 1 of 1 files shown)


ctron:

pkg/controller/iotconfig/file_device_registry.go
```diff
@@ -49,7 +49,7 @@ func (r *ReconcileIoTConfig) processFileDeviceRegistry(ctx context.Context, conf
 
 	if !util.IsOpenshift() {
 		rc.ProcessSimple(func() error {
-			return r.processService(ctx, nameDeviceRegistry + "-external", config, r.reconcileFileDeviceRegistryServiceExternal)
+			return r.processService(ctx, nameDeviceRegistry+"-external", config, r.reconcileFileDeviceRegistryServiceExternal)
```
kornys: why ?????

ctron: > This also includes a few formatting changes automatically applied by `go fmt`.

This is the way go formats its code. I would impose anything other than using `go fmt`.

PR #2972, opened 2019-07-07
Ensure subserv exits in timely fashion

k-wall:
I noticed that `subserv` fails to respond to SIGTERM so its pod hangs around for 30s afters Kubernetes/OpenShift shuts it down (presumably a timeout before killing it).  This change ensures that the Node process handles SIGTERM and closes down the process.  This should also have a positive affect on the length of the test run.


k-wall:
@enmasse-ci run tests testcase=io.enmasse.systemtest.standard.mqtt.**

PR #2973, opened 2019-07-07
Fix failing AuthenticationTest caused by vertx upgrade

k-wall:
New version of Vertx introduces a new exception `MechanismMismatchException`.  Handle that too.


k-wall:
@enmasse-ci run tests testcase=AuthenticationTest

PR #2987, opened 2019-07-11
Only sleep when needed to re-try

ctron:


ctron:

Issue #2953, opened 2019-06-30
Artemis runs as root when installed in openshift-operators (installed by OLM)


k-wall:
**Describe the bug**

If EnMasse 0.28.1 is installed in open shift-operators (as is the case when installing from OLM), the Broker runs as `root`.  Whilst this doesn't give a functional problem, running as root is against security best practice.  Also, the journal files etc become root owned.

When installed in another namespace in OpenShift 4.1 (for instance, enmasse-infra), the Broker process runs as a UID drawn from a range, assigned by SCC, I think).

**To Reproduce**
1. On OpenShift 4.1, install EnMasse 0.29.1 from either the Operator Hub, or direct from the bundle/ansible.
2. Create an address space and queue address so that a Broker is run.
3. `oc rsh`  into the Broker pod
4. Run `ps -ef`.  Observe that the Broker is running as root.

**Expected behavior**

Non root user should be executing the Broker.

**Additional context**

I notice that root is chosen because this is Dockerfile default.

k-wall:
I think this issue relates to https://docs.openshift.com/container-platform/4.1/authentication/managing-security-context-constraints.html.   Testing shows me that when EnMasse installed into openshift-operators the behaviour with respect to UID and the FS Group that owns the persistent volume is different.  I expect that this will somehow relate the SCC annotations on the openshift-operators namespace, but I don't understand these properly yet.

k-wall:
If https://github.com/operator-framework/operator-lifecycle-manager/issues/927 were to be resolved, this issue would be fixed.

PR #2991, opened 2019-07-11
Allow SecurityContext.fsGroup to be overridden for brokers, standard auth service etc

k-wall:
Downstream, components don't run as root (#2953).  The brokers, standard auth service, and IoT Device Registry fails when deployed to openshift-operators namespace (as happens under OLM) as the persistent storage is not owned by running user.  If this https://github.com/operator-framework/operator-lifecycle-manager/issues/927 is actioned, this workaround could be removed.


pkg/util/env.go
```diff
@@ -10,6 +10,13 @@ import (
 	"strings"
 )
 
+type ApplyEnvFn func(key string, value string, ok bool) error
+
+func ApplyEnv(key string, consumer ApplyEnvFn) error {
```
ctron: I know I am not very good at this myself, but for this method I wish we would have a comment, which explains how it works :grinning:

k-wall: Done :)

PR #3003, opened 2019-07-15
Create jenkins test only switch

obabec:


obabec:

PR #3005, opened 2019-07-15
Docs: issue-3002: Uninstall using Ansible 

jenmalloy:
and related install and uninstall doc fixes

### DIFF
--- a/documentation/assemblies/assembly-installing.adoc
+++ b/documentation/assemblies/assembly-installing.adoc
@@ -30,11 +30,11 @@ ifeval::["{cmdcli}" == "oc"]
 
 ifdef::OperatorSource[]
 {ProductName} can be installed by applying the YAML files using the OpenShift Container Platform command-line interface, by using
-the Operator Marketplace, or by running an link:https://www.ansible.com/[Ansible^] playbook.
+the Operator Marketplace, or by running the link:https://www.ansible.com/[Ansible^] playbook.
 endif::OperatorSource[]
 
 ifndef::OperatorSource[]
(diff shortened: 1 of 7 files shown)


documentation/modules/proc-uninstalling.adoc
```diff
@@ -18,7 +20,7 @@ ifeval::["{cmdcli}" == "oc"]
 endif::[]
 
 ifeval::["{cmdcli}" == "oc"]
-. Delete "cluster level" resources:
+. Delete the cluster-level resources:
```
jenmalloy: @k-wall, should this be "cluster-level custom resources"?

k-wall: @jenmalloy no, it is correct as it stands.

documentation/assemblies/assembly-installing.adoc
```diff
@@ -28,12 +29,12 @@ endif::[]
 ifeval::["{cmdcli}" == "oc"]
 
 ifdef::OperatorSource[]
-{ProductName} can be installed by applying the YAML files using the `oc` command-line tool, by using
-the Operator Marketplace, or by running the link:https://www.ansible.com/[Ansible^] playbooks.
+{ProductName} can be installed by applying the YAML files using the OpenShift Container Platform command-line interface, by using
+the Operator Marketplace, or by running an link:https://www.ansible.com/[Ansible^] playbook.
```
k-wall: I think 'the' was correct, we ship one playbook.

jenmalloy:
@k-wall, I changed "an Ansible playbook" to "the Ansible playbook" to address your comment, and left the other bit alone (per your comment)

PR #3007, opened 2019-07-16
Fix PR pipeline test_only switch

kornys:


kornys:

PR #3010, opened 2019-07-17
fix failing testSwitchAddressSpacePlan

famartinrh:
While investigating openshift 4 test failures I needed to fix this test


famartinrh:
@enmasse-ci run tests testcase=**.FirefoxGlobalConsoleTest#testSwitchAddressSpacePlan

PR #3017, opened 2019-07-18
 Shared address space plan resources decreased

obabec:


obabec:
@enmasse-ci run tests test_only

PR #3023, opened 2019-07-19
Increase timeout on receive MsgPatternsInternalTest

k-wall:
Previously the consumer was using a timeout of 200ms, which I think is too low in a stressed environment.  Also improved the logging for diagnostic purposes (Test change only).


k-wall:
@enmasse-ci run tests testcase=MsgPatternsInternalTest

PR #3029, opened 2019-07-22
Increase the timeout to 10sec. from the default 200ms

vbusch:
Increase the timeout to 10sec. from the default 200ms.  

To solve: org.opentest4j.AssertionFailedError: Expected 50 received messages ==> expected: <50> but was: <10> 


systemtests/src/test/java/io/enmasse/systemtest/bases/clients/ClientTestBase.java
```diff
@@ -293,6 +293,7 @@ protected void doDrainQueueTest(AbstractClient sender, AbstractClient receiver)
         arguments.put(ClientArgument.ADDRESS, dest.getSpec().getAddress());
         arguments.put(ClientArgument.COUNT, Integer.toString(expectedMsgCount));
         arguments.put(ClientArgument.MSG_CONTENT, "msg no. %d");
+        arguments.put(ClientArgument.TIMEOUT, "10");  // In seconds, maximum time the consumer waits for a single message
```
k-wall: I think since this is a receiver argument, it would be better moved down a few lines.

PR #3040, opened 2019-07-26
increase timeouts in smoke tests

famartinrh:
Increases some timeouts for the smoke tests because in openshift 4 test runs are slower


famartinrh:

PR #3041, opened 2019-07-26
collect logs in monitoring namespace in case MonitoringTest fails

famartinrh:
collects the logs of all pods in `enmasse-monitoring` namespace in case MonitoringTest fails


famartinrh:

PR #3042, opened 2019-07-30
Collect events from monitoring namespace if MonitoringTest fails

k-wall:
Also made `all_events log` capture a UTC timestamp too.


k-wall:

PR #3044, opened 2019-07-30
Wait longer for addressspace to be ready

k-wall:
Currently the test waits only 30s, which might be insuffiicent for the resources to spin-up and, due to the polling nature of the console, to detect the state change.


k-wall:
@enmasse-ci run tests testcase=GlobalConsoleTest

Issue #2918, opened 2019-06-19
Error installing monitoring


famartinrh:
**Describe the bug**
Following the openshift docs about installing monitoring to enmasse I encounter a repetitive error in the logs of `application-monitoring-operator` 

**To Reproduce**
On okd cluster with enmasse installed, following the docs https://enmasse.io/documentation/master/openshift/#monitoring-messaging
after deploying monitoring-operator I can see a constant and repetitive error in monitoring-operator logs

```
{"level":"info","ts":1560976847.1619184,"logger":"controller_applicationmonitoring","caller":"applicationmonitoring/applicationmonitoring_controller.go:104","msg":"Reconciling ApplicationMonitoring","Request.Namespace":"enmasse-monitoring","Request.Name":"middleware"}
{"level":"info","ts":1560976847.1620324,"logger":"controller_applicationmonitoring","caller":"applicationmonitoring/applicationmonitoring_controller.go:158","msg":"Phase: Create Prometheus CRs"}
{"level":"error","ts":1560976847.1918697,"logger":"kubebuilder.controller","caller":"controller/controller.go:209","msg":"Reconciler error","Controller":"applicationmonitoring-controller","Request":"enmasse-monitoring/middleware","error":"Error getting host from route: host value empty","errorVerbose":
[...]

PR #3055, opened 2019-07-31
Bounce the monitoring pods rather than fully deleting/recreating the whole namespace.

k-wall:
(I still see errors in an OCP4 environment with original test workaround.  Testing this an alternative approach #2918.)


k-wall:

k-wall:

k-wall:
This approach is ineffective, cancelling this PR.

PR #3061, opened 2019-08-01
Handle error when creating addresses fails

ctron:
Also, clean up a few things, reduce log output.


ctron:

PR #3066, opened 2019-08-01
Augment the server side provided firefox profile with devtools.console.stdout.content=true

k-wall:
Augment the existing firefox profile provided by the server rather than replacing it completely.


k-wall:

PR #3070, opened 2019-08-05
Docs issue-3049: Add example of how tenant can override properties fo…

jenmalloy:
…r external auth service


documentation/modules/ref-external-auth-service-example-allow-overrides.adoc
```diff
@@ -27,10 +27,10 @@ spec:
 ----
 <1> (Optional) The `realm` is passed in the authentication request. If not specified, an identifier in the form of _namespace-addressspace_ is used as the realm.
 <2> The host name of the external authentication service.
-<3> The port of the external authentication service.
+<3> The port number of the external authentication service.
 <4> (Optional) The CA certificate to trust when connecting to the authentication service.
-<5> (Optional) 
-<6> (Optional) Specifies whether overrides are allowed on the host name, port number, and CA certificate. Valid values are `true` or `false`.
+<5> (Optional) The client certificate to trust when connecting to the authentication service.
```
k-wall: <5>  The wording isn't good.   The client cert is the certificate that the client presents to the server in order to identify itself.

jenmalloy: @k-wall, Fixed in the last commit: "The certificate that the client presents when connecting to the authentication server."

jenmalloy:
@k-wall, Can you review/approve? Thanks!

PR #3073, opened 2019-08-06
Await API service registration before ending CommonTest#testMessagingDuringRestartComponents

k-wall:
Fixes unreliable test.


k-wall:
@enmasse-ci run tests testcase=CommonTest

PR #3077, opened 2019-08-07
Put the IoT tenant name in the status section

ctron:
Put the IoT tenant name in an explicit field in the status section.


ctron:

PR #3079, opened 2019-08-07
Drop code no longer required with Hono 1.0-M6

ctron:
This PR cleans up the Tenant Service to drop the deprecated APIs from Hono 1.0-M6, remove the unnecessary HTTP endpoint, and use the new, lighter, application startup.


ctron:

iot/iot-tenant-service/src/main/java/io/enmasse/iot/tenant/impl/TenantServiceImpl.java
```diff
@@ -131,4 +129,16 @@ public void get(final X500Principal subjectDn, final Span span,
                 CacheDirective.maxAgeDirective(this.configuration.getCacheTimeToLive().getSeconds()));
     }
 
+    @Override
+    public void get(String tenantId, Handler<AsyncResult<TenantResult<JsonObject>>> resultHandler) {
+        // TODO Auto-generated method stub
```
k-wall: Are the empty methods intentional?  If so, I think it is worthy of a comment.

ctron: Woohoo … no, they are not! Thanks for spotting this. Should be fixed now.

PR #3080, opened 2019-08-08
Design proposal: versioning/upgrade refactoring

k-wall:


documentation/design/proposals/versioning_upgrade_refactoring.adoc
```diff
@@ -0,0 +1,74 @@
...
+ ** For addressspaces, upgrade triggered by the fact that major.minor version on the plan disagrees with the major.minor
+    of the running software.   There is no way to trigger the upgrade path for a mirco increment.
+ ** For authenticationservices and consoleservice the code follows the OLM framework reconcile pattern.  For the upgrade
+    use-case this is functional but the code has no knowledge of the 'from' major/minor/mirco.  This may be an impediment
+    if we ever need to handle more sophisticated upgrades.
+ ** The addressspace's OAuth configuration becomes stale if the consoleservice is updated after the addressspace is created.   
+* No easy way for the user to determine major/minor/micro of the product being run.
+ ** The console displays a version - this is always an upstream version even in the downstream product.
+ ** No mechanism to determine the version from a command line/API.
+ ** (Downstream) No easy way to ensure that you are running a set of images from the same major/minor/micro release.
```
vbusch: Downstream images have the major/minor available, as JBOSS_IMAGE_VERSION=1.2 in the environment variables.  But no micro version.

documentation/design/proposals/versioning_upgrade_refactoring.adoc
```diff
@@ -0,0 +1,74 @@
...
+The CR's spec will carry the configuration items that are currently expressed as environment variables on the enmasse-operator, api-server, address-space-controller.   If the CR does not specify a configuration item, its default will be taken
+from the environment variables passed to the enmasse-operator itself.
+
+The CR will also carry the major/minor/micro version number of the version of EnMasse that created the record.  The status section will expose the current version.
+
+TODO: design the CRD
+
+== Image Streams
+
+Jens - can you elaborate?
```
vbusch: Image streams, will let us remove the repository name from the templates.  This will make it easier to test with images from different repos (docker.io, local, quay.io, etc.).
You need to define all your image streams.  And these will reference the repos.  But then the templates, just refer to the image stream by name.
Image streams are not supported in kubernetes.  And have only become available for statefulsets in openshift 3.9.

documentation/design/proposals/versioning_upgrade_refactoring.adoc
```diff
@@ -0,0 +1,74 @@
...
+
+To do this we would need to reimplement addressspace watching in the api-server. Or alternatively, we reimplement addressspaces as a CRD.   We would need the upgrade to promote the addressspace config maps into addressspace CR instances, but this might be simpler that implementing the watch proxy idea, and would mean we have much less code).
+
+== Images labeled with major/minor/micro product version
+
+Images will be labelled with major/minor/micro product version to ease identification.
+
+== Questions
+
+What version numbering scheme do we use downstream?
```
vbusch: Our documentation links are on a page titled _Product Documentation for Red Hat AMQ 7.4_
To help people find the correct version, our docs start with:

> RED HAT AMQ 7.4
> For use with AMQ Online 1.2

Is there an easier way to link our version to the AMQ 7.4 version?

documentation/design/proposals/versioning_upgrade_refactoring.adoc
ctron: As @vbusch already explained, we could drop the repository from the image names, at least for OpenShift. Image streams should also help with offline installations and might simplify the use of the terms based repository, as the actual pods don't need to have a service account with the pull secrets attached.

documentation/design/proposals/versioning_upgrade_refactoring.adoc
ctron: Technically we should be able to add both versions to the image, either via env-vars or through container labels. So the UI could simply pick up those values and show them.

lulf:
According to https://github.com/operator-framework/operator-lifecycle-manager/issues/999, The ability to configure deploymens in OLM manifest is coming: https://github.com/operator-framework/operator-lifecycle-manager/blob/master/Documentation/contributors/design-proposals/operator-config.md

documentation/design/proposals/versioning_upgrade_refactoring.adoc
lulf: Is it possible to deploy imagestreams via OLM?

documentation/design/proposals/versioning_upgrade_refactoring.adoc
lulf: I think both env-vars and container labels would be useful. Labels allow you to inspect version in the downstream catalog, and env-vars allow you to check version of a running container (not sure if you can read labels easily from within a container).

documentation/design/proposals/versioning_upgrade_refactoring.adoc
```diff
@@ -0,0 +1,74 @@
...
+
+Jens - can you elaborate?
+
+== Address Space Controller Refactor
+
+The address space controller currently uses templates to put down the resources required for standard or brokered addressspaces.  We will move the responsibilities for the address space controller to enmasse-operator and reimplement
+the logic following the OLM framework reconcile pattern.   This will also mean that changes in the consoleservice can be
+reflected on the address space deployments.
+
+Do we add version information to the addresspace object too and use this to trigger upgrades?
```
lulf: I think when we have fine grade reconciliation like this, we don't need to version the address space. If there are changes to i.e. the deployment, the operator will update it. Also, the operator will detect that image version is different from what it should be based on operator configuration, so it will also update the components automatically.

documentation/design/proposals/versioning_upgrade_refactoring.adoc
```diff
@@ -0,0 +1,74 @@
...
+
+To do this we would need to reimplement addressspace watching in the api-server. Or alternatively, we reimplement addressspaces as a CRD.   We would need the upgrade to promote the addressspace config maps into addressspace CR instances, but this might be simpler that implementing the watch proxy idea, and would mean we have much less code).
```
lulf: I think we should do this, migrate both AddressSpace and Address. As they are in the same API group, we cannot switch just one to be a CRD without a breaking API change.

documentation/design/proposals/versioning_upgrade_refactoring.adoc
```diff
@@ -0,0 +1,74 @@
...
+
+There will be an EnMasse CRD with an instance of this CR representing EnMasse itself.
+
+The enmasse-operator will watch its own namespace for the creation of  the enmasse CR.  On its appearence, it will take care
+of the deployment.  Conversely, if the CR is removed it will be removed.  This will go towards fixing #2863.
+
+The CR's spec will carry the configuration items that are currently expressed as environment variables on the enmasse-operator, api-server, address-space-controller.   If the CR does not specify a configuration item, its default will be taken
+from the environment variables passed to the enmasse-operator itself.
+
+The CR will also carry the major/minor/micro version number of the version of EnMasse that created the record.  The status section will expose the current version.
```
lulf: I think with OLM manifest gaining the ability to configure deployments, and migrating stuff to the enmasse-operator, the value of a toplevel resource like this is minimal IMO. It could function as a proxy for creating default instances of ConsoleService,AuthenticationService and IoTConfig for instance, they can also be created manually, so I'm not sure....

rgodfrey: The question is, how easy is this going to be to configure... are the permissions needed to install the operator going to be significantly higher than the rights needed to create/modify a CR with the configuration

lulf: So assuming api-server, address-space-controller is gone, then what is it that really needs to be configured? Only the parameters of the enmasse-operator process itself, which you can't configure via a CR, because the operator must already be deployed in order for the CR to be processed.

k-wall:
This proposal is made out of date by #3265, a new proposal can be raise in light of these changes.

Issue #3081, opened 2019-08-08
Agent's allocation changed log message spurious 


k-wall:
For diagnostic purposes, the agent emits log lines like the following and an address's broker allocation changes.  The intent was that the message is emitted when the address changes, but the implementation gives the opposite.

There is no known functional impact, but the noise clutters the logs.

`2019-08-06T12:08:35.459Z agent INFO  allocation changed for anycast command_response/xxxxxx376: undefined <-> undefined`

k-wall:
There was in fact a functional defect closely associated with this issue, see the PR for details.

PR #3083, opened 2019-08-09
Correct address allocation equality.

k-wall:
This defect meant that all existing addresses were considered modified even if they weren't.
This lead to unnecessary reconcilations against all routers and brokers.

### DIFF
--- a/agent/bin/agent.js
+++ b/agent/bin/agent.js
@@ -15,6 +15,7 @@
  */
 'use strict';
 
+var v8 = require('v8');
 var log = require("../lib/log.js").logger();
 var AddressSource = require('../lib/internal_address_source.js');
 var BrokerAddressSettings = require('../lib/broker_address_settings.js');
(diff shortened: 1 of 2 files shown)


k-wall:
During my test cases with large numbers of addresses and routers, the system came unresponsive and probes began to fail.  Node application heap used stayed reasonable, however.

k-wall:
Fix: #3081

k-wall:
@enmasse-ci run tests

agent/bin/agent.js
```diff
@@ -82,6 +83,10 @@ function start(env) {
                         exitHandler();
                     });
                 });
+
+                setInterval(() => {
+                    log.info("Heap statistics : %j", v8.getHeapStatistics());
+                }, 60000);
```
vbusch: Is printing the heap every minute, temporary?  And we can remove it once we know why it is getting so large?

k-wall: I'm inclined to keep it, though possible at a reduced frequency.  I'm not a node expert, but I couldn't find any external method to get the virtual machine's state at run time so was rather blind.  Having some heap statistics over time helps.

PR #3085, opened 2019-08-13
Turn down noisy per reconcilation per address logging

k-wall:
When processing thousands of addresses, the logs become very large owing to the "`Constructing config for queue<queue> allocated to: <broker>` that appears for every queue address for every router reconciliation.


k-wall:

PR #3087, opened 2019-08-13
Adapting IoT docs to the new registry API and doing some small imrpov…

dejanb:
…ements

### DIFF
--- a/documentation/modules/proc-iot-api.adoc
+++ b/documentation/modules/proc-iot-api.adoc
@@ -27,23 +27,12 @@ endif::[]
 
 .Procedure
 
-. In the Hono project, run:
-+
-[options="nowrap",subs="attributes"]
-----
-cd cli
-----
-. Run the following command at least once:
-+
-[options="nowrap",subs="attributes"]
-----
-mvn package -am
-----
+. Download the link:https://www.eclipse.org/hono/downloads/[Eclipse Hono command-line client].
 . Get the messaging endpoint certificate:
 +
 [options="nowrap",subs="attributes"]
 ----
-{cmdcli} -n myapp get addressspace iot -o jsonpath={.status.caCert} | base64 --decode > target/config/hono-demo-certs-jar/tls.crt
+{cmdcli} -n myapp get addressspace iot -o jsonpath={.status.caCert} | base64 --decode > tls.crt
 ----
 
 . Export the messaging endpoint host and port:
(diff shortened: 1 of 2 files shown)


documentation/modules/proc-iot-api.adoc
```diff
@@ -27,23 +27,12 @@ endif::[]
 
 .Procedure
 
-. In the Hono project, run:
-+
-[options="nowrap",subs="attributes"]
-----
-cd cli
-----
-. Run the following command at least once:
```
jenmalloy: What does "at least once" mean? Assuming this step is to install the command-line client, suggest changing to: "Run the following command to install the command-line client:"

documentation/modules/proc-iot-creating-device.adoc
```diff
@@ -21,17 +21,31 @@ endif::[]
...
-. Register a device:
+. Export the device registry host:
 +
 [options="nowrap",subs="attributes"]
 ----
-curl --insecure -X POST -i -H 'Content-Type: application/json' --data-binary '{"device-id": "4711"}' https://{registry}/registration/myapp.iot
+export REGISTRY_HOST={registry}
+----
+
+. Register a device with a defined id (in this example we're using `4711`):
```
jenmalloy:
```suggestion
. Register a device with a defined ID (in this example we're using `4711`):
```

documentation/modules/proc-iot-creating-device.adoc
```diff
@@ -21,17 +21,31 @@ endif::[]
...
 . Add credentials for a device:
 +
 [options="nowrap",subs="attributes"]
 ----
-curl --insecure -X POST -i -H 'Content-Type: application/json' --data-binary '{"device-id": "4711","type": "hashed-password","auth-id": "sensor1","secrets": [{"hash-function" : "sha-512","pwd-plain":"'hono-secret'"}]}' https://{registry}/credentials/myapp.iot
+curl --insecure -X PUT -i -H 'Content-Type: application/json' --data-binary '[{
+	"type": "hashed-password",
+	"auth-id": "sensor1",
+	"secrets": [{
+		"hash-function" : "sha-512",
```
ctron: IIRC the hash-function should not be present when putting in a plain password, as the server side decides the hash function.

ctron: Maybe also add an explanation about the concept of handling/storing hashed passwords.

documentation/modules/proc-iot-api.adoc
dejanb: This is now removed from the doc. It meant that you need to compile Hono command line source code at least once. With this PR we're downloading binary client, so no need for this anymore.

documentation/modules/proc-iot-creating-device.adoc
dejanb: Yeah, I'll add another PR that have more info around all topics. So it'll be covered there.

PR #3094, opened 2019-08-15
Correct address status equality check

k-wall:
The address status equality check was out of date.  It considered the `message` field rather than then `messages` array.  I saw a case during testing where the addresses stuck at isReady false, and this was the cause.

### DIFF
--- a/agent/lib/internal_address_source.js
+++ b/agent/lib/internal_address_source.js
@@ -64,6 +64,11 @@ function ready (addr) {
 }
 
 function same_allocation(a, b) {
+    if (a === b) {
+        return true;
+    } else if (a == null || b == null || a.length !== b.length) {
+        return false;
+    }
     for (var i in a) {
         var equal = false;
         for (var j in b) {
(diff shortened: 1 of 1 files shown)


agent/lib/internal_address_source.js
```diff
@@ -79,6 +79,22 @@ function same_allocation(a, b) {
...
 
+function same_messages(a, b) {
+    for (var i in a) {
+        var equal = false;
+        for (var j in b) {
+            if (a[i] === b[j]) {
+                equal = true;
+                break;
+            }
+        }
```
vbusch: Inner loop could be simplified with:
if (!b.includes(a[i])) {
  return false;
}

vbusch: Should ensure the 2 lists are the same length.

k-wall: Done.

k-wall: Yes, you are right.  Your comment made me realise that there was a further defect in `same_allocation`. ` same_allocation(a, a')` would have returned erroneously true where `a` is  allocated to 1, 2 and `a' `is allocated to 1,2,3. My change makes the implementation symmetric.

k-wall:

vbusch:
@enmasse-ci run tests testcase=StandardPersistentMessagesTest

PR #3099, opened 2019-08-16
Adjust node max_old_space_size to a percentage of available container memory

k-wall:
Previously, no account was taken on the container's memory.  This meant that the memory size assumed by node processes (agent, subserv) could be larger than the amount of memory allocated to the container.  This change adjusts the max_old_space_size to a percent (80% by default) of the container.


k-wall:

vbusch:
@enmasse-ci run tests testcase=StandardPersistentMessagesTest

Issue #3101, opened 2019-08-16
Agent seen to delete broker queues on restart


k-wall:
I've seen this issue a couple of times whilst test 0.26.x and master.   I had two queues defined myqueue1 and myqueue3 with a plan that required no sharding.  Admin pod was bounced. On start-up, agent deleted the two queues (confirmed by looking at the Artemis console), then a few moments later recreated them.

You can see from the log below that the Agent knew the addresses were assigned to broker-r26pbiyo1b-j5dv-0, but yet it still went on to the delete the queues.

```
Oslo:enmasse keith$ oc logs admin.r26pbiyo1b-59684b97d7-g8td2 -c agent -f
2019-08-08T00:06:03.134Z agent INFO  GET /apis/user.openshift.io => 200
2019-08-08T00:06:03.169Z agent INFO  GET
/oapi/v1/namespaces/enmasse-infra/routes/messaging-r26pbiyo1b => 403
2019-08-08T00:06:03.169Z agent INFO  could not retrieve messaging route hostname
2019-08-08T00:06:03.205Z agent INFO  GET
/api/v1/namespaces/enmasse-infra/configmaps?labelSelector=type%3Daddress-config%2CinfraUuid%3Dr26pbiyo1b
=> 200
2019-08-08T00:06:03.221Z agent INFO  GET
/api/v1/watch/namespaces/enmasse-infra/configmaps?labelSelector=type%3Daddress-config%2CinfraUuid%3Dr26pbiyo1b
=> 200
2019-08-08T00:06:03.308Z agent INFO  KWDEBUG addresses updated:
[...]

PR #3107, opened 2019-08-20
Ensure that addresses are not synced until addresses have been defined

lulf:
As proposed in #3101

### DIFF
--- a/agent/bin/agent.js
+++ b/agent/bin/agent.js
@@ -47,7 +47,6 @@ function start(env) {
             }
 
             server_promise.then(() => {
-                address_source.start();
                 if (env.ADDRESS_SPACE_TYPE === 'brokered') {
                     bind_event(address_source, 'addresses_defined', console_server.metrics);
                     console_server.listen_health(env);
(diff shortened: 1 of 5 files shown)


vbusch:
I like the waiting until the addresses are initializized.  A different different approach is to wait until the listeners are available.  See https://github.com/vbusch/enmasse/tree/init-watchers

lulf:
I've been able to verify this patch manually following the steps in #3101 

@vbusch I'm not sure what approach I prefer. If you would like to use the other approach, I'm happy to close this. In any case, it would be good to get a fix in so that we can start release builds...

grs:
@vbusch if the listeners are added in the same tick as the watcher is created, then I believe they can't miss the first event (which will have to be in the next tick at earliest). However you could also move the address_source.start() after the listeners are registered, which perhaps is more intuitively correct anyway.

vbusch:
@grs yes, I agree, that it would probably be impossible for the event to be processed in that tick.  Created https://github.com/lulf/enmasse/pull/3 as a precaution.

PR #3108, opened 2019-08-20
Release 0.26 agent fixes

lulf:


agent/bin/agent.js
```diff
@@ -15,6 +15,7 @@
  */
 'use strict';
 
+var v8 = require('v8');
```
vbusch: 0.26 uses node 6.1.  Will need to remove the v8.getHeapStatistics

vbusch:
Can the automated build still use docker?  If not, we will need https://github.com/EnMasseProject/enmasse/pull/2746

PR #3114, opened 2019-08-22
Improve api server address create performance

### LINKED ISSUE #3111: Performance of creating addresses is degraded after creating 7-8k addresses
This behavior occurs when roughly 7000-8000 addresses were already created in the address space before. After this amount of addresses the slowdown increases slightly with each request and eventually take 6-7 seconds to succeed or fails entirely with status 503.

lulf:
This change improves api server address create performance by almost an order of magnitude when 1000 addresses are defined.
The changes moves the validation of spec.address to standard-controller for the standard address space. For the brokered address space, the validation remains in the api-server, as it would require adding write-back capability to the agent. Future refactoring/consolidation of agent/standard-controller should incorporate this validation.

Fixes #3111


lulf:
@enmasse-ci run tests

PR #3117, opened 2019-08-22
Docs issue-3116: add doc for override of certs for external auth

jenmalloy:


documentation/modules/ref-address-space-example-external-auth-service-override.adoc
```diff
@@ -31,6 +31,10 @@ spec:
       realm: {ProductNamespace}-space-standard-auth
       host: standard-authservice-{ProductNamespace}.apps.wfd-28d9.openshiftworkshop.com
       port: 5671
+      caCertSecret:
+        name: standard-authservice-cert
```
jenmalloy: @robshelly, For better consistency, should the names in ref-external-auth-service-example-allow-overrides.adoc match the ones here (that is, change `my-ca-cert` and `my-client-cert` to `standard-authservice-cert`)?

robshelly: TBH I don't think it matters that much. In fact using `standard-authservice-cert` might even be somewhat confusing to some people given that it contains `standard` in it.

That's just the example I was able to get working, i.e. using a standard authservice deployed elsewhere as the server that external-authservice points to.

I you want to keep them the same, maybe change it to `my-ca-cert` and `my-client-cert` in both places

PR #3126, opened 2019-08-26
Add first edition of development guidelines

lulf:
This is an attempt to capture a set of development guidelines that should be followed when developing EnMasse. The idea is that PRs should be reviewed against these guidelines and ensure that we end up with a code base consistent with these guidelines.

This first edition is just an extraction of the current state from what is there in the project and also coloured by my subjective opinion. I'd like to get an agreement on the basics, and any additions or suggestions to changes are welcome.

Some sections are a bit short, so I'd love suggestions on what we could put in them.

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -22,7 +22,10 @@ If you need a function or class utility that requires adding a new dependency, f
 
 Java source code is built using Maven.
 
-Follow the [Oracle Java Code Conventions](https://www.oracle.com/technetwork/java/codeconvtoc-136057.html).
+Source code should follow the [Google Java Style Guide](https://google.github.io/styleguide/javaguide.html) with the following amendments:
+
+* Use 4 spaces indentation
+* Newlines should be used between annotations and function/class definitions
 
 Prefer to use plain Java whenever you can. Code should be easy to understand without intimate knowledge of specific frameworks. VertX, Jackson, JAX-RS and tests framework annotations such as Junit are in use at present, and should be preferred instead of introducing other frameworks. However, some dependencies like Eclipse Hono uses Spring, in which case it cannot be ignored.
 
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -36,8 +36,7 @@ Use `gofmt` to format all the source code.
 JavaScript dependencies are managed using `npm`, but it is invoked using maven and the
 `frontend-maven-plugin`.
 
-JavaScript code is indented using 4 spaces and most formatting rules used in Java applies in
-JavaScript as well with a few exceptions:
+JavaScript code generally follows [JavaScript Style Guide and Coding Conventions](https://www.w3schools.com/js/js_conventions.asp) with the most important being:
 
-* Variable names and functions use snake case
-* ...
+* 2 spaces indentation
+* Variable names and functions use camelcase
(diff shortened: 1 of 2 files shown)


DEVELOPMENT.md
```diff
@@ -0,0 +1,42 @@
+# Development Guidelines
+
+This document describes the development guidelines when writing and maintaining EnMasse source code. 
+
+## General considerations
+
+As EnMasse is a polyglot project with multiple languages in the same repo, and uses `make` as the
```
jbtrystram:
```suggestion
As EnMasse is a polyglot project with multiple languages in the same repo, it uses `make` as a
single entry point for invoking a full project build. Make will invoke the language-specific build 
```

DEVELOPMENT.md
```diff
@@ -0,0 +1,42 @@
...
+* If the cost of implementing and maintaining the functionality yourself is worthwhile in the same manner as above.
+
+## Java
+
+Java source code is built using Maven.
+
+Follow the [Oracle Java Code Conventions](https://www.oracle.com/technetwork/java/codeconvtoc-136057.html).
+
+Prefer to use plain Java whenever you can. Code should be easy to understand without intimate knowledge of specific frameworks. VertX, Jackson, JAX-RS and tests framework annotations such as Junit are in use at present, and should be preferred instead of introducing other frameworks. However, some dependencies like Eclipse Hono uses Spring, in which case it cannot be ignored.
+
```
jbtrystram: should we add a note about the java version we use ?

DEVELOPMENT.md
```diff
@@ -0,0 +1,43 @@
...
+* If the dependency has a reliable release schedule and track record for security updates.
+* If the dependency pulls in transitive dependencies that increase the footprint significantly
+* If you can copy the function or class (assuming the license allows it) into the EnMasse source tree. This should only be done for trivial amounts of code where the maintenance cost is expected to be low.
+* If the cost of implementing and maintaining the functionality yourself is worthwhile in the same manner as above.
+
+## Java
+
+Java source code is built using Maven.
+
+Follow the [Oracle Java Code Conventions](https://www.oracle.com/technetwork/java/codeconvtoc-136057.html).
```
ctron: Taking a look at the link:

>  The information on this page is for Archive Purposes Only
>
>This page is not being actively maintained. Links within the documentation may not work and the information itself may no longer be valid. The last revision to this document was made on April 20, 1999

Can't we come up with something better?

kornys:
@lulf maybe for java is better this convention https://google.github.io/styleguide/javaguide.html

DEVELOPMENT.md
lulf: Did you have anything specific in mind?

lulf:
@kornys sure, except that our code base uses 4 spaces and not 2. We could maybe add this as an exception to that style? 

The reason for linking to that oracle guide was because I think it’s what matches “IntelliJ default” the most. 

I’m fine referencing the google and noting the list of exceptions to that.

DEVELOPMENT.md
lulf: Actually on a second thought @jbtrystram , i wonder if this is better stated in HACKING.md. Otherwise we will have to remember to update this document everytime we change javascript, java or go version, which doesn't really change anything around the guidelines.

PR #3127, opened 2019-08-26
Remove jenkins pipelines

kornys:


kornys:

.jenkins/Jenkinsfile-pr
```diff
@@ -1,5 +1,6 @@
 #!/usr/bin/env groovy
-def lib
+
+@Library('amq-ci') _
```
lulf: What are the requirements on this library in order for a developer without access to the jenkins to be able to use this jenkinsfile?

kornys: @lulf only redhat developers are able to access our Library in amq-ci repo

kornys:

kornys:

.jenkins/Jenkinsfile-pr
lulf: @kornys theoretically, would it be possible to provide “spec” on what functions the library should have, and then someone could write a library implementing the expected functions( in order to run this without being red hat developer)? I’m asking as this could be a potential improvement later on, because this is one downside of this removal

kornys:

.jenkins/Jenkinsfile-pr
kornys: @lulf so do you mean something like have it upstream shared library for pr jenkinsfile?

lulf: @kornys I'm not saying we should provide anything at this point, but yeah, something like a 'lightweight' implementation of the 'amq-ci' "interface" if you will. I'm not sure how the @Library annotation works, is it just picking this up from Jenkins?

kornys: @lulf yeah global jenkins configuration load shared libraries and you can use it using @Library annotation, or you can load them dynamicaly using library('name')

PR #3134, opened 2019-08-28
Proposal on bridging to external endpoints

lulf:
Link for readability: https://github.com/EnMasseProject/enmasse/blob/16857d258e4776205c1705bfd9b6e1660286d72f/documentation/design/proposals/bridging_external.adoc

### DIFF
--- a/documentation/design/proposals/bridging_external.adoc
+++ b/documentation/design/proposals/bridging_external.adoc
@@ -29,6 +29,8 @@ spec:
   - name: remote1
     hostname: example.com
     port: 5672
+    failoverUrls:
+    - amqps://failover.example.com:5671 // On the form [(amqp|amqps|ws|wss)://]host_or_ip[:port]
     tls: # If not specified, do not enable TLS. Use just 'tls: {}' to enable TLS without any cert configuration.
       caCertSecret:
         name: remote1-secret # Secret containing CA to be trusted for this connector. NB! Secret must be readable by the system:serviceaccount:enmasse-infra:address-space-controller service account.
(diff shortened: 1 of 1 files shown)


documentation/design/proposals/bridging_external.adoc
```diff
@@ -0,0 +1,80 @@
...
+
+==== Api server
+
+The api server will need to perform an extra set of validation when creating address spaces to ensure that references from forwarder to connector are correct and that remoteAddress matches an address in the connector. The api-server should not require remoteAddress in forwarders to exist because it may not have been created yet.
+
+==== Address space controller
+
+The address space controller will need to pass additional configuration to the router when starting up. When creating the address space, the router config configmap need to be programatically crafted in order for the additional information to be optionally added to the config. The connector and forwarder definitions will essentially end up defining a set of connectors and autoLinks in the router config.
+
+This kind of fine-grained reconsiliation is a direction we're likely to go in when moving the address-space-controller logic to the controller-manager (enmasse-operator) in the future, but will create additional code that have to be rewritten if we implement this in the address-space-controller.
```
k-wall: reconciliation.

documentation/design/proposals/bridging_external.adoc
```diff
@@ -0,0 +1,80 @@
...
+To avoid that, as an alternative, we could put the responsibility of creating the router config configmap to the controller-manager as a first step in this direction (the configmap would then have to be removed from the template). From a maintenance POV it it might be better to have the logic in the same 'place' though, and the controller-manager would not be aware of any processing error in the address-space-controller such as failing to create the address space, in which case one would end up with a stale configmap.
+
+== Testing
+
+* Set up an AMQP broker (i.e. Artemis) on a remote host
+* Configure an address space with connectors and forwarders for the 4 different use cases listed initially
+
+== Documentation
+
+The addifional fields should be added to the address space reference, and the tenant guide should be extended to cover the use cases as a realistic scenario with examples.
```
k-wall: I think the existing authorisation scheme we have for address (as implemented in standard) should work with your proposal without change.  The user could express patterns for remote queues `remote1/foo*` or `remote1/foobar`

documentation/design/proposals/bridging_external.adoc
```diff
@@ -0,0 +1,80 @@
...
+The use cases covered by this proposal are:
+
+1. Forwarding messages from a local queue in a local address space to a destination on a remote AMQP endpoint
+1. Forwarding messages to a local queue in a local address space from a destination on a remote AMQP endpoint
+1. Sending messages to a remote AMQP endpoint via a local address space - by creating a connector and using prefixing
+1. Receiving messages from a remote AMQP endpoint via a local address space - by creating a connector and using prefixing
+
+== Design
+
+The address space specification would be expanded with 2 addition sub-objects, `connectors` and `forwarders`. A connector configures the infrastructure to connect to a remote AMQP endpoint and add address matching rules with the specified addresses. A forwarder specifies message routing rules to/from a local address to a remote address.
```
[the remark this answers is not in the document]
lulf: I'll add this comment: "The external bridging feature requires a dispatch router to support the above use cases. This means that this feature will only be available in the standard address space to start off with, but may be implemented in the brokered address space as well by adding a router to handle the bridging."

documentation/design/proposals/bridging_external.adoc
```diff
@@ -0,0 +1,80 @@
...
+
+1. Forwarding messages from a local queue in a local address space to a destination on a remote AMQP endpoint
+1. Forwarding messages to a local queue in a local address space from a destination on a remote AMQP endpoint
+1. Sending messages to a remote AMQP endpoint via a local address space - by creating a connector and using prefixing
+1. Receiving messages from a remote AMQP endpoint via a local address space - by creating a connector and using prefixing
+
+== Design
+
+The address space specification would be expanded with 2 addition sub-objects, `connectors` and `forwarders`. A connector configures the infrastructure to connect to a remote AMQP endpoint and add address matching rules with the specified addresses. A forwarder specifies message routing rules to/from a local address to a remote address.  
+
```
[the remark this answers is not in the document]
lulf: We have the capability to restrict outgoing connections via network policies today. Is that sufficient? Restricting which addresses to forward doesn't feel right to put in the infra configs somehow, since it does not have that knowledge.

documentation/design/proposals/bridging_external.adoc
```diff
@@ -0,0 +1,80 @@
...
+    addresses: # A list of remote addresses accessible via this address space. The addresses will be prefixed with the connector name (remote1/foo*). Addresses can be suffixed by a 'match all' wildcard.
+    - name: foo*
+  forwarders:
+  - name: fwd1
+    connector: remote1 # Reference to the connector used for the forwarding
+    localAddress: myqueue # Address within local address space
+    remoteAddress: foo/queue1 
+    direction: in | out # Specifies if messages are forwarded from localAddress to remoteAddress or from remoteAddress to localAddress.
+```
+
```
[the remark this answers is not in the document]
lulf: On making a separate CR for the `connector`: I am a bit reluctant to add another CR mainly because I think we already have too many. We would also be adding this to satisfy some persona we haven't yet clearly defined. However, maybe we can get both the simplicity and the required complexity if we allow the connectors array to either be a `spec` or a `ref` to a connector resource? I would then argue that we hold off adding the extra CR until a user ("customer") comes along with the use case.

On making a remote-address and forwarder address type: I'm not sure why these should be separate types, but instead something like:

```
kind: Address
spec:
  address: myqueue // Previously known as localAddress
  type: forwarded
  properties:
    direction: in
    remoteAddress: foo/queue1
```

Another question is how do we know which connector to validate this address against? Should a reference to the connector also be in the properties?

documentation/design/proposals/bridging_external.adoc
lulf: Yes, that is the intention. I'll add a note about it.

documentation/design/proposals/bridging_external.adoc
```diff
@@ -0,0 +1,80 @@
+== Overview
+
+A feature requested by several users is the ability to connect address spaces to external AMQP endpoints. NOTE: this proposal does not cover bridging between address spaces on the same cluster.
```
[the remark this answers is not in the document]
lulf: You are correct, its just another AMQP endpoint and nothing is stopping the user from connecting to it if it wants to. I think thats fine. What I wanted to emphasize with the comment was that the proposal to bridge address spaces on the same cluster would likely be different and simpler from user POV.

documentation/design/proposals/bridging_external.adoc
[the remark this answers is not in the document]
lulf: Document is updated and resolve after.

documentation/design/proposals/bridging_external.adoc
```diff
@@ -27,6 +29,8 @@ spec:
   - name: remote1
     hostname: example.com
     port: 5672
+    failoverUrls:
```
k-wall: As we now have the ability to express failover url(s), doesn't that make host and port redundant?

lulf: Yeah, I felt this was perhaps a bit weird. The reason i made it a url was to follow the router syntax which also does this. However, maybe we should change this to:

```
failoverHosts:
- hostname: failover.example.com
  port: 5672
```

Then the scheme will be restricted to the same as the original host (i.e. same tls settings).

Come to think of it, maybe change `hostname` to `host` both for failoverHosts and the connector `hostname` field?

k-wall: The point I was trying to make was that if we've got a list, then having separate hostname/port is redundant.  We could just enforce that the list has at least one entry.  Maybe `endpointHosts` would be a better name than `failoverHosts`.

I think host/port is better than hostname/port as there is then consistency with the external auth service.

lulf: Ah, that is a good point. And in terms of router configuration, we will use the first entry for the main connection.

grs:
For the forwarders, if there is a connector and autolink on all routers, you will have multiple forwarding links, which would mean the order would be lost. (Not that there are strong guarantees on ordering in standard space anyway of course, just pointing it out in case it was not obvious).

Did you consider separating the embedding of remote addresses into the local address-space from the main connector definition, having a new address type, 'remote', which could reference a connector? It would avoid having to prevent collisions between connector names and address definitions. It would also allow the user to define the prefix through which remote addresses could be accessed (so e.g. changing from a local queue to a remote queue would not require changing the actual address, just the type of the address). Just a suggestion though.

lulf:
Your comments are always appreciated, thanks @grs !

> For the forwarders, if there is a connector and autolink on all routers, you will have multiple forwarding links, which would mean the order would be lost. (Not that there are strong guarantees on ordering in standard space anyway of course, just pointing it out in case it was not obvious).

Yes, this is something we might want to make configurable at some point. In the first instance, I think its easiest to start of with configuring all the routers, but we should definitely consider a scheme where you can enforce ordering by configuring only 1 router to make the connection. I will update the proposal with a note about this.


> Did you consider separating the embedding of remote addresses into the local address-space from the main connector definition, having a new address type, 'remote', which could reference a connector? It would avoid having to prevent collisions between connector names and address definitions. It would also allow the user to define the prefix through which remote addresses could be accessed (so e.g. changing from a local queue to a remote queue would not require changing the actual address, just the type of the address). Just a suggestion though.

I think we discussed this briefly, but found explicitly referencing the connector from the address a bit odd, and that a 'remote' address is prefix so kind of a group of addresses, so we chose to avoid introducing an additional type. Conflicts between connector names and address definitions is something we can handle at address creation time (in the same way we would have to prevent conflicts between prefixes chosen by the remote addresses and local addresses). However I don't think this necessarily conflicts with the current proposal, and is perhaps something we can add later depending on user feedback.

What should have been stated more clearly in the design doc is that all remote addresses will be prefixed by the connector name, which we considered a 'good enough' granularity. I also wonder if we should change the connector field `addresses` to `prefixes` to make it clearer?

PR #3135, opened 2019-08-28
Clean up a few odd things in the IoTConfig controller

ctron:
* All adapter containers are now called "adapter".
* All code required to be executed per-adapter is now controlled by
  an array of adapters.
* A label was added to the device registry, indicating which type it
  actually is.
* A duplicate replica assignment was dropped.
* The default logback configuration no longer uses profiles which we
  didn't use in the first place.
* Pass enabled status to status section

### DIFF
--- a/pkg/controller/iotconfig/infinispan_device_registry.go
+++ b/pkg/controller/iotconfig/infinispan_device_registry.go
@@ -57,6 +57,7 @@ func (r *ReconcileIoTConfig) processInfinispanDeviceRegistry(ctx context.Context
 func (r *ReconcileIoTConfig) reconcileInfinispanDeviceRegistryDeployment(config *iotv1alpha1.IoTConfig, deployment *appsv1.Deployment) error {
 
 	install.ApplyDeploymentDefaults(deployment, "iot", deployment.Name)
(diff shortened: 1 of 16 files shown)


pkg/controller/iotconfig/infinispan_device_registry.go
```diff
@@ -57,6 +57,7 @@ func (r *ReconcileIoTConfig) processInfinispanDeviceRegistry(ctx context.Context
 func (r *ReconcileIoTConfig) reconcileInfinispanDeviceRegistryDeployment(config *iotv1alpha1.IoTConfig, deployment *appsv1.Deployment) error {
 
 	install.ApplyDeploymentDefaults(deployment, "iot", deployment.Name)
+	deployment.Annotations["iot.registry.type"] = "infinispan"
```
lulf: How about 'enmasse.io/iot-registry-type'? This matches the format we use for other annotation.

ctron: I am using `iot.enmasse.io` now, as this is the groupId of the resource. I also created a constant for this string.

ctron:

ctron:

PR #3141, opened 2019-08-29
Implement IoT device registry management API authorization

dejanb:


lulf:
@dejanb looks like an OK approach to me :)

dejanb:

dejanb:

jbtrystram:

dejanb:

dejanb:
@lulf @ctron squashed commits, this is now ready for merging IMHO

templates/iot/examples/iot-config-infinispan.yaml
```diff
@@ -19,3 +19,6 @@ spec:
       endpoint:
         secretNameStrategy:
           secretName: iot-mqtt-adapter-tls
+#  imageOverrides:
```
ctron: I am not sure we should keep this in here. However I think it would make sense having a proper "override" example.

pkg/controller/iotconfig/infinispan_device_registry.go
```diff
@@ -58,6 +58,7 @@ func (r *ReconcileIoTConfig) reconcileInfinispanDeviceRegistryDeployment(config
 
 	install.ApplyDeploymentDefaults(deployment, "iot", deployment.Name)
 	deployment.Annotations[RegistryTypeAnnotation] = "infinispan"
+	deployment.Spec.Template.Spec.ServiceAccountName = "iot-device-registry"
```
ctron: In order for this to work, we also do need to create, and set up the service account.

You have the YAML files in the templates folder, but we do require them as well in the CSV, for the operator manifest.

Then again, maybe we could let the IoT Config controller create it instead of using YAML files.

templates/iot/examples/iot-config-infinispan.yaml
dejanb: I introduced `iot-config-dev.yaml` for easier dev deployments.

pkg/controller/iotconfig/infinispan_device_registry.go
dejanb: I added resources to the csv for now. We can move this to the controller later on.

Issue #3133, opened 2019-08-28
enmasse-operator goes OOM in a system with many addresses


lulf:
**Describe the bug**
When creating a lot of addresses, the enmasse-operator will start crashing due to OOM. A working theory is that the operator is watching configmaps and not doing the appropriate filtering to avoid reading all address configmaps.

**To Reproduce**
1. Deploy EnMasse
2. Create ~5-10k anycast addresses
3. enmasse-operator pod should be getting restarts

**Expected behavior**
Operator should continue to run without crashes

lulf:
This is the cause I believe:

https://github.com/EnMasseProject/enmasse/blob/master/pkg/controller/consoleservice/consoleservice_controller.go#L115

A watch on all configmaps is set. Setting a label selector or a not-label selector should help

k-wall:
> This is the cause I believe:
> 
> https://github.com/EnMasseProject/enmasse/blob/master/pkg/controller/consoleservice/consoleservice_controller.go#L115
> 
> A watch on all configmaps is set. Setting a label selector or a not-label selector should help

Seems plausible.  This watch need only listen to configmaps that back address spaces, so if there was someway to filter by label that would be an improvement.  Initially I thought that passing a Predicate would help, but after reading https://github.com/operator-framework/operator-sdk/issues/1808 I'm not too sure.

PR #3145, opened 2019-09-02
Bump default limits of operator

lulf:
As a temporary "fix" for #3133, use default memory settings to increase
unlikelyness that the issue will be seen.


lulf:
@enmasse-ci run tests

PR #3148, opened 2019-09-03
Global Console phase 2 - HLD - address/connections and statistics.

k-wall:
Link to document for convenience:  https://github.com/EnMasseProject/enmasse/blob/global-console-phase-two/documentation/design/proposals/global-console-phase-2.adoc


k-wall:
Need to look at option for "bring your own prometheus".
Prometheus deployment with operator.

lulf:
> Maybe a third compromise option would be to configure prometheus with ephemeral storage to act as an in-memory fixed-window metrics store to provide simpler and also more efficient querying of the data. Restart etc. of the prometheus instance would then cause historical data to be lost, but it would make the operational aspect a lot simpler. The console would then get the 'consistent' view when scaled. With this scope, I would think being able to hook into an existing prometheus would not be necessary. We would have to agree on the window size, and put sufficient restrictions on prometheus to not exceed limits etc.

Just elaborating on this. The window could be configurable within some limits, and we would deduce the required ephemeral storage size based on it. The main reason for using some sort of ephemeral store would be to avoid caring about data migration, backup, corruption etc. It would need to be investigated what storage requirements is needed for ~1 day of metrics though.

k-wall:
@lulf I'm in favour of your third compromise (prometheus with an ephemeral storage default + fixed window).   That way we retain the ability to lean on prometheus's metrics computational abilities and we present the user with a single tap from which they can take an stream of metrics for digestion into their own metrics system (i.e. they can [federate](https://prometheus.io/docs/prometheus/latest/federation/) EnMasse's prometheus into their own).

documentation/design/proposals/global-console-phase-2.adoc
```diff
@@ -86,22 +86,25 @@ To signal a query across all namespaces, a * is used in the path name element.
 
 ==== exposing metrics
 
-The console-server will accept configuration that defines the statistics avaiable for each configured resource type.  This will take the form a list of named PromQL queries that are associated with each resource type (e.g. address).  These PromQL queries will ultimately produce the values that we wish to display on the console.  The idea is to avoid the need for the console UI to compute values itself.  
+The console-server will expose a endpoint which will proxy the to Prometheus PromQL api/v1/query API https://prometheus.io/docs/prometheus/latest/querying/api/.
```
lulf: Thinking from a monitoring perspective, can this endpoint easily be scraped by prometheus or does some custom code have to be written in order to fetch the metrics?

k-wall: Originally I had thought that the backend would associate a list of named queries with each resource type.  The UI could then request these named queries and use the results to populate the display.  The difficulty I faced was that I needed these queries to be a parameterised.   There seemed little point in introducing another API layer wrapping PromQL, so it seemed simpler to me to expose Prometheus itself as an authenticated proxy.

Orthogonal point - I was thinking that we'd just employ the prometheus operator to deploy the prometheus instance used by the console.  Is there a reason to disallow the user from interacting with this prometheus instance (and maybe grafana too) directly?

lulf: > Orthogonal point - I was thinking that we'd just employ the prometheus operator to deploy the prometheus instance used by the console. Is there a reason to disallow the user from interacting with this prometheus instance (and maybe grafana too) directly?

As long as its via a proxy that restricts which address spaces a user can get metrics from, I don't see a big problem with that. However, it does feel more natural to not expose prometheus directly as it is an implementation-specific part of the architecture.

I don't think there is a need to deploy Grafana?

k-wall: If we need to restrict access to metrics by address space, then I think the idea to expose PromQL doesn't work well.  The console-server would need to intercept the queries and ensure that no metric would be considered or returned that falls outside the address space(s) the user can see.   That implementation sounds unappealing.

k-wall: @lulf  I do have alternative idea.   At the moment, the console UI consumes the kubernetes addressspace API directly and the intention was to extend that for addresses and connections (latter kubernetes like).   As an alternative, what if we had the console-api exposed a different API for list/get operations.  The response would be an envelope combining  kubernetes object and the metrics.

 The console UI would still mutate using the existing kubernetes API but all reads would be done through this alternative.

The list of metrics included in each response envelope would be held in static configuration, with a list a separate list of PromQL queries held for each object type.

This means that the console-ui won't need to perform a separate operation to get the metrics, nor will it need to join the two results set together.

A addressspace list response would look like this:

```
{
  "items": [
    {
      "addresSpace": {
        "apiVersion": "enmasse.io/v1beta1",
        "kind": "AddressSpace",
        "metadata": {},
        "spec": {}
      },
      "metrics": [
        {
          "metricName": "enmasse_addresses",
          "prometheusMetricType": "guage",
          "prometheusMetricValue": 10
        },
        {
          "metricName": "enmasse_connections",
          "prometheusMetricType": "guage",
          "prometheusMetricValue": 10
        },
        {
          "metricName": "enmasse_messages_in_total",
          "prometheusMetricType": "counter",
          "prometheusMetricValue": 20
        }
      ]
    },
    {
      "addresSpace": {
        "apiVersion": "enmasse.io/v1beta1",
        "kind": "AddressSpace",
        "metadata": {},
        "spec": {}
      },
      "metrics": [
        {
          "metricName": "enmasse_addresses",
          "prometheusMetricType": "guage",
          "prometheusMetricValue": 12
        }
      ]
    }
  ]
}
```

in addition, there would be an API that would provide the metrics details for addresses and connections (i.e. the metrics for the senders/receivers).

lulf: Yeah I think that could work. I don't think there is a big reason why we need to expose the same API as Kubernetes. We should expose whatever makes sense for this use case. However, to not cause too much changes, maybe separating the read and write path is ok.

FWIW this is where I believe GraphQL would fit very well, giving the client the ability to select which parts of the response that should be returned. I.e. not returning metrics, or just a specific metric.

k-wall: My original inclination was to avoid wrapping/transforming the kubernetes objects.  I was driven by the Console's future need to support other CRDs types (Strimizi etc). Keeping the UI speaking Kubernetes API seemed like a good approach. However, that approach leads you to a separate API for querying metrics and the need for the UI to join the two results set.

documentation/design/proposals/global-console-phase-2.adoc
```diff
@@ -62,7 +62,26 @@ It will have the following responsibilities:
...
+The GraphQL interface will provide paginated result sets where necessary.  It will following the offset
+pagination pattern https://graphql.org/learn/pagination/ rather than the cursor one.  The cursor one doesn't meet
+the requirements of the UI.
+
+There will be the following root queries:
+
+```
+addressSpacePlans()                                                  # all defined addressSpacePlans
+addressPlans()                                                       # all defined addressPlans
+addressSpaces(first: Int, offset: Int, filter: String, sort: String) # all visible addressspaces across all viewable namesspaces, paginated
```
lulf: One of the references mention the cursor based pagination approach. AFAICT, the downside of offset based pagination is that it could lead to outdated views if the underlying result set changes. In our case, maybe that is not something we expect to happen, is that why the cursor based pagination is not used?

k-wall: The offset approach is driving by PF4 requirement design.  It allows you to jump a specific page.  If the client knows the page size and the the number of records, it can calculate the number of pages, and compute the first/offset required to jump to any of those pages.

lulf:
@k-wall Is this mergeable or do you wish to flesh out more ideas?

PR #3154, opened 2019-09-04
Iot tenant cleaner job

jbtrystram:
Initial simple app to trigger the deletion of a tenant resources in infinispan. 
I am going to need a bit of help for the test part :)

### DIFF
--- a/iot/iot-service-base/pom.xml
+++ b/iot/iot-service-base/pom.xml
@@ -46,6 +46,24 @@
       <version>${spring.boot.version}</version>
     </dependency>
 
+    <dependency>
+      <groupId>org.infinispan</groupId>
+      <artifactId>infinispan-client-hotrod</artifactId>
+      <optional>true</optional>
+    </dependency>
+
+    <dependency>
+      <groupId>org.infinispan</groupId>
+      <artifactId>infinispan-core</artifactId>
+      <optional>true</optional>
+    </dependency>
+
+    <dependency>
+      <groupId>org.infinispan</groupId>
+      <artifactId>infinispan-remote-query-client</artifactId>
+      <optional>true</optional>
+    </dependency>
+
     <!-- testing -->
     <dependency>
       <groupId>org.junit.jupiter</groupId>
--- a/iot/iot-service-base/src/main/java/io/enmasse/iot/service/base/infinispan/device/DeviceKey.java
+++ b/iot/iot-service-base/src/main/java/io/enmasse/iot/service/base/infinispan/device/DeviceKey.java
@@ -3,7 +3,7 @@
  * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
  */
 
-package io.enmasse.iot.registry.infinispan.device.data;
+package io.enmasse.iot.service.base.infinispan.device;
 
 import java.io.Serializable;
 import java.util.Objects;
@@ -14,7 +14,7 @@ import org.infinispan.protostream.annotations.ProtoMessage;
 import com.google.common.base.MoreObjects;
 import com.google.common.base.MoreObjects.ToStringHelper;
 
-import io.enmasse.iot.registry.infinispan.tenant.TenantHandle;
+import io.enmasse.iot.service.base.infinispan.tenant.TenantHandle;
 
 /**
  * A custom class to be used as key in the backend key-value storage.
--- a/iot/iot-tenant-cleaner/pom.xml
+++ b/iot/iot-tenant-cleaner/pom.xml
@@ -42,16 +42,6 @@
       <artifactId>infinispan-client-hotrod</artifactId>
       <optional>true</optional>
     </dependency>
-    <dependency>
-      <groupId>org.infinispan</groupId>
-      <artifactId>infinispan-core</artifactId>
-      <optional>true</optional>
-    </dependency>
-    <dependency>
-      <groupId>org.infinispan</groupId>
-      <artifactId>infinispan-remote-query-client</artifactId>
-      <optional>true</optional>
-    </dependency>
 
     <dependency>
       <groupId>ch.qos.logback</groupId>
--- a/iot/iot-tenant-cleaner/pom.xml
+++ b/iot/iot-tenant-cleaner/pom.xml
@@ -1,0 +1,100 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project
+  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd"
+  xmlns="http://maven.apache.org/POM/4.0.0"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>io.enmasse</groupId>
+    <artifactId>iot-application-base</artifactId>
+    <version>0.30-SNAPSHOT</version>
+    <relativePath>../iot-application-base</relativePath>
+  </parent>
+
+  <artifactId>iot-tenant-cleaner</artifactId>
+
+  <dependencies>
+    <dependency>
+      <groupId>io.vertx</groupId>
+      <artifactId>vertx-core</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>io.vertx</groupId>
+      <artifactId>vertx-config-yaml</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>io.vertx</groupId>
+      <artifactId>vertx-config</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>${project.groupId}</groupId>
+      <artifactId>iot-service-base</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+
+    <dependency>
+      <groupId>org.infinispan</groupId>
+      <artifactId>infinispan-query-dsl</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.infinispan</groupId>
+      <artifactId>infinispan-client-hotrod</artifactId>
+      <optional>true</optional>
+    </dependency>
+    <dependency>
+      <groupId>org.infinispan</groupId>
+      <artifactId>infinispan-core</artifactId>
+      <optional>true</optional>
+    </dependency>
+    <dependency>
+      <groupId>org.infinispan</groupId>
+      <artifactId>infinispan-remote-query-client</artifactId>
+      <optional>true</optional>
+    </dependency>
+
+    <dependency>
+      <groupId>ch.qos.logback</groupId>
+      <artifactId>logback-classic</artifactId>
+    </dependency>
+  </dependencies>
+
+  <build>
+    <plugins>
+      <plugin>
+        <artifactId>maven-compiler-plugin</artifactId>
+      </plugin>
+      <plugin>
+        <artifactId>maven-shade-plugin</artifactId>
+        <executions>
+          <execution>
+            <phase>package</phase>
+            <goals>
+              <goal>shade</goal>
+            </goals>
+            <configuration>
+              <transformers>
+                <transformer implementation="org.apache.maven.plugins.shade.resource.ApacheNoticeResourceTransformer"/>
+                <transformer implementation="org.apache.maven.plugins.shade.resource.ServicesResourceTransformer"/>
+                <transformer implementation="org.apache.maven.plugins.shade.resource.ManifestResourceTransformer">
+                  <mainClass>io.enmasse.iot.tools.cleanup.Application</mainClass>
+                </transformer>
+              </transformers>
+              <filters>
+                <filter>
+                  <artifact>*:*</artifact>
+                  <excludes>
+                    <exclude>META-INF/*.SF</exclude>
+                    <exclude>META-INF/*.DSA</exclude>
+                    <exclude>META-INF/*.RSA</exclude>
+                  </excludes>
+                </filter>
+              </filters>
+            </configuration>
+          </execution>
+        </executions>
+      </plugin>
+    </plugins>
+  </build>
+
+</project>
--- a/iot/iot-tenant-cleaner/src/main/java/io/enmasse/iot/tools/cleanup/Application.java
+++ b/iot/iot-tenant-cleaner/src/main/java/io/enmasse/iot/tools/cleanup/Application.java
@@ -1,0 +1,39 @@
+/*
+ *  Copyright 2019, EnMasse authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+
+package io.enmasse.iot.tools.cleanup;
+
+import io.vertx.core.Future;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class Application {
+
+    private static final Logger log = LoggerFactory.getLogger(Application.class);
+
+    public static void main(String args[]) {
+
+        if (args.length < 1){
+            log.error("The path to the config file must be given as the 1st argument.");
+            System.exit(1);
+        }
+
+        InfinispanTenantCleaner app = new InfinispanTenantCleaner(args[0]);
+
+        Future<Void> startPromise = app.configure().compose(config -> {
+                    Future<Void> runFuture = Future.future();
+                    return app.run(runFuture);
+                });
+
+         startPromise.setHandler(res -> {
+             if(res.succeeded()){
+                     System.exit(0);
+            } else {
+                log.error("Failure ",res.cause());
+                System.exit(1);
+            }
+        });
+    }
+}
--- a/iot/iot-tenant-cleaner/src/main/java/io/enmasse/iot/tools/cleanup/Application.java
+++ b/iot/iot-tenant-cleaner/src/main/java/io/enmasse/iot/tools/cleanup/Application.java
@@ -6,6 +6,7 @@
 package io.enmasse.iot.tools.cleanup;
 
 import io.vertx.core.Future;
+import java.nio.file.Path;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -15,12 +16,14 @@ public class Application {
 
     public static void main(String args[]) {
 
-        if (args.length < 1){
-            log.error("The path to the config file must be given as the 1st argument.");
-            System.exit(1);
+        Path configFile = null;
+
+        if (args.length == 1){
+            configFile = Path.of(args[0]);
+            log.debug("config file specified : "+configFile);
         }
 
-        InfinispanTenantCleaner app = new InfinispanTenantCleaner(args[0]);
+        InfinispanTenantCleaner app = new InfinispanTenantCleaner(configFile);
 
         Future<Void> startPromise = app.configure().compose(config -> {
                     Future<Void> runFuture = Future.future();
--- a/iot/iot-tenant-cleaner/src/main/java/io/enmasse/iot/tools/cleanup/CleanerConfigValues.java
+++ b/iot/iot-tenant-cleaner/src/main/java/io/enmasse/iot/tools/cleanup/CleanerConfigValues.java
@@ -1,0 +1,140 @@
+/*
+ *  Copyright 2019, EnMasse authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+
+package io.enmasse.iot.tools.cleanup;
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import io.enmasse.iot.service.base.infinispan.config.InfinispanProperties;
+
+import java.util.ArrayList;
+
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class CleanerConfigValues {
+
+    @JsonProperty("iotProject")
+    private String tenantId;
+
+    private int deletionChuckSize = 100000;
+
+    private String host;
+    private int port = 11222;
+    private String username;
+    private String password;
+
+    @JsonProperty("saslServerName")
+    private String saslServer;
+    private String saslRealm;
+
+    public String getTenantId() {
+        return tenantId;
+    }
+
+    public void setTenantId(String tenantId) {
+        this.tenantId = tenantId;
+    }
+
+    public String getHost() {
+        return host;
+    }
+
+    public void setHost(String host) {
+        this.host = host;
+    }
+
+    public int getPort() {
+        return port;
+    }
+
+    public void setPort(int port) {
+        this.port = port;
+    }
+
+    public String getUsername() {
+        return username;
+    }
+
+    public void setUsername(String username) {
+        this.username = username;
+    }
+
+    public String getPassword() {
+        return password;
+    }
+
+    public void setPassword(String password) {
+        this.password = password;
+    }
+
+    public String getSaslServer() {
+        return saslServer;
+    }
+
+    public void setSaslServer(String saslServer) {
+        this.saslServer = saslServer;
+    }
+
+    public String getSaslRealm() {
+        return saslRealm;
+    }
+
+    public void setSaslRealm(String saslRealm) {
+        this.saslRealm = saslRealm;
+    }
+
+    public int getDeletionChuckSize() {
+        return deletionChuckSize;
+    }
+
+    public void setDeletionChuckSize(int deletionChuckSize) {
+        this.deletionChuckSize = deletionChuckSize;
+    }
+
+    public String verify(){
+
+        ArrayList<String> missingValues = new ArrayList<>();
+
+        if (getTenantId() == null) {
+            missingValues.add("iotProject ");
+        }
+        if (getHost() == null) {
+            missingValues.add("host");
+        }
+        if (getUsername() == null) {
+            missingValues.add("username");
+        }
+        if (getPassword() == null) {
+            missingValues.add("password");
+        }
+        if (getSaslServer() == null) {
+            missingValues.add("saslServer");
+        }
+        if (getSaslRealm() == null) {
+            missingValues.add("saslRealm");
+        }
+
+        if (missingValues.size() != 0){
+            final String message = "Missing configuration value(s): ";
+            return message + String.join(", ", missingValues);
+        } else {
+            return null;
+        }
+    }
+
+    public static InfinispanProperties createInfinispanProperties(CleanerConfigValues config) {
+
+        InfinispanProperties infinispanProperties = new InfinispanProperties();
+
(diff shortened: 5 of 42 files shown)


iot/iot-service-base/pom.xml
```diff
@@ -46,6 +46,21 @@
       <version>${spring.boot.version}</version>
     </dependency>
 
+    <dependency>
+      <groupId>org.infinispan</groupId>
+      <artifactId>infinispan-client-hotrod</artifactId>
+    </dependency>
```
ctron: We shouldn't force this on all dependents. Could you mark them as "optional" here and then add them to the actual project.

iot/iot-service-base/src/main/java/io/enmasse/iot/service/base/infinispan/device/DeviceKey.java
```diff
@@ -41,7 +41,7 @@ protected DeviceKey() {}
      * @param tenantId the id of the tenant owning the registration key.
      * @param deviceId the id of the device being registered.
      */
-    private DeviceKey(final String tenantId, final String deviceId) {
+    public DeviceKey(final String tenantId, final String deviceId) {
```
ctron: The idea is to not use the constructor, but the `deviceKey` method, in order to do some checks before.

iot/iot-tenant-cleaner/pom.xml
```diff
@@ -0,0 +1,77 @@
...
+            <phase>package</phase>
+            <goals>
+              <goal>shade</goal>
+            </goals>
+            <configuration>
+              <transformers>
+                <transformer implementation="org.apache.maven.plugins.shade.resource.ApacheNoticeResourceTransformer"/>
+                <transformer implementation="org.apache.maven.plugins.shade.resource.ServicesResourceTransformer"/>
+                <transformer implementation="org.apache.maven.plugins.shade.resource.ManifestResourceTransformer">
+                  <mainClass>io.enmasse.iot.tenantCleaner.infinispan.Application</mainClass>
```
ctron: Please use a more standard package name. No camel case. e.g. `io.enmasse.iot.tools.cleanup.InfinispanTenantCleaner`.

iot/iot-tenant-cleaner/src/main/java/io/enmasse/iot/tenantCleaner/infinispan/Application.java
```diff
@@ -0,0 +1,182 @@
...
+
+        // Query and delete entries in devicesCache
+        QueryFactory queryFactory = Search.getQueryFactory(devicesCache);
+        Query query = queryFactory.from(DeviceInformation.class)
+                .select(Expression.property("deviceId"))
+                .having("tenantId")
+                .eq(TENANT_ID_TO_CLEAN)
+                .build();
+
+        List<String[]> matches = query.list();
```
ctron: This might cause you to load 1.000.000 device IDs into memory, I don't think it is a good idea.

iot/iot-tenant-cleaner/src/main/java/io/enmasse/iot/tenantCleaner/infinispan/Application.java
```diff
@@ -0,0 +1,182 @@
...
+        matches.forEach(deviceInfo -> {
+            System.out.println(deviceInfo[0]);
+            DeviceKey key = new DeviceKey(TENANT_ID_TO_CLEAN, deviceInfo[0]);
+            devicesCache.removeAsync(key);
```
ctron: Calling "async" without waiting for the operation to complete will put in on an internal queue. Having 1.000.000 devices, this would mean a task queue of 1.000.000 async operations, and probably cause some issues.

iot/iot-tenant-cleaner/src/main/java/io/enmasse/iot/tenantCleaner/infinispan/Application.java
```diff
@@ -0,0 +1,182 @@
...
+
+            DeviceConnectionKey conKey = new DeviceConnectionKey(TENANT_ID_TO_CLEAN, deviceInfo[0]);
+            devicesConnectionCache.removeAsync(conKey);
```
ctron: Same async issue here.

iot/iot-tenant-cleaner/src/main/java/io/enmasse/iot/tenantCleaner/infinispan/Application.java
```diff
@@ -0,0 +1,182 @@
...
+        });
+
+        // Stop the cache managers and release all resources
+        try {
+            deviceconProvider.stop();
```
ctron: Not waiting for the operations to complete, and stopping here will probably not work.

iot/iot-tenant-cleaner/src/main/java/io/enmasse/iot/tenantCleaner/infinispan/Application.java
```diff
@@ -0,0 +1,182 @@
...
+
+        Future configured = Future.future();
+        retriever.getConfig(ar -> {
+            if (ar.failed()) {
+                System.out.println(String.format("[ERROR] Failed to read configuration :{}", ar.cause().getMessage()));
+                configured.fail("cannot read config");
+            } else {
+                JsonObject config = ar.result();
+
+                TENANT_ID_TO_CLEAN = config.getString("iotProject");
```
ctron: Can't we deserialize that into a proper value object? This looks like bit too much glue code, and it isn't in any performance critical path.

systemtests/src/test/java/io/enmasse/systemtest/iot/registry/DeviceRegistryTestBase.java
```diff
@@ -341,6 +367,10 @@ private void checkCredentials(String authId, String password, boolean authFail)
         }
     }
 
+    private void removeTenant() throws Exception {
+        // ?? remove the iot project ?
```
ctron: yes

dejanb: Is the test working without actually removing tenant?

iot/iot-tenant-cleaner/src/main/java/io/enmasse/iot/tenantCleaner/infinispan/Application.java
```diff
@@ -0,0 +1,182 @@
...
+            System.err.println("[WARNING] Could not properly release cache manager resources in infinispan");
+        }
+    }
+
+    private void configure(){
+        ConfigStoreOptions yaml = new ConfigStoreOptions()
+                .setType("file")
+                .setFormat("yaml")
+                .setConfig(new JsonObject()
+                        .put("path", "/resources/config.yaml")
```
dejanb: We should make path configurable

iot/iot-tenant-cleaner/src/main/java/io/enmasse/iot/tenantCleaner/infinispan/Application.java
```diff
@@ -0,0 +1,182 @@
...
+                INFINISPAN_SASL_SERVER = configInfinispan.getString("saslServerName");
+                INFINISPAN_SASL_REALM = configInfinispan.getString("saslRealm");
+
+                verifyConfigurationValues();
+                configured.complete();
+            }
+        });
+
+        try {
+            configured.wait(10000);
```
dejanb: Instead of wait(), we should set handler for this future

iot/iot-tenant-cleaner/src/main/java/io/enmasse/iot/tenantCleaner/infinispan/Application.java
jbtrystram: I am looking into this, but does `query.list().stream()` really make a difference ?

jbtrystram: I did not find any way to get a stream of results with the query API, and the infinispan example [1] does the same thing

[1] : https://infinispan.org/tutorials/simple/streams/

iot/iot-tenant-cleaner/src/main/java/io/enmasse/iot/tenantCleaner/infinispan/Application.java
jbtrystram: Don't mind, it's very silly, I was investigating strange vertx behaviour

iot/iot-tenant-cleaner/src/main/java/io/enmasse/iot/tenantCleaner/infinispan/Application.java
ctron: I think that is something different … they are using streams on a cache. You can of course check, what the actual type at runtime is (the list). If that is something which proxies calls to the remote service, then that will work. Otherwise it we need to resolve it.

iot/iot-tenant-cleaner/src/main/java/io/enmasse/iot/tenantCleaner/infinispan/Application.java
jbtrystram: So make the path for the configuration file configurable ? Sounds like chicken and egg situation :)

iot/iot-tenant-cleaner/src/main/java/io/enmasse/iot/tools/cleanup/InfinispanTenantCleaner/Application.java
```diff
@@ -25,31 +26,36 @@
...
 
     private static String TENANT_ID_TO_CLEAN;
 
-    private static String INFINISPAN_HOST;
-    private static int INFINISPAN_PORT;
-    private static String INFINISPAN_USERNAME;
-    private static String INFINISPAN_PASSWORD;
-    private static String INFINISPAN_SASL_SERVER;
-    private static String INFINISPAN_SASL_REALM;
+    private String INFINISPAN_HOST;
```
ctron: non-static fields should be lowercase

iot/iot-tenant-cleaner/src/main/java/io/enmasse/iot/tools/cleanup/InfinispanTenantCleaner/Application.java
```diff
@@ -25,31 +26,36 @@
...
+    private String INFINISPAN_SASL_SERVER;
+    private String INFINISPAN_SASL_REALM;
 
     private static RemoteCache<DeviceKey, DeviceInformation> devicesCache;
     private static RemoteCache<DeviceConnectionKey, String> devicesConnectionCache;
 
     DeviceManagementCacheProvider mgmtProvider;
     DeviceConnectionCacheProvider deviceconProvider;
 
+    private static final Logger log = LoggerFactory.getLogger(Application.class.getName());
```
ctron: `Application.class` should be fine

iot/iot-tenant-cleaner/src/main/java/io/enmasse/iot/tools/cleanup/InfinispanTenantCleaner/Application.java
```diff
@@ -61,8 +67,8 @@ public void start(Future<Void> fut) {
             deviceconProvider.start();
             devicesConnectionCache = deviceconProvider.getDeviceStateCache();
         } catch (Exception e){
-            System.err.println("[ERROR] Unable to access Infinispan caches. "+ e.getMessage());
-            System.exit(1);
+            log.error("Unable to access Infinispan caches. {}", e);
```
ctron: You don't need a `{}` if the last argument is a `Throwable`.

iot/iot-device-registry-infinispan/pom.xml
```diff
@@ -123,6 +123,16 @@
         <configuration>
           <mainClass>io.enmasse.iot.registry.infinispan.InfinispanRegistry</mainClass>
         </configuration>
+        <executions>
```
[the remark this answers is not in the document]
jbtrystram: I think I can remove it, since I moved all the necessary dependencies in `iot-service-base` !

iot/iot-tenant-cleaner/src/main/java/io/enmasse/iot/tenantCleaner/infinispan/Application.java
jbtrystram: The path to the config file is now expected as a command line argument

iot/iot-tenant-cleaner/src/main/java/io/enmasse/iot/tools/cleanup/InfinispanTenantCleaner/InfinispanTenantCleaner.java
```diff
@@ -0,0 +1,181 @@
...
+import java.util.List;
+import org.infinispan.client.hotrod.RemoteCache;
+import org.infinispan.client.hotrod.Search;
+import org.infinispan.query.dsl.Expression;
+import org.infinispan.query.dsl.Query;
+import org.infinispan.query.dsl.QueryFactory;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class InfinispanTenantCleaner {
```
ctron: Please don't use `static`s in this class. I think everything should be a normal field.

iot/iot-tenant-cleaner/src/main/java/io/enmasse/iot/tools/cleanup/InfinispanTenantCleaner/InfinispanTenantCleaner.java
```diff
@@ -0,0 +1,181 @@
...
+            if (json.failed()) {
+                log.error("Failed to read configuration.", json.cause());
+                configured.fail(json.cause());
+            } else {
+                JsonObject config = json.result();
+
+                tenantIdToClean = config.getString("iotProject");
+
+                JsonObject configInfinispan = config.getJsonObject("infinispan", new JsonObject());
+                System.out.println(configInfinispan.encodePrettily());
```
ctron: Use slf4j

iot/iot-tenant-cleaner/src/main/java/io/enmasse/iot/tools/cleanup/InfinispanTenantCleaner/InfinispanTenantCleaner.java
```diff
@@ -0,0 +1,181 @@
...
+        retriever.getConfig(json -> {
+            if (json.failed()) {
+                log.error("Failed to read configuration.", json.cause());
+                configured.fail(json.cause());
+            } else {
+                JsonObject config = json.result();
+
+                tenantIdToClean = config.getString("iotProject");
+
+                JsonObject configInfinispan = config.getJsonObject("infinispan", new JsonObject());
```
ctron: Is there a chance for using  valueobject, instead of manually extracting information.

iot/iot-tenant-cleaner/src/main/java/io/enmasse/iot/tools/cleanup/InfinispanTenantCleaner/InfinispanTenantCleaner.java
```diff
@@ -0,0 +1,181 @@
...
+                System.out.println(configInfinispan.encodePrettily());
+                infinispanHost = configInfinispan.getString("host");
+                infinispanPort = configInfinispan.getInteger("port", 11222);
+                infinispanUsername = configInfinispan.getString("username");
+                infinispanPassword = configInfinispan.getString("password");
+                infinispanSaslServer = configInfinispan.getString("saslServerName");
+                infinispanSaslRealm = configInfinispan.getString("saslRealm");
+
+                try {
```
ctron: I don't understand why you use the future and the try/catch to handle errors as the same time. I think using a plain exception may be ok, if the method is ensured to not block.

iot/iot-tenant-cleaner/src/main/java/io/enmasse/iot/tools/cleanup/InfinispanTenantCleaner/Application.java
```diff
@@ -5,204 +5,32 @@
...
+        InfinispanTenantCleaner app = new InfinispanTenantCleaner(args[0]);
 
-        ConfigRetriever retriever = ConfigRetriever.create(vertx, new ConfigRetrieverOptions()
-                .addStore(new ConfigStoreOptions()
-                        .setType("file")
-                        .setFormat("yaml")
-                        .setConfig(new JsonObject().put("path", "/tmp/config.yaml"))));
-                //.addStore(new ConfigStoreOptions()
-                //.setType("env")));
+        app.configure().setHandler(ar ->{
```
ctron: I think that should be chained up using `compose()`, or something like that.

iot/iot-tenant-cleaner/src/main/java/io/enmasse/iot/tools/cleanup/InfinispanTenantCleaner/InfinispanTenantCleaner.java
jbtrystram: this is a mistake, I should have removed the try/catch when switching to using the futures.

iot/iot-tenant-cleaner/src/main/java/io/enmasse/iot/tools/cleanup/InfinispanTenantCleaner/InfinispanConfigValues.java
```diff
@@ -0,0 +1,115 @@
...
+        }
+        if (saslServer == null) {
+            missingValues.add("saslServer");
+        }
+        if (saslRealm == null) {
+            missingValues.add("saslRealm");
+        }
+
+        if (missingValues.size() != 0){
+            StringBuilder sb = new StringBuilder("Missing configuration value(s) : ");
```
ctron: You should be able to use `java.lang.String.join(CharSequence, Iterable<? extends CharSequence>)`instead

iot/iot-tenant-cleaner/src/main/java/io/enmasse/iot/tools/cleanup/CleanerConfigValues.java
```diff
@@ -0,0 +1,140 @@
...
+            missingValues.add("password");
+        }
+        if (getSaslServer() == null) {
+            missingValues.add("saslServer");
+        }
+        if (getSaslRealm() == null) {
+            missingValues.add("saslRealm");
+        }
+
+        if (missingValues.size() != 0){
```
ctron: It is always better to use "is empty", as it doesn't require counting the list elements. For an array list that may not make much difference, but I think it s worth always taking the same approach:
```suggestion
        if (!missingValues.isEmpty()){
```

iot/iot-tenant-cleaner/src/main/java/io/enmasse/iot/tools/cleanup/CleanerConfigValues.java
```diff
@@ -0,0 +1,140 @@
...
+        if (getPassword() == null) {
+            missingValues.add("password");
+        }
+        if (getSaslServer() == null) {
+            missingValues.add("saslServer");
+        }
+        if (getSaslRealm() == null) {
+            missingValues.add("saslRealm");
+        }
+
```
ctron: Verify that "deletion chunk" size is not negative

iot/iot-tenant-cleaner/src/main/java/io/enmasse/iot/tools/cleanup/Application.java
```diff
@@ -0,0 +1,39 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class Application {
+
+    private static final Logger log = LoggerFactory.getLogger(Application.class);
+
+    public static void main(String args[]) {
+
+        if (args.length < 1){
```
ctron: Do we require a config file? I though that we can also use env-vars, which would be much simpler in the Kubernetes environment. Otherwise we could need to manage a ConfigMap for each Job that we create.

iot/iot-tenant-cleaner/pom.xml
```diff
@@ -0,0 +1,100 @@
...
+    </dependency>
+
+    <dependency>
+      <groupId>org.infinispan</groupId>
+      <artifactId>infinispan-query-dsl</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.infinispan</groupId>
+      <artifactId>infinispan-client-hotrod</artifactId>
+      <optional>true</optional>
```
ctron: Is this still optional here?

iot/iot-service-base/pom.xml
```diff
@@ -46,6 +46,24 @@
       <version>${spring.boot.version}</version>
     </dependency>
 
+    <dependency>
+      <groupId>org.infinispan</groupId>
+      <artifactId>infinispan-client-hotrod</artifactId>
+      <optional>true</optional>
+    </dependency>
```
ctron: I am curios, does this module really have a dependency on Infinispan?

iot/iot-tenant-cleaner/src/main/java/io/enmasse/iot/tools/cleanup/InfinispanTenantCleaner.java
```diff
@@ -0,0 +1,148 @@
...
+    private RemoteCache<DeviceKey, DeviceInformation> devicesCache;
+    private RemoteCache<DeviceConnectionKey, String> devicesConnectionCache;
+
+    DeviceManagementCacheProvider mgmtProvider;
+    DeviceConnectionCacheProvider deviceconProvider;
+
+    private final Logger log = LoggerFactory.getLogger(InfinispanTenantCleaner.class);
+    private final Vertx vertx;
+
+    private final String pathToConfig;
```
ctron: Can we make it a `Path`. Same for the constructor.

iot/iot-tenant-cleaner/src/main/java/io/enmasse/iot/tools/cleanup/InfinispanTenantCleaner.java
```diff
@@ -0,0 +1,148 @@
...
+        try {
+            mgmtProvider = new DeviceManagementCacheProvider(infinispanProperties);
+            mgmtProvider.start();
+            devicesCache = mgmtProvider.getDeviceManagementCache();
+
+            deviceconProvider = new DeviceConnectionCacheProvider(infinispanProperties);
+            deviceconProvider.start();
+            devicesConnectionCache = deviceconProvider.getDeviceStateCache();
+        } catch (Exception e) {
+            log.error("Unable to access Infinispan caches.", e.getCause());
```
ctron:
```suggestion
            log.error("Unable to access Infinispan caches.", e);
```

iot/iot-tenant-cleaner/src/main/java/io/enmasse/iot/tools/cleanup/InfinispanTenantCleaner.java
```diff
@@ -0,0 +1,148 @@
...
+                .select(Expression.property("deviceId"))
+                .having("tenantId")
+                .eq(tenantIdToClean)
+                .maxResults(maxQueryResults);
+
+
+        List<Object[]> matches;
+        do {
+            matches = query
+                    //.startOffset(loopNumber.getAndIncrement()*maxQueryResults)
```
ctron: Do we need it?

iot/iot-tenant-cleaner/src/main/java/io/enmasse/iot/tools/cleanup/InfinispanTenantCleaner.java
```diff
@@ -0,0 +1,148 @@
...
+
+    private final String pathToConfig;
+
+    public InfinispanTenantCleaner(String pathToConfig) {
+        VertxOptions vxOptions = new VertxOptions().setBlockedThreadCheckInterval(200000);
+        vertx = Vertx.vertx(vxOptions);
+        this.pathToConfig = pathToConfig;
+    }
+
+    public Future<Void> run(Future<Void> startPromise) {
```
ctron: Most of the operations in this method will block. I think it is ok to implement this method classic, sync-style. And then invoke it with the "vertx blocking" wrapper. Otherwise it will only block the vertx main loop.

iot/iot-tenant-cleaner/src/main/java/io/enmasse/iot/tools/cleanup/InfinispanTenantCleaner.java
```diff
@@ -0,0 +1,148 @@
...
+    DeviceManagementCacheProvider mgmtProvider;
+    DeviceConnectionCacheProvider deviceconProvider;
+
+    private final Logger log = LoggerFactory.getLogger(InfinispanTenantCleaner.class);
+    private final Vertx vertx;
+
+    private final String pathToConfig;
+
+    public InfinispanTenantCleaner(String pathToConfig) {
+        VertxOptions vxOptions = new VertxOptions().setBlockedThreadCheckInterval(200000);
```
ctron: Remove this.

iot/iot-tenant-cleaner/src/main/java/io/enmasse/iot/tools/cleanup/Application.java
jbtrystram: env-vars are fine as well, I'll edit !

iot/iot-tenant-cleaner/src/main/java/io/enmasse/iot/tools/cleanup/InfinispanTenantCleaner.java
jbtrystram: did not know about that Class ! :)

iot/iot-service-base/pom.xml
jbtrystram: looks like no : 
```
+- org.infinispan:infinispan-client-hotrod:jar:9.4.15.Final:compile
[INFO] |  +- org.infinispan:infinispan-commons:jar:9.4.15.Final:compile
[INFO] |  |  \- org.jboss.logging:jboss-logging:jar:3.3.0.Final:compile
[INFO] |  +- org.jboss.marshalling:jboss-marshalling-osgi:jar:2.0.6.Final:compile
[INFO] |  \- org.jboss.spec.javax.transaction:jboss-transaction-api_1.2_spec:jar:1.1.1.Final:compile
```

iot/iot-tenant-cleaner/src/main/java/io/enmasse/iot/tools/cleanup/InfinispanTenantCleaner.java
jbtrystram: @ctron  like this ? https://github.com/EnMasseProject/enmasse/pull/3154/commits/f859fb0e97ddef3fb2b63d9fbe1ee28a00f2d84b

jbtrystram:
@ctron i am closing this since it's been replaced with #3438

PR #3155, opened 2019-09-04
Drop iot-gc in favor of finalizers

ctron:


ctron:

ctron:

pkg/controller/iotconfig/collector.go
```diff
@@ -8,46 +8,24 @@ package iotconfig
...
-		}
-
-		// set default resource limits
 
-		container.Resources = corev1.ResourceRequirements{
-			Limits: corev1.ResourceList{
-				corev1.ResourceMemory: *resource.NewQuantity(128*1024*1024 /* 128Mi */, resource.BinarySI),
-			},
-		}
+	// we no longer need the "iot-gc", so we only remove it
```
dejanb: Why do we need this? To deal with upgrades?

ctron: Yes. When yo perform an upgrade, it will remove the deployment.

PR #3166, opened 2019-09-09
Issue 3150 connector support

lulf:
Create router connectors and addresses based on address space connector specification.

* Reconcile secrets storing the connector certs
* Update router config based on connectors
* Add generation annotation to statefulset pod template to allow triggering a rolling update when the router config changes.
* Unit test for reconciling secrets and router config

@famartinrh With this PR, it should be possible to write tests to connect to remote brokers and send/receive messages via address space to them. Forwarding will come in another PR.

### DIFF
--- a/api-model/src/main/java/io/enmasse/address/model/StringOrSecretSelector.java
+++ b/api-model/src/main/java/io/enmasse/address/model/StringOrSecretSelector.java
@@ -47,7 +47,6 @@ public class StringOrSecretSelector extends AbstractWithAdditionalProperties {
     @Override
     public String toString() {
         return "StringOrSecretSelector{" +
-                "value='" + value + '\'' +
                 ", valueFromSecret=" + valueFromSecret +
                 '}';
     }
--- a/api-model/src/main/java/io/enmasse/address/model/StringOrSecretSelector.java
+++ b/api-model/src/main/java/io/enmasse/address/model/StringOrSecretSelector.java
@@ -1,0 +1,54 @@
+/*
+ * Copyright 2019, EnMasse authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+package io.enmasse.address.model;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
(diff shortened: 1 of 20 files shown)


lulf:

lulf:
@enmasse-ci run tests

lulf:

api-model/src/main/java/io/enmasse/address/model/StringOrSecretSelector.java
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    public void setValueFromSecret(SecretKeySelector valueFromSecret) {
+        this.valueFromSecret = valueFromSecret;
+    }
+
+    @Override
+    public String toString() {
+        return "StringOrSecretSelector{" +
+                "value='" + value + '\'' +
```
k-wall: Including value in toString risks logging a password.

k-wall:
Also a late comment on `io.enmasse.controller.RouterConfigController#generateConfig`, when processing the AddressSpaceSpecConnectorEndpoint, if AddressSpaceSpecConnectorEndpoint#port is 0, I think the code should assume the IANA registered port for AMQP (or AMQPS if TLS is enabled).

k-wall:
I've been having a play with the feature. I notice a couple of problems:

If I config a connector to refer to a host that does not exist (wibble:5672), I see the whole addresspace go into a unready state but the connector status reports ready:

```
Oslo:enmasse keith$ oc get addressspace -o wide
NAME       TYPE       PLAN             READY     PHASE     AGE       STATUS
standard   standard   standard-small   false     Failed    8m        The following stateful sets are not ready: [qdrouterd-ec7atf8b6o]
```

```
    connectorStatuses:
    - isReady: true
      messages: []
      name: qpid

```

The router is detecting the condition:

`2019-09-10 10:27:42.012767 +0000 SERVER (info) [C51] Connection to wiibble:5672 failed: proton:io Name or service not known - connect to  wiibble:5672`

I'd expect the addressspace overall to report ready true, whereas the connector's status should report the failure.

lulf:
@k-wall updated with port defaults and not leaking secret values. The status support will be raised in a separate PR.

lulf:

PR #3172, opened 2019-09-10
Add router status check for connectors

lulf:
(@k-wall currently based on #3166 but will rebase this once that is merged)

* Refactor and move router status collection into amqp-utils
* Add controller in address-space-controller for checking the router status by querying all routers for an address space looking for the connector and verifying its state.

### DIFF
--- a/address-space-controller/src/main/java/io/enmasse/controller/AddressSpaceController.java
+++ b/address-space-controller/src/main/java/io/enmasse/controller/AddressSpaceController.java
@@ -23,6 +23,7 @@ import io.enmasse.user.keycloak.KubeKeycloakFactory;
 import io.fabric8.kubernetes.api.model.ConfigMap;
 import io.fabric8.kubernetes.client.DefaultKubernetesClient;
 import io.fabric8.kubernetes.client.NamespacedKubernetesClient;
+import io.vertx.core.Vertx;
 import org.slf4j.Logger;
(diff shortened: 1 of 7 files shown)


lulf:
@enmasse-ci run tests

k-wall:
> Looks good. I think what we really desire is the the connector status reports something like "authentication fail" or "no route to host" for a router connector that's failing to make a remote connection, however, I interpret your comment "Until the connector entity allows querying for the status.." means that the router currently doesn't make that information is available, and instead we are having to infer the state of the connector from the absence or otherwise of a connection. I think a feature request against Dispatch Router would be a good idea. What do you think?

@grs WDYT?  Would it be a useful router enhancement to expose the connector's last error via AMQP management?

kornys:
@enmasse-ci run tests

k-wall:
> > Looks good. I think what we really desire is the the connector status reports something like "authentication fail" or "no route to host" for a router connector that's failing to make a remote connection, however, I interpret your comment "Until the connector entity allows querying for the status.." means that the router currently doesn't make that information is available, and instead we are having to infer the state of the connector from the absence or otherwise of a connection. I think a feature request against Dispatch Router would be a good idea. What do you think?
> 
> @grs WDYT? Would it be a useful router enhancement to expose the connector's last error via AMQP management?

https://issues.apache.org/jira/browse/DISPATCH-1419

PR #3189, opened 2019-09-13
Use simpler way of querying router

lulf:


lulf:

Issue #3150, opened 2019-09-03
Add support for external bridges


lulf:
**Overview**

A feature requested by several users is the ability to connect address spaces to external AMQP endpoints. NOTE: this proposal does not cover bridging between address spaces on the same cluster.

The use cases covered by this proposal are:

1. Forwarding messages from a local queue in a local address space to a destination on a remote AMQP endpoint
1. Forwarding messages to a local queue in a local address space from a destination on a remote AMQP endpoint
1. Sending messages to a remote AMQP endpoint via a local address space - by creating a connector and using prefixing
1. Receiving messages from a remote AMQP endpoint via a local address space - by creating a connector and using prefixing

The external bridging feature requires a dispatch router to support the above use cases. This means that this feature will only be available in the standard address space to start off with, but may be implemented in the brokered address space as well by adding a router to handle the bridging.

**Design**

[External Bridging Design Doc](https://github.com/EnMasseProject/enmasse/blob/master/documentation/design/proposals/bridging_external.adoc)

**Tasklist**

* [x] Dynamically create and update router configuration independently of the template in address-space-controller
* [x] Add connectors and address config to router config based on address space config
* [x] Update address space status based on connector status
* [x] Report connector metrics in address space controller prometheus endpoint
* [x] Validate connectors in api-server when creating address space
* [x] Validate forwarders in api-server when creating addresses
* [x] Create/update router linkroutes based on addresses in agent
* [x] Update address status based on forwarder status in standard-controller
* [x] Report forwarder metrics on standard controller prometheus endpoint

lulf:
@rgodfrey @k-wall 

Regarding bridging and forwarding, I believe we have 2 options for the implementation:

1. Message routing based forwarding (original proposal): Create a pair of autolinks to set up forwarding. 1 autolink that is configured on the remote connector, and 1 autoLink that is configured for the local queue/broker(s). This means that you have 1 link from each router to the remote endpoint, and 1 link per queue shard to one of the routers.

2. Link routing based forwarding: Create another connector from broker to router with source/target set to the remote address so that it is link-routed (already configured) to the external endpoint. This means you get only 1 link to the remote endpoint (or, 1 for each shard of a queue). We also know that link routing has better performance fwiw.

grs:
How does option 2. work? What triggers the link?

lulf:
(As we discussed in chat, just mentioning here for reference) In option 2, the broker creates the link using the remote address, thereby triggering the linkroute.

[...]

PR #3192, opened 2019-09-16
Add validation of model

lulf:
* Rename field names for statuses
* Validate connector names and address patterns
* Validate forwarder remoteAddress connector reference
* Validate forwarder address type

Issue #3150

### DIFF
--- a/api-model/src/main/java/io/enmasse/model/validation/StringOrSecretSelectorValidator.java
+++ b/api-model/src/main/java/io/enmasse/model/validation/StringOrSecretSelectorValidator.java
@@ -1,0 +1,25 @@
+/*
+ * Copyright 2019, EnMasse authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+
+package io.enmasse.model.validation;
+
+import io.enmasse.address.model.StringOrSecretSelector;
+
+import javax.validation.ConstraintValidator;
+import javax.validation.ConstraintValidatorContext;
+
+public class StringOrSecretSelectorValidator implements ConstraintValidator<StringOrSecretSelectorSpec, StringOrSecretSelector> {
+
+    @Override
+    public boolean isValid(StringOrSecretSelector selector, ConstraintValidatorContext context) {
+
+        if (selector == null) {
+            return true;
+        }
+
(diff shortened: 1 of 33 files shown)


famartinrh:
@lulf sorry you need to rebase because a reference to `getForwarderStatuses` was added to `o.enmasse.systemtest.utils.AddressUtils` apart from that LGTM

api-model/src/main/java/io/enmasse/model/validation/StringOrSecretSelectorValidator.java
```diff
@@ -0,0 +1,25 @@
...
+public class StringOrSecretSelectorValidator implements ConstraintValidator<StringOrSecretSelectorSpec, StringOrSecretSelector> {
+
+    @Override
+    public boolean isValid(StringOrSecretSelector selector, ConstraintValidatorContext context) {
+
+        if (selector == null) {
+            return true;
+        }
+
+        return (selector.getValue() == null || selector.getValueFromSecret() == null);
```
k-wall: Logically, should that be ^ (XOR) rather than || as we want one or the other, but not both?

lulf: Fixed.

api-server/src/main/java/io/enmasse/api/v1/http/HttpAddressServiceBase.java
```diff
@@ -170,11 +173,42 @@ static Address setAddressDefaults(String namespace, String addressSpace, Address
...
+            if (!Arrays.asList("queue", "subscription").contains(address.getSpec().getType())) {
+                throw new BadRequestException(String.format("Unable to create forwarders for address type '%s': Forwarders can only be created for address types 'queue' and 'subscription'", address.getSpec().getType()));
+            }
+
+            for (AddressSpecForwarder forwarder : address.getSpec().getForwarders()) {
+                boolean found = false;
+                for (AddressSpaceSpecConnector connector : addressSpace.getSpec().getConnectors()) {
+                    if (forwarder.getRemoteAddress().startsWith(connector.getName())) {
+                        found = true;
+                    }
```
k-wall: Might as well `break;`

lulf:
@k-wall fixed

PR #3193, opened 2019-09-16
Remove infraType label to prevent accidental delete

lulf:
Use ownerreferences instead to make sure address configmaps are deleted
when an address space configmap is deleted.

### DIFF
--- a/k8s-api/src/main/java/io/enmasse/k8s/api/ConfigMapAddressApi.java
+++ b/k8s-api/src/main/java/io/enmasse/k8s/api/ConfigMapAddressApi.java
@@ -54,11 +54,13 @@ public class ConfigMapAddressApi implements AddressApi, ListerWatcher<ConfigMap,
     private final WorkQueue<ConfigMap> cache = new EventCache<>(new HasMetadataFieldExtractor<>());
     private ObjectMapper mapper = new ObjectMapper();
     private final OwnerReference ownerReference;
+    private final String version;
 
-    public ConfigMapAddressApi(NamespacedKubernetesClient client, String infraUuid, OwnerReference ownerReference) {
+    public ConfigMapAddressApi(NamespacedKubernetesClient client, String infraUuid, OwnerReference ownerReference, String version) {
         this.client = client;
         this.infraUuid = infraUuid;
         this.ownerReference = ownerReference;
+        this.version = version;
     }
 
     @Override
@@ -193,6 +195,10 @@ public class ConfigMapAddressApi implements AddressApi, ListerWatcher<ConfigMap,
             builder.editMetadata()
                 .addToOwnerReferences(ownerReference)
                 .endMetadata();
+        }
+
+        if (version != null) {
+            address.putAnnotation(AnnotationKeys.VERSION, version);
         }
 
         if (address.getMetadata().getResourceVersion() != null) {
--- a/k8s-api/src/main/java/io/enmasse/k8s/api/ConfigMapAddressApi.java
+++ b/k8s-api/src/main/java/io/enmasse/k8s/api/ConfigMapAddressApi.java
@@ -16,6 +16,7 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
 
+import io.fabric8.kubernetes.api.model.OwnerReference;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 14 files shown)


k-wall:
Thinking about the upgrade between versions of enmasse, is it possible to retrospectively add the `NewOwnerReference` to existing address resources.

lulf:
@k-wall Good point, that should be doable. We anyway need the uid of the 'parent', so some more work needed still

k8s-api/src/main/java/io/enmasse/k8s/api/ConfigMapAddressApi.java
```diff
@@ -173,16 +173,23 @@ private String getConfigMapName(String namespace, String name) {
...
         ConfigMapBuilder builder = new ConfigMapBuilder()
                 .editOrNewMetadata()
                 .withName(getConfigMapName(address.getMetadata().getNamespace(), address.getMetadata().getName()))
                 .addToLabels(address.getMetadata().getLabels())
                 .addToLabels(LabelKeys.TYPE, "address-config")
                 .addToLabels(LabelKeys.INFRA_UUID, infraUuid)
-                .addToLabels(LabelKeys.INFRA_TYPE, "any")
+                .addNewOwnerReference()
+                .withApiVersion("v1")
+                .withKind("ConfigMap")
```
ctron: I think the "kind" and "apiVersion" should point to "AddressSpace" instead.

lulf:

k-wall:
Changes in themselves look good, however as I mentioned in another channel, I think this work should include an upgrader to set the owner references on existing addresses.  Approaches that minimise difference/favour uniformity tend to have fewer bugs in the long run.

I know we are talking about AddressSpace/Address/CRD refactoring, but I don't think this is contemplated for 0.30. I'm in favour of a upgrader for this change.

standard-controller/src/main/java/io/enmasse/controller/standard/StandardController.java
```diff
@@ -100,9 +102,23 @@ public void start() throws Exception {
                 .withUid(addressSpace.getMetadata().getUid())
                 .build();
 
+        ConfigMapAddressApi addressApi = new ConfigMapAddressApi(kubeClient, options.getInfraUuid(), ownerReference);
+
+        // Replace resources as part of upgrade if version is different
+        for (Address address : addressApi.listAddresses(options.getAddressSpaceNamespace())) {
+            if (!options.getVersion().equals(address.getAnnotation(AnnotationKeys.VERSION))) {
```
k-wall: What makes you drive this from version rather than checking for the expected owner reference (address.getMetadata().getOwnerReferences().contains(ownerReference))?

If driving from version is the way to go, doesn't io.enmasse.k8s.api.ConfigMapAddressApi#create need to set the version too?

lulf: @k-wall The ownerreference is set on the configmap and not on the Address object. This is because finalizers do not work with our custom resources at this point. So checking became a bit clunky, and I thought an annotation on the address would perhaps be desireable anyway.

You are right it should be set on the address create as well!.

I am also working on adding support to do the conversion in the agent as well (for the brokered address space).

lulf:
@k-wall I have added conversion of configmaps to both standard and brokered address space. For brokered, the logic for replacing is a bit different in standard, but in brokered agent will always try to update configmaps at startup, so updating the ownerref while doing that seems to work.

I have also made it so that version is always set if address is created/replaced both via api server and console.

lulf:

k-wall:
Looks good.. only one problem.  if a standard address space is upgraded, its addresses get the version annotation as part of the upgrade.  if a brokered address space is upgraded, the addresses don't get a version annotation.  They won't get a version annotation until the address is updated via the API (which may be never).  Having an address population where we have to remember that some addresses may not have a version will be tricky.

It occurs to me if `AddressSource.prototype.update_status` were change to the address API for updates, that would have the effect of setting ownerReference and the annotation version.  What is ugly about this idea is that the agent will be watching the CM and writing to the Address.

lulf:
@k-wall Good point... i've pushed a fix that should work for that that doesnt require that. TBH I think we end up with too much complicated additions to the code for doing this conversions which I feel is riskier than not adding the ownerref to existing entries.

lulf:

k-wall:
> @k-wall Good point... i've pushed a fix that should work for that that doesnt require that. TBH I think we end up with too much complicated additions to the code for doing this conversions which I feel is riskier than not adding the ownerref to existing entries.

Great - your change is better. (I didn't realise the node components already had a mechanism to get the version)

PR #3210, opened 2019-09-18
Provide endpoint status information

ctron:
Add endpoint status information to the IoTConfig. This way, you can easily get the endpoint information for all the services.


ctron:

PR #3211, opened 2019-09-18
Move and sort IoT packages

obabec:
* Create new packages
* Sort and move tests to new packages


ctron:

Issue #3191, opened 2019-09-16
Monitoring deployment does not work


kornys:
**Describe the bug**
io.enmasse.systemtest.isolated.monitoring.MonitoringTest.testAddressSpaceRules started failing on waiting for expected pods. I see that test follows enmasse doc.

**Additional context**
```
Error Message

Failed to wait for: Prometheus ready

Stacktrace

java.lang.IllegalStateException: Failed to wait for: Prometheus ready
	at io.enmasse.systemtest.isolated.monitoring.MonitoringTest.waitUntilPrometheusReady(MonitoringTest.java:87)
	at io.enmasse.systemtest.isolated.monitoring.MonitoringTest.installMonitoring(MonitoringTest.java:82)

Standard Output

2019-09-15T15:25:42.876Z &amp#27;[31mWARN &amp#27;[0;39m [TestWatcher] Before test exec
2019-09-15T15:25:42.877Z &amp#27;[34mINFO &amp#27;[0;39m [TestInfo] Testtag: isolated
2019-09-15T15:25:42.877Z &amp#27;[34mINFO &amp#27;[0;39m [TestInfo] Testtag: isolated-standard
2019-09-15T15:25:42.877Z &amp#27;[34mINFO &amp#27;[0;39m [TestInfo] Test is not shared!
2019-09-15T15:25:42.877Z &amp#27;[34mINFO &amp#27;[0;39m [TimeMeasuringSystem] Start time of operation TEST_EXECUTION is correctly stored
2019-09-15T15:25:42.877Z &amp#27;[34mINFO &amp#27;[0;39m [ITestSeparator] ####################################################################################################
2019-09-15T15:25:42.877Z &amp#27;[34mINFO &amp#27;[0;39m [ITestSeparator] io.enmasse.systemtest.isolated.monitoring.MonitoringTest.testAddressSpaceRules-STARTED
2019-09-15T15:25:42.877Z &amp#27;[34mINFO &amp#27;[0;39m [TestInfo] Testtag: isolated
2019-09-15T15:25:42.877Z &amp#27;[34mINFO &amp#27;[0;39m [TestInfo] Testtag: isolated-standard
2019-09-15T15:25:42.877Z &amp#27;[34mINFO &amp#27;[0;39m [TestInfo] Test is not shared!
2019-09-15T15:25:42.878Z &amp#27;[34mINFO &amp#27;[0;39m [Executor] Running command - oc new-project enmasse-monitoring
[...]

PR #3213, opened 2019-09-18
Various systemtest fixes

lulf:


lulf:
@kornys @famartinrh authservice test should be fixed too. Several of the others seems to fail on not enough persistent volumes. I'll await new test results with the PV fix before looking more into test failures.

kornys:
@lulf I guess we can merge it. Monitoring test failed because of https://github.com/EnMasseProject/enmasse/issues/3191

PR #3219, opened 2019-09-19
Cleanups for the managed strategy

ctron:


ctron:

ctron:

ctron:

ctron:

api-model/pom.xml
```diff
@@ -50,6 +50,11 @@
       <groupId>org.slf4j</groupId>
       <artifactId>slf4j-api</artifactId>
     </dependency>
+    <dependency>
+      <groupId>com.google.guava</groupId>
+      <artifactId>guava</artifactId>
+    </dependency>
+
```
lulf: I don't think its justified to add this dependency just for the toString helper methods.

templates/iot/api/010-CustomResourceDefinition-iot-project.yaml
```diff
@@ -32,9 +32,9 @@ spec:
     type: boolean
     description: If TLS is enabled
     JSONPath: .status.downstreamEndpoint.tls
-  - name: Ready
-    type: boolean
-    description: If the IoT project is ready
-    JSONPath: .status.isReady
```
lulf: Is it intentional to not keep isready? I think it could be useful.

ctron: Yes. I switch to useing "Phase", which seems to be the way Kubernetes does things. The problem with "ready" being a boolean was, that OLM didn't properly understand the boolean type.

It looks to me as if Kubernetes uses Phase/PhaseReason + Conditions to show that something is ready.

lulf: Good point. Should IsReady be removed from the model as well?

It could be worth standardizing on the 'conditions' array pattern for readiness instead, probably before API is declared beta.

lulf: Something like.... https://github.com/strimzi/strimzi-kafka-operator/blob/master/install/cluster-operator/040-Crd-kafka.yaml#L3538

ctron: > Good point. Should IsReady be removed from the model as well?
> 
> It could be worth standardizing on the 'conditions' array pattern for readiness instead, probably before API is declared beta.

Yes and no :) For example, for the "oc get" command it is easier to work with a non-array field. On the other hand conditions allow a more complex description of the current status. The API model currently uses the v1.Conditions from Kubernetes, so I think that aligns well with the Kubernetes world. I think having both Phase + Conditions is ok. Only having a third (boolean) seemed to be bit too much :)

api-model/pom.xml
ctron: Well, we have this dependency already. So I don't think is makes a difference in the final application.

lulf: Good point, that part is OK then. I do find it a bit strange to use this only for some classes and not for others though.

ctron:

systemtests/src/main/java/io/enmasse/systemtest/utils/AddressSpaceUtils.java
```diff
@@ -58,6 +59,10 @@ public static boolean existAddressSpace(String namespace, String addressSpaceNam
                 .map(addressSpace -> addressSpace.getMetadata().getName()).collect(Collectors.toList()).contains(addressSpaceName);
     }
 
+    public static BooleanSupplier adddressSpaceExists(final String namespace, final String name) {
```
famartinrh: typo adddress

systemtests/src/test/java/io/enmasse/systemtest/iot/project/ManagedTest.java
```diff
@@ -0,0 +1,120 @@
...
+import io.enmasse.systemtest.logs.CustomLogger;
+import io.enmasse.systemtest.utils.IoTUtils;
+import io.enmasse.user.model.v1.DoneableUser;
+import io.enmasse.user.model.v1.User;
+import io.enmasse.user.model.v1.UserList;
+import io.fabric8.kubernetes.client.dsl.MixedOperation;
+import io.fabric8.kubernetes.client.dsl.Resource;
+
+@Tag(TestTag.SHARED_IOT)
+public class ManagedTest extends AbstractIoTProjectTestBase  {
```
famartinrh: this class collides a bit with `IoTProjectManagedTest` , but it's fine, we have a refactor for iot tests in progress so we will take care of this too

systemtests/src/main/java/io/enmasse/systemtest/utils/AddressSpaceUtils.java
ctron: Fixed!

PR #3220, opened 2019-09-19
Avoid upgrade processing deleted broker clusters

k-wall:
This defect didn't produce a functional defect.


k-wall:

PR #3222, opened 2019-09-20
Turn off hostname verification on mqtt clients in failing tests

dejanb:


dejanb:

k-wall:
Why is it we need to do this now?  Does this relate to the Paho upgrade?

dejanb:
@k-wall yes, I believe before 1.2.1 hostname verification didn't work at all per https://github.com/eclipse/paho.mqtt.java/issues/506

kornys:
@lulf @dejanb can we merge it?

PR #3232, opened 2019-09-24
systemtests for connectors and forwarders with mutual tls 

famartinrh:


famartinrh:
@enmasse-ci run tests testcase=io.enmasse.systemtest.isolated.bridging.**

PR #3234, opened 2019-09-25
Configure global DLQ for rejected messages

lulf:


agent/lib/broker_controller.js
```diff
@@ -327,8 +327,7 @@ function translate(addresses_in, excluded_names, excluded_types) {
     for (var name in addresses_in) {
         if (excluded_names && excluded_names(name)) continue;
         var a = addresses_in[name];
-        if (excluded_types && excluded_types(a.type)) continue;
-        if (is_temp_queue(a)) {
+        if (excluded_types && excluded_types(a.type)) continue; if (is_temp_queue(a)) {
```
k-wall: Don't think you meant to remove this newline.

api-model/src/main/java/io/enmasse/model/validation/AddressNameValidator.java
```diff
@@ -51,6 +51,13 @@ public boolean isValid(final Address address, final ConstraintValidatorContext c
             }
         }
 
+        if (address.getSpec() != null && address.getSpec().getAddress() != null && address.getSpec().getAddress().startsWith("__")) {
```
k-wall: Don't you mean _!?

PR #3239, opened 2019-09-25
Fix skip cleanup

kornys:


kornys:
@lulf yup plans, infraconfigs, auth services are skipped

PR #3243, opened 2019-09-27
Set outcomes and default outcome for outgoing links

lulf:
### Type of change

_Select the type of your PR_

- Enhancement / new feature

### Description

Set supported outcomes and default outcome for outgoing connector sender links.

@k-wall I wasn't sure what the most appropriate default outcome would be. My thinking was that released would allow us to re-send.

### DIFF
--- a/broker-plugin/amqp-connector/src/main/java/org/apache/activemq/artemis/integration/amqp/LinkInitiator.java
+++ b/broker-plugin/amqp-connector/src/main/java/org/apache/activemq/artemis/integration/amqp/LinkInitiator.java
@@ -258,7 +258,6 @@ public class LinkInitiator implements EventHandler {
       Source source = new Source();
       source.setAddress(linkInfo.getSourceAddress());
       source.setDurable(TerminusDurability.UNSETTLED_STATE);
-      source.setDefaultOutcome(new Released());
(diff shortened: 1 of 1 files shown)


broker-plugin/amqp-connector/src/main/java/org/apache/activemq/artemis/integration/amqp/LinkInitiator.java
```diff
@@ -249,6 +258,8 @@ private void createSender(Session session) {
       Source source = new Source();
       source.setAddress(linkInfo.getSourceAddress());
       source.setDurable(TerminusDurability.UNSETTLED_STATE);
+      source.setDefaultOutcome(new Released());
```
k-wall: Whilst it is not wrong to set the default-outcome, since omitting it implies Accepted (i think) and this works for our use-case, I would do that.  I'm just fearful of bugs in less trodden paths of the implementations.

k-wall: In case my comment wasn't clear, I am simply suggesting removing line 261

lulf: Yes it was clear, just haven't gotten to it yet @k-wall  :)

PR #3253, opened 2019-09-30
Created broker tool

robshelly:
### Type of change

- Enhancement / new feature

### Description

Added broker tool to allow running commands on Artemis

### DIFF
--- a/amqp-utils/pom.xml
+++ b/amqp-utils/pom.xml
@@ -8,6 +8,10 @@
   </parent>
   <modelVersion>4.0.0</modelVersion>
   <artifactId>amqp-utils</artifactId>
+  <properties>
+    <maven.compiler.source>8</maven.compiler.source>
+    <maven.compiler.target>8</maven.compiler.target>
+  </properties>
   <dependencies>
     <dependency>
       <groupId>org.slf4j</groupId>
--- a/broker-plugin/broker-cli/pom.xml
+++ b/broker-plugin/broker-cli/pom.xml
@@ -1,0 +1,71 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <parent>
+    <groupId>io.enmasse</groupId>
+    <artifactId>broker-plugin</artifactId>
+    <version>0.30-SNAPSHOT</version>
+  </parent>
+  <modelVersion>4.0.0</modelVersion>
+  <artifactId>broker-cli</artifactId>
+
+  <dependencies>
+    <dependency>
+      <groupId>io.enmasse</groupId>
+      <artifactId>amqp-utils</artifactId>
+      <scope>compile</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.slf4j</groupId>
+      <artifactId>slf4j-api</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>ch.qos.logback</groupId>
+      <artifactId>logback-classic</artifactId>
+      <scope>runtime</scope>
+    </dependency>
+  </dependencies>
+
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-shade-plugin</artifactId>
+        <executions>
+          <execution>
+            <phase>package</phase>
+            <goals>
+              <goal>shade</goal>
+            </goals>
+            <configuration>
+              <transformers>
+                <transformer implementation="org.apache.maven.plugins.shade.resource.ServicesResourceTransformer"/>
+                <transformer implementation="org.apache.maven.plugins.shade.resource.ManifestResourceTransformer">
+                  <mainClass>io.enmasse.broker.cli.Cli</mainClass>
+                </transformer>
+              </transformers>
+              <filters>
+                <filter>
+                  <artifact>*:*</artifact>
+                  <excludes>
+                    <exclude>META-INF/*.SF</exclude>
+                    <exclude>META-INF/*.DSA</exclude>
+                    <exclude>META-INF/*.RSA</exclude>
+                  </excludes>
+                </filter>
+                <filter>
+                  <artifact>io.fabric8:kubernetes-model</artifact>
+                  <excludes>
+                    <exclude>io/fabric8/kubernetes/internal/KubernetesDeserializer**</exclude>
+                  </excludes>
+                </filter>
+              </filters>
+            </configuration>
+          </execution>
+        </executions>
+      </plugin>
+    </plugins>
+  </build>
+
+</project>
--- a/broker-plugin/broker-cli/src/main/java/io/enmasse/broker/cli/Cli.java
+++ b/broker-plugin/broker-cli/src/main/java/io/enmasse/broker/cli/Cli.java
@@ -1,0 +1,110 @@
+/*
+ * Copyright 2016-2019, EnMasse authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+
+package io.enmasse.broker.cli;
+
+
+import io.enmasse.amqp.Artemis;
+import io.enmasse.amqp.SyncRequestClient;
+import io.vertx.core.Vertx;
+import io.vertx.core.cli.CLI;
+import io.vertx.core.cli.CommandLine;
+import io.vertx.core.cli.Option;
+import io.vertx.core.cli.impl.DefaultCLI;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.Closeable;
+import java.util.Arrays;
+import java.util.concurrent.TimeoutException;
+
+public class Cli implements Closeable {
+
+    private static final Logger log = LoggerFactory.getLogger(Cli.class.getName());
+    private SyncRequestClient brokerClient;
+    private Artemis artemis;
+    private String command;
+    private String queue;
+
+
+    private boolean parserArgs(String[] args) {
+
+        CLI parser = new DefaultCLI().setName("broker-cli.jar");
+        parser.addOption(new Option()
+            .setLongName("command")
+            .setShortName("c")
+            .setRequired(true)
+            .addChoice("purgeQueue")
+            .addChoice("getMessageCount")
+            .setDescription("Command to run on broker"));
+        parser.addOption(new Option()
+            .setLongName("queue")
+            .setShortName("q")
+            .setRequired(true)
+            .setDescription("Queue to run command against"));
+
+        StringBuilder builder = new StringBuilder();
+        parser.usage(builder);
+
+        try {
+            CommandLine cl = parser.parse(Arrays.asList(args));
+            this.command = cl.getOptionValue("c");
+            this.queue = cl.getOptionValue("q");
+            return true;
+        } catch (Exception e) {
+            log.info("\n" + builder.toString());
+        }
+        return false;
+    }
+
+    private void start() throws Exception {
+
+        Vertx vertx = Vertx.vertx();
+        ClientFactory clientFactory = new ClientFactory(vertx);
+        brokerClient = clientFactory.connectBrokerManagementClient("localhost", 5673);
+        if (brokerClient == null) throw new IllegalStateException("Failed to connect to broker localhost:5673");
+        artemis = new Artemis(brokerClient);
+    }
+
+    private void run() throws TimeoutException {
+
+        switch (command) {
+            case "getMessageCount":
+                log.info("Count: " + artemis.getQueueMessageCount(queue));
+                break;
+            case "purgeQueue":
+                artemis.purgeQueue(queue);
+                break;
+            default:
+                throw new IllegalStateException();
+        }
+    }
+
+    public static void main(String[] args) {
+        try (Cli cli = new Cli()) {
+            if (cli.parserArgs(args)) {
+                cli.start();
+                cli.run();
+            } else {
+                System.exit(1);
+            }
+        } catch (Exception e) {
+            e.printStackTrace();
+            System.exit(1);
+        }
+        System.exit(0);
+    }
+
+    @Override
+    public void close() {
+        if (brokerClient != null) {
+            try {
+                this.brokerClient.close();
+            } catch (Exception e) {
+                e.printStackTrace();
+            }
+        }
+    }
+}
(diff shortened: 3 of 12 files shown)


broker-plugin/broker-cli/pom.xml
```diff
@@ -0,0 +1,72 @@
...
+  <dependencies>
+    <dependency>
+      <groupId>io.enmasse</groupId>
+      <artifactId>amqp-utils</artifactId>
+      <scope>compile</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.slf4j</groupId>
+      <artifactId>slf4j-api</artifactId>
+      <version>1.7.5</version>
```
k-wall: We use <dependencyManagement> section in the top level POM to globally control version numbers across the whole project.  You should remove the <version/> tag so you inherit from that.

broker-plugin/broker-cli/src/main/java/io/enmasse/broker/cli/Cli.java
```diff
@@ -0,0 +1,90 @@
...
+    private void start() {
+
+        Vertx vertx = Vertx.vertx();
+        ClientFactory clientFactory = new ClientFactory(vertx, certDir);
+
+        try {
+            brokerClient = clientFactory.connectBrokerManagementClient("localhost", 5673);
+            artemis = new Artemis(brokerClient);
+
+        } catch (NullPointerException e) {
```
lulf: This feels wrong. What is throwing the NPE?

broker-plugin/broker-cli/src/main/java/io/enmasse/broker/cli/Cli.java
```diff
@@ -0,0 +1,90 @@
...
+    private void closeClient() {
+        try {
+            brokerClient.close();
+        } catch (Exception e) {
+            log.error("Error closing client. " + e.toString());
+        }
+    }
+
+    private String executeCommandWithReturn(String command, String... queue) throws InvocationTargetException, IllegalAccessException, NoSuchMethodException {
+        log.info("Invoking Method with return");
```
lulf: Should be debug

broker-plugin/broker-cli/src/main/java/io/enmasse/broker/cli/Cli.java
```diff
@@ -0,0 +1,90 @@
...
+        String result ="";
+            Method method = artemis.getClass().getMethod(command, String.class);
+            log.info("Return type: " + method.getReturnType());
```
lulf: Should be debug

broker-plugin/broker-cli/src/main/java/io/enmasse/broker/cli/Cli.java
```diff
@@ -0,0 +1,90 @@
...
+        try {
+            result = String.valueOf(cli.executeCommandWithReturn(command, queue));
+
+            log.info("Result: " + result);
+        } catch (InvocationTargetException e) {
+            log.info("Error executing command: " + e.toString());
+        } catch (IllegalAccessException e) {
+            log.info("Error executing command: " + e.toString());
+        } catch (NoSuchMethodException e) {
+            log.info("Error executing command: " + e.toString());
```
lulf: Just catch all here

broker-plugin/plugin/src/main/sh/init-broker.sh
```diff
@@ -90,6 +90,8 @@ function pre_configuration() {
 
     cp $CONFIG_TEMPLATES/bootstrap.xml $BROKER_CONF_DIR/bootstrap.xml
     cp $CONFIG_TEMPLATES/jolokia-access.xml $BROKER_CONF_DIR/jolokia-access.xml
+    cp /opt/broker-plugin/lib/broker-cli.jar $BROKER_CUSTOM/bin/broker-cli.jar
+    cp -r /etc/enmasse-certs $BROKER_CUSTOM/enmasse-certs
```
lulf: Instead of copying the certs, you could point the script to the JKS trust and keystore generated for the broker and configure the ProtonClientOptions to point to that instead of the PEM.

broker-plugin/broker-cli/src/main/java/io/enmasse/broker/cli/Cli.java
k-wall: In general catching NPE is horrid!  If another part of the code called within this try/catch encounters a NPE, you've mislead the person investigating into looking for a create/connect failure.

Better to handle the connection failed separately.  I think you just want a if (brokerClient == null) { throw nice meaningful exception}

broker-plugin/broker-cli/src/main/java/io/enmasse/broker/cli/Cli.java
```diff
@@ -0,0 +1,90 @@
+/*
+ * Copyright 2016-2018, EnMasse authors.
```
k-wall: stale version year.

broker-plugin/broker-cli/src/main/java/io/enmasse/broker/cli/Cli.java
```diff
@@ -0,0 +1,90 @@
...
+            brokerClient.close();
+        } catch (Exception e) {
+            log.error("Error closing client. " + e.toString());
+        }
+    }
+
+    private String executeCommandWithReturn(String command, String... queue) throws InvocationTargetException, IllegalAccessException, NoSuchMethodException {
+        log.info("Invoking Method with return");
+        String result ="";
+            Method method = artemis.getClass().getMethod(command, String.class);
```
lulf: I thing we should do proper command line argument parsing and validation and invoke the methods explicitly rather than relying on reflection  in order to produce a more robust tool and reduce weird errors.

broker-plugin/broker-cli/src/main/java/io/enmasse/broker/cli/Cli.java
```diff
@@ -0,0 +1,90 @@
...
+                result = "success";
+            } else {
+                result = method.invoke(artemis, queue).toString();
+            }
+        return result;
+    }
+
+    public static void main(String[] args) {
+
+        Cli cli = new Cli();
```
k-wall: You could make Cli implement Closable so you could use try-with-resources pattern.

broker-plugin/broker-cli/src/main/java/io/enmasse/broker/cli/Cli.java
```diff
@@ -0,0 +1,90 @@
...
+            log.info("Result: " + result);
+        } catch (InvocationTargetException e) {
+            log.info("Error executing command: " + e.toString());
+        } catch (IllegalAccessException e) {
+            log.info("Error executing command: " + e.toString());
+        } catch (NoSuchMethodException e) {
+            log.info("Error executing command: " + e.toString());
+        }
+        cli.closeClient();
+        System.exit(0);
```
k-wall: Programs that are designed to be used on the command line should return a non-zero exit status in failure cases.

broker-plugin/broker-cli/pom.xml
```diff
@@ -0,0 +1,72 @@
...
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <parent>
+    <groupId>io.enmasse</groupId>
+    <artifactId>broker-plugin</artifactId>
+    <version>0.30-SNAPSHOT</version>
+  </parent>
+  <modelVersion>4.0.0</modelVersion>
+  <artifactId>broker-cli</artifactId>
```
k-wall: Are we intending to publish this Jar (maven-central)?

lulf: To keep things simple, I think we should only use it as part of the broker image for now.

k-wall: That's fine.. I just wanted to understand our intended direction.  I think the justification for publication would need to be strong, as it brings with it a lot of work.

broker-plugin/broker-cli/src/main/java/io/enmasse/broker/cli/Cli.java
```diff
@@ -0,0 +1,105 @@
...
+        }
+        return false;
+    }
+
+    private void start() throws Exception {
+
+        Vertx vertx = Vertx.vertx();
+        ClientFactory clientFactory = new ClientFactory(vertx);
+        brokerClient = clientFactory.connectBrokerManagementClient("localhost", 5673);
+        artemis = new Artemis(brokerClient);
```
k-wall: `io.enmasse.controller.standard.MutualTlsBrokerClientFactory#connectBrokerManagementClient` returns null if the connection doesn't successfully form.  You should handle this case explicitly rather than letting Artemis() class produce a NPE.

broker-plugin/broker-cli/src/main/java/io/enmasse/broker/cli/Cli.java
```diff
@@ -0,0 +1,105 @@
...
+    private void run() throws TimeoutException {
+
+        switch (command) {
+            case "getMessageCount":
+                log.info("Count: " + artemis.getQueueMessageCount(queue));
+                break;
+            case "purgeQueue":
+                artemis.purgeQueue(queue);
+                break;
+        }
```
k-wall: I see you've configured the cli parsing to allow only these possibilities, but I'd still add a default: label with a  throw IllegalStateException if command is somehow unexpected.

broker-plugin/broker-cli/src/main/java/io/enmasse/broker/cli/ClientFactory.java
```diff
@@ -0,0 +1,68 @@
...
+    public SyncRequestClient connectBrokerManagementClient(String host, int port) throws Exception {
+        ProtonRequestClient client = null;
+        try {
+            client = new ProtonRequestClient(vertx, System.getenv("HOSTNAME"));
+            CompletableFuture<Void> promise = new CompletableFuture<>();
+            client.connect(host, port, protonClientOptions, "activemq.management", promise);
+
+            promise.get(30, TimeUnit.SECONDS);
+        } catch (Exception e) {
+            log.error("Error creating/connecting to client: " + e.toString());
```
k-wall: Chain the exception to the error i.e. use log.error(String, Throwable)

broker-plugin/broker-cli/pom.xml
```diff
@@ -0,0 +1,71 @@
...
+
+  <dependencies>
+    <dependency>
+      <groupId>io.enmasse</groupId>
+      <artifactId>amqp-utils</artifactId>
```
k-wall: The JAR file for this is produced for Java 11. Is this going to be okay with Broker's JVM election?

amqp-utils/pom.xml
```diff
@@ -8,6 +8,10 @@
   </parent>
   <modelVersion>4.0.0</modelVersion>
   <artifactId>amqp-utils</artifactId>
+  <properties>
+    <maven.compiler.source>8</maven.compiler.source>
```
lulf: I think source can be left at 11?

broker-plugin/broker-cli/src/main/java/io/enmasse/broker/cli/Cli.java
```diff
@@ -0,0 +1,108 @@
...
+            .setShortName("Q")
+            .setRequired(true)
+            .setDescription("Queue to run command against"));
+
+        StringBuilder builder = new StringBuilder();
+        parser.usage(builder);
+
+        try {
+            CommandLine cl = parser.parse(Arrays.asList(args));
+            this.command = cl.getOptionValue("C");
```
lulf: I would personally prefer lowercase shortnames here.

broker-plugin/broker-cli/src/main/java/io/enmasse/broker/cli/Cli.java
```diff
@@ -0,0 +1,108 @@
...
+                throw new IllegalStateException();
+        }
+    }
+
+    public static void main(String[] args) {
+        try (Cli cli = new Cli()) {
+            if (cli.parserArgs(args)) {
+                cli.start();
+                cli.run();
+            }
```
lulf: I think if parserArgs returns false, it should exit with 1 instead of 0.

amqp-utils/pom.xml
k-wall: Probably yes.. it is not something I have ever needed to do, but it should work.

broker-plugin/broker-cli/src/main/java/io/enmasse/broker/cli/ClientFactory.java
```diff
@@ -0,0 +1,68 @@
...
+    public SyncRequestClient connectBrokerManagementClient(String host, int port) throws Exception {
+        ProtonRequestClient client = null;
+        try {
+            client = new ProtonRequestClient(vertx, System.getenv("HOSTNAME"));
+            CompletableFuture<Void> promise = new CompletableFuture<>();
+            client.connect(host, port, protonClientOptions, "activemq.management", promise);
+
+            promise.get(30, TimeUnit.SECONDS);
+        } catch (Exception e) {
+            log.error("Error creating/connecting to client: ", e.toString());
```
k-wall: Chain like so:
`log.error("Error creating/connecting to client: ", e);`

broker-plugin/broker-cli/src/main/java/io/enmasse/broker/cli/Cli.java
```diff
@@ -0,0 +1,110 @@
...
+
+        switch (command) {
+            case "getMessageCount":
+                log.info("Count: " + artemis.getQueueMessageCount(queue));
+                break;
+            case "purgeQueue":
+                artemis.purgeQueue(queue);
+                break;
+            default:
+                throw new IllegalStateException();
```
k-wall: How about:
`throw new IllegalStateException("Unrecognised command " + command);`

broker-plugin/broker-cli/src/main/java/io/enmasse/broker/cli/Cli.java
```diff
@@ -0,0 +1,110 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.Closeable;
+import java.util.Arrays;
+import java.util.concurrent.TimeoutException;
+
+public class Cli implements Closeable {
+
+    private static final Logger log = LoggerFactory.getLogger(Cli.class.getName());
```
k-wall: When using classnames for the logging hierarchy, this way is preferred.  The user can then precisely control the log statements emitted from different subsystems but adjusting the level associated with a whole branch,

private static final Logger log = LoggerFactory.getLogger(Cli.class);

broker-plugin/broker-cli/src/main/java/io/enmasse/broker/cli/Cli.java
lulf: Isn't IllegalArgumentException a better fit?

PR #3261, opened 2019-10-03
Build cleanup

lulf:
### Type of change

_Select the type of your PR_

- Refactoring

### Checklist

- [ ] Update documentation
- [ ] Update CHANGELOG.md


lulf:

templates/Makefile
```diff
@@ -169,6 +168,9 @@ install: ansible_install scripts_install component_install $(ENMASSE_BUNDLE) $(M
 package: prepare install
 	tar -czf build/$(INSTALLNAME).tgz -C build $(INSTALLNAME)
 
+deploy:
+	mvn deploy:deploy-file -Dfile=build/$(INSTALLNAME).tgz
```
k-wall: What is the idea of deploying this file to the remote repository?  Might this cause problems in some build environments?

templates/olm.enmasse.env
```diff
@@ -1,6 +1,6 @@
 # variables used in the olm yaml files
 
-VERSIONED_NAME=enmasse.${VERSION}.0
+VERSIONED_NAME=enmasse.${MAVEN_VERSION}
```
k-wall: As VERSIONED_NAME is used as the basis of the Operator's version number, don't we need to maintain a level of indirection to suit downstream builds?

lulf: There is already a downstream patch to set a different value for this field. This prevents the need to change an additional line in templates/csv/enmasse.package.yaml

templates/Makefile
lulf: The idea is that it contains the build 'zip'. I'm not sure if we're actually using it, so I can remove it.

PR #3266, opened 2019-10-04
[IoT] Create basic overview diagram

dejanb:
### Type of change
- Documentation

### Description

The birds-eye view of the IoT components. Note that this should just introduce concepts of IoT services, so not mentioning configs and projects.

I think we would need one more diagram, explaining telemetry and c&c concepts as well (before this one), like the one we have on a front page of https://www.eclipse.org/hono/


dejanb:
The link to the raw png

ctron:
This is definitely cool. I only think that we should add one more component, the "customer application". The "why".

ctron:
But you can also merge this, and do it in another PR!

PR #3276, opened 2019-10-07
improve testMessagingDuringRestartComponents

famartinrh:
### Description

this PR reimplements the test CommonTest#testMessagingDuringRestartComponents in order to consume less resources because we have seen cases in which this test fails because of the recursion used

### Checklist

- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/systemtests/src/test/java/io/enmasse/systemtest/isolated/CommonTest.java
+++ b/systemtests/src/test/java/io/enmasse/systemtest/isolated/CommonTest.java
@@ -29,16 +29,14 @@ import org.slf4j.Logger;
 import java.util.AbstractMap;
 import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.Collections;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.UUID;
 import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
-import java.util.concurrent.atomic.AtomicInteger;
 import java.util.stream.Collectors;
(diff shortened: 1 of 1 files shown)


famartinrh:
@enmasse-ci testcase=CommonTest#testMessagingDuringRestartComponents profile=isolated

famartinrh:
@enmasse-ci testcase=CommonTest#testMessagingDuringRestartComponents profile=isolated

famartinrh:
@enmasse-ci run tests testcase=CommonTest#testMessagingDuringRestartComponents profile=isolated

k-wall:
What's the case actually trying to test?    From the test name, I would have thought that the test ought to be testing the case that sending and receiving work both before and after the restart and that no message loss occurs.  If that's true, I don't think the code is actually achieving that safelly.  I think the code ought to be making sure at least one message is received before causing the restart, then after the restart completes, ensuring that at least one more message is sent and the same one received.

famartinrh:
I think the cases you described are covered by testRestartAdminComponent and testRestartComponents . With test testMessagingDuringRestartComponents I think we try to test a countinuous message sending while enmasse components are restarted(no routers or brokers) @k-wall

k-wall:
The point I was trying to make is there is nothing in the code (at least from a quick glance) that ensures messaging takes place before and after the restart.  The code should be written so there is at least one message sent/received before, then at least one send/receive after.

famartinrh:
I understand and I agree, it's easy to add that, new commit is coming

famartinrh:
@enmasse-ci run tests testcase=CommonTest#testMessagingDuringRestartComponents profile=isolated

PR #3284, opened 2019-10-07
Replace glide with go modules

### LINKED ISSUE #3283: Switch go dependency management to use go modules
**Overview**

Now that Kubernetes itself uses go modules for tracking dependencies, migrating enmasse to use go modules as well should be possible. 

This should simplify dependency managed for Go code.

**Design**

Convert glide.yaml and glide.lock to go modules.

lulf:
Update vendors as well.

Fixes #3283


lulf:

lulf:
I'll do some more testing before creating the pr...

PR #3300, opened 2019-10-11
Run router status collection in a separate thread

### LINKED ISSUE #3299: Router status check blocks reconiliation loop
**Describe the bug**
The router status check is run on every address change. In a system with a large number of addresses where new addresses are added, the status check can take up to 10 seconds. This ends up causing a long time between create and ready to use for new addresses.

**Expected behavior**
Reconciliation loop should not be blocked by status check

lulf:
### Type of change

_Select the type of your PR_

- Bugfix

### Description

This change spins of router status collection into a separate thread. This prevents the status collection to block the reconciliation loop and should speed up the time between address creation and addresses being ready.

Fixes #3299

### Checklist

- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/k8s-api/src/main/java/io/enmasse/k8s/api/ResourceChecker.java
+++ b/k8s-api/src/main/java/io/enmasse/k8s/api/ResourceChecker.java
@@ -29,6 +29,8 @@ public class ResourceChecker<T> implements CacheWatcher<T>, Runnable {
     public void start() {
         running = true;
         thread = new Thread(this);
+        thread.setName("resource-checker");
+        thread.setDaemon(true);
         thread.start();
 
     }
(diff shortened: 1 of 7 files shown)


lulf:

lulf:

lulf:

k-wall:
One further idea - probably a separate PR - how about parallelising io.enmasse.controller.standard.RouterStatusCollector#collect calls so that multiple routers can be queried in parallel?  io.enmasse.controller.standard.RouterStatusCollector#collect would return a future, which would be awaited.    It would also be nice if RSC used a persistent router collection, so avoid so much noise in the router logs.

lulf:
@k-wall I'll raise a separate PR for the parallelisation of collect if you don't mind.

PR #3320, opened 2019-10-15
Implemented purge functionality

vbusch:
-Removed the inactive suspend menu item.
- Hooked up the purge menu item.

### Type of change

_Select the type of your PR_

- Bugfix
- Enhancement / new feature
- Refactoring
- Documentation

### Description

_Please describe your pull request_

### Checklist

- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/agent/lib/queue.js
+++ b/agent/lib/queue.js
@@ -96,17 +96,25 @@ function purge_brokered_queue(queueName) {
                 return pod_watcher.get_pod_definition(pod);
             });
 
-            var brokers = create_brokers(podDefs, 'brokered');
-            var results = brokers.map(function (broker) {
-                return broker.purgeQueue(queueName);
-            });
+            var broker_cons = create_brokers(podDefs, 'brokered');
+            var purgedQueues = broker_cons.map(broker => broker.purgeQueue(queueName));
 
             //for the brokered address space, there will never be more then 1 broker.
-            if (results.length > 0) {
-                return results[0];
-            }
+            return Promise.all(purgedQueues)
+                .then(sum_total_purged)
+                .finally(() => {
+                    Promise.all(broker_cons.map(c => c.close()))
+                        .catch((e) => log.warn("Failed to close purge connection", e));
+                });
         });
 };
+
+function sum_total_purged(results) {
+    var totalPurged = 0;
+    results.forEach((c) => totalPurged += c);
+    log.debug("Purged %d message(s) from %d shard(s)", totalPurged, results.length);
+    return totalPurged;
+}
 
 Queue.prototype.purge = function () {
     var queueName = this.name;
@@ -128,19 +136,16 @@ Queue.prototype.purge = function () {
                var purgedQueues = broker_cons.map(broker => broker.purgeQueue(queueName));
 
                return Promise.all(purgedQueues)
-                   .then(results => {
-                       var total = 0;
-                       results.forEach((c) => total += c);
-                       log.debug("Purged %d message(s) from %d shard(s)", total, purgedQueues.length);
-                       return total;
-                   })
+                   .then(sum_total_purged)
                    .finally(() => {
                        Promise.all(broker_cons.map(c => c.close()))
                            .catch((e) => log.warn("Failed to close purge connection", e));
                    });
+
            });
     }
 };
+
 
 module.exports = function (name) {
     return new Queue(name);
--- a/agent/lib/queue.js
+++ b/agent/lib/queue.js
@@ -63,7 +63,7 @@ function create_brokers(podDefs, address_space) {
     return podDefs.map(function (podDef){
 
         var options;
-        if (address_space=='brokered') {
+        if (address_space === 'brokered') {
             options = {
                 username: 'console',
                 id:     podDef.name,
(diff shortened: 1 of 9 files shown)


agent/lib/queue.js
```diff
@@ -0,0 +1,90 @@
...
+
+function find_queue (name, pods) {
+    return Promise.all(pods.map(function (pod) {
+        return pod.broker.getQueueNames();
+    } )).then(function (results) {
+            var found = false;
+            var pod = undefined;
+            var min = undefined;
+            for (var i = 0; !found && i < results.length; i++) {
+                if (results[i].indexOf(name) >= 0) {
```
k-wall: I query the use of `indexOf`.  if I have addresses `foo` and `foobar`, and I purge `foo`, what stops this code finding erroneously finding `foobar`?

agent/lib/queue.js
```diff
@@ -0,0 +1,90 @@
...
+                    pod = pods[i];
+                    found = true;
```
k-wall: You are indexing the pod array (pods) with an index that relates to the queue array.  That seems wrong.

agent/lib/queue.js
```diff
@@ -0,0 +1,90 @@
...
+        function(currentPods) {
+
+            var podDefs = currentPods.items.map(function (pod) {
+                return pod_watcher.get_pod_definition(pod);
+            });
+            pods.update(podDefs);
+
+            return find_queue(queueName, pods.pod_list()).then(
+                function (result) {
+
```
k-wall: what if a queue is shared across many brokers?  we need to purge each shard.

lulf:
@vbusch @k-wall so thinking more about my initial comment... using status.brokerStatuses is a nice way to get all 'shards' of a queue. However, this would only work in standard address space. In brokered space, there is no such thing set in status as its just 1 broker anyway.

k-wall:
> @vbusch @k-wall so thinking more about my initial comment... using status.brokerStatuses is a nice way to get all 'shards' of a queue. However, this would only work in standard address space. In brokered space, there is no such thing set in status as its just 1 broker anyway.

For me I don't see a problem if we take a different approach at the backend for standard vs brokered.

vbusch:
I'll Update the way I get the broker for the standard address space.

agent/lib/queue.js
vbusch: results, is an array of arrays.  One for each pod.  
eg. [ ['foo', 'bar'], [queue_on_other_pod] ]
indexOf on an array, won't return foobar when searching for foo.

agent/lib/queue.js
vbusch: Relates to the comment above.
'i' is the index of the results for all the promises.  eg. [ ['foo', 'bar'], ['queue_on_other_pod'] ]
The 1st pod queried returned 'foo', and 'bar', and the 2nd pod queried returned 'queue_on_other_pod'.
The results are in the same order as the promises in Promise.all

agent/lib/queue.js
vbusch: using the address definition status, to get the brokers for standard address space.  If there are shards, it gets all of them.

agent/lib/queue.js
```diff
@@ -0,0 +1,148 @@
...
+                id:     podDef.name,
+            };
+        }
+        try {
+            options = tls_options.get_client_options(options);
+        } catch (error) {
+            log.error(error);
+        }
+
+        return artemis.connect(options);
```
k-wall: How is this being closed?

k-wall: I've used a finally callback.

agent/lib/queue.js
```diff
@@ -0,0 +1,148 @@
...
+                       return broker.purgeQueue(queueName);
+                   });
+               });
+
+               Promise.all(purgedQueues)
+                   .then(function (results) {
+                       if (results.length > 0) {
+                           return results[0];
+                       } else {
+                           return false;
```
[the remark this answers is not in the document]
k-wall: Done

agent/lib/queue.js
```diff
@@ -0,0 +1,148 @@
...
+               var brokers = create_brokers(podDefs, 'standard');
+               var purgedQueues = brokers.map(function (broker) {
+                   return (new function () {
+                       return broker.purgeQueue(queueName);
+                   });
+               });
+
+               Promise.all(purgedQueues)
+                   .then(function (results) {
+                       if (results.length > 0) {
```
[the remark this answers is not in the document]
k-wall: I now sum them all.

agent/lib/queue.js
```diff
@@ -0,0 +1,148 @@
...
+               });
+
+               var brokers = create_brokers(podDefs, 'standard');
+               var purgedQueues = brokers.map(function (broker) {
+                   return (new function () {
+                       return broker.purgeQueue(queueName);
+                   });
+               });
+
+               Promise.all(purgedQueues)
```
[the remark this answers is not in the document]
k-wall: Done

agent/lib/queue.js
```diff
@@ -0,0 +1,148 @@
...
+            });
+
+            //for the brokered address space, there will never be more then 1 broker.
+            if (results.length > 0) {
+                return results[0];
+            }
+        });
+};
+
+Queue.prototype.purge = function () {
```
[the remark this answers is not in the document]
vbusch: Yes, and in another PR we can log a kubernetes event.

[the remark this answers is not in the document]
k-wall: Re. logging a kubernetes event for purge queue from agent is tricky.  We have an variable variable ENABLE_EVENT_LOGGER which controls whether the component logs events or not.  For standard, this is unset for agent (as standard-controller writes the event log messages), but set for brokered.  To log a purge address message for both spaces, I need to ignore this env var and log the event unconditionally.  So it is possible, but the code will be ugly.  The purge PR as it stands logs to the agent log that fact that the purge has happened, the number of messages and the userid of the invoking user.  Maybe this is sufficient for now.

PR #3321, opened 2019-10-15
Fix get openshift console route

kornys:


kornys:
@enmasse-ci run tests testcase=ServiceCatalogWebTest test_only

systemtests/src/test/java/io/enmasse/systemtest/bases/TestBase.java
```diff
@@ -166,9 +166,9 @@ protected Endpoint getMessagingRoute(AddressSpace addressSpace) throws Exception
 
     protected String getOCConsoleRoute() {
         if (kubernetes.getOcpVersion() == 4) {
-            return String.format("https://console-openshift-console.%s", environment.kubernetesDomain());
+            return String.format("https://console-openshift-console.%s", environment.kubernetesDomain()).replaceAll("(?<!(http:|https:))[//]+", "/");
```
k-wall: I'm missing your intent here.  Doesn't the code currently just replace a slash, which must be proceeded with a http/https protocol part with a slash?

kornys: nope it finds all double slash and replace with single slash (except double slash after http ot https)

kornys: see log from job: 2019-10-15T15:26:56.127Z INFO  [OpenshiftWebPage] Opening openshift web page on route https://<ip>:8443/console

k-wall: I misread it as a positive backward non grouping match, rather than negative one.   Are you saying ` environment.kubernetesDomain()`can include slashes?

Separately, repeating the same character in a character class is superfluous.  i.e. you can change `[//]+` to `[/]+` without changing meaning.

PR #3328, opened 2019-10-17
Stop removing iot project after every test

kornys:
Kubernetes and ocp clusters has problems when we remove and create same namespace in evenry test. This should fix this problem.


kornys:
@enmasse-ci run tests profile=iot test_only

PR #3330, opened 2019-10-17
try to fix racey test testMessagingDuringRestartComponents

famartinrh:
### Type of change

- Bugfix

### Description

test CommonTest#testMessagingDuringRestartComponents has a possible race condition in the way the sender and receiver are stopped. This PR tries to fix that introducing a wait between the stop of those two. Other solution could be changing the assertion and not focus on exact messages sent and received, we can discuss that.


famartinrh:
@enmasse-ci run tests testcase=CommonTest#testMessagingDuringRestartComponents profile=isolated

famartinrh:
@enmasse-ci run tests testcase=CommonTest#testMessagingDuringRestartComponents profile=isolated

PR #3338, opened 2019-10-21
Test for purge messages

kornys:


kornys:
@enmasse-ci run tests testcase=FirefoxWebConsoleTest#testPurgeAddress test_only

kornys:
@lulf I've just added test iteration with sharded queue

lulf:
@kornys Great, even thought it does not necessarily guarantee that messages are spread across 2 brokers, but I think the probability for that is high when you send 1000 messages

PR #3345, opened 2019-10-21
Handle exception in callbacks

kornys:
Just proposal for handle lifecycle callbacks

### DIFF
--- a/systemtests/src/main/java/io/enmasse/systemtest/listener/JunitCallbackListener.java
+++ b/systemtests/src/main/java/io/enmasse/systemtest/listener/JunitCallbackListener.java
@@ -5,6 +5,7 @@
 package io.enmasse.systemtest.listener;
 
 import io.enmasse.systemtest.Environment;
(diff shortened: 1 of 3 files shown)


systemtests/src/main/java/io/enmasse/systemtest/listener/JunitCallbackListener.java
```diff
@@ -148,14 +139,23 @@ public void handleAfterAllMethodExecutionException(ExtensionContext context, Thr
...
+        } catch (Exception ex) {
+            try {
+                saveKubernetesState(context, ex);
+            } catch (Throwable ignored) {
+            }
+            throw ex;
         }
     }
 
+    private interface MyRunnable
```
famartinrh: could you give it a proper name :)

kornys: sure it was just a quick code :)

PR #3359, opened 2019-10-24
Check router mesh connectivity

### LINKED ISSUE #3355: Router mesh status check
**Overview**

A status check for the router mesh in the standard address space could uncover a situation where routers are not fully connected.

**Design**

The RouterStatusController class in address-space-controller which currently only checks connectors seems like an appropriate place to put this check. 

The check should query all router pods for the address space and verify that they are all connected toeachother using the same mechanism as `qdstat -n`.

lulf:
### Type of change

_Select the type of your PR_

- Enhancement / new feature

### Description

This adds a check for router mesh connectivity by querying the routers
for its connected nodes in the router network. If the list of nodes does
not contain the number of ready router pods, the address space status is
set to false, and an error message is added to the status and in the
log.

Fixes #3355

### Checklist

- [ ] Update documentation
- [ ] Update CHANGELOG.md


lulf:

PR #3360, opened 2019-10-24
Add http status code metrics

### LINKED ISSUE #3357: Expose counter metrics for request status codes in api-server
**Overview**

Providing counters for different status codes allows alerting on a lot of failed requests and that api-server should be looked into.

**Design**

The following metrics should be exposed:

* Number of 2xx
* Number of 4xx
* Number of 5xx

Alerts for 5xx should be provided.

lulf:
# Type of change

_Select the type of your PR_

- Enhancement / new feature

### Description

Expose metrics for the total number of requests, the number of 4xx
requests, the number of 5xx requests.

Some improvements to the metrics-api to allow updating a value without
redeclaring the metric.

Fixes #3357

### Checklist

- [ ] Update documentation
- [ ] Update CHANGELOG.md


lulf:

api-server/src/main/java/io/enmasse/api/server/RequestLogger.java
```diff
@@ -4,19 +4,56 @@
...
+    private final Clock clock;
+    private final Duration metricUpdateInterval = Duration.ofSeconds(5);
+    private Instant lastMetricUpdate = Instant.MIN;
+    private final Metric totalRequestsMetric = new Metric(
+            "api_server_requests_total",
+            "The total number of HTTP requests",
+            MetricType.counter);
+
+    private final Metric code4xxRequestsMetrics = new Metric(
+            "api_server_requests_4xx_total",
```
k-wall: I don't have production experience of Prometheus so I'd happily defer to other how do, but I wonder if status code should be normalised as a label so that the Prometheus user can do this: http_requests_total{status!~"4.."} to find all the client errors.

I'm being guided by the docs:

https://prometheus.io/docs/practices/naming/#labels
https://prometheus.io/docs/prometheus/latest/querying/examples/

lulf: Good point, I agree labels is the appropriate thing to use. I will update the pr

lulf:

api-server/src/main/java/io/enmasse/api/server/RequestLogger.java
```diff
@@ -70,19 +61,21 @@ public void filter(ContainerRequestContext containerRequestContext, ContainerRes
...
     private void updateMetrics(int status) {
-        if (status >= 400 && status < 500) {
-            num4xx++;
-        } else if (status >= 500) {
-            num5xx++;
+        if (status < 0 || status >= MAX_CODE) {
+            return;
         }
-        numTotal++;
+        statusCodeCounter[status]++;
```
k-wall: This isn't thread safe.

api-server/src/main/java/io/enmasse/api/server/RequestLogger.java
```diff
@@ -70,19 +61,21 @@ public void filter(ContainerRequestContext containerRequestContext, ContainerRes
...
-            totalRequestsMetric.updateValues(new MetricValue(numTotal, nowMillis));
-            code4xxRequestsMetrics.updateValues(new MetricValue(num4xx, nowMillis));
-            code5xxRequestsMetrics.updateValues(new MetricValue(num5xx, nowMillis));
+            List<MetricValue> values = new ArrayList<>();
+            for (int i = 0; i < MAX_CODE; i++) {
+                if (statusCodeCounter[i] > 0) {
+                    values.add(new MetricValue(statusCodeCounter[i], nowMillis, new MetricLabel("status", String.format("%3d", i))));
+                }
+            }
+            requestsMetric.updateValues(values);
```
k-wall: the synchronisation on this method will effectively single thread the API server every time `now.minus(metricUpdateInterval).isAfter(lastMetricUpdate)` is true.  There could be many threads awaiting this mutex.  Similarly, this is no guarantee that blocked threads will wake up in order so lastMetricUpdate may be updated incorrectly.

lulf:
@k-wall Ah, very good points. I will rework the metric lib

api-server/src/main/java/io/enmasse/api/server/RequestLogger.java
lulf: Oops, meant to make that an array of atomics

lulf:
@k-wall I've yet to test this on a cluster, but the refactoring got a bit bigger than I originally though :rofl:  I had to change the metric api quite a bit as well.

Anyway, I've added the support for histograms as well, I'm going to play around with the grafana dashboard a bit to see if the buckets make sense and if the formatting is correct.

api-server/src/main/java/io/enmasse/api/server/RequestLogger.java
```diff
@@ -4,19 +4,82 @@
...
+
+    // Buckets:
+    // 0-9 ms
+    // 10-99 ms
+    // 100-999 ms
+    // 1000-9999 ms
+    // 10000-+Inf ms
+    private static final int NUM_BUCKETS = 5;
+    private final AtomicLongArray latencyBucketsCount = new AtomicLongArray(NUM_BUCKETS);
+    private static final Long latencyBucketLimits[] = new Long[] {
```
k-wall: Are we publishing sufficient latency information to Prometheus for it to be useful?  Don't we need to breakdown by HTTP verb and endpoint?

api-server/src/main/java/io/enmasse/api/server/RequestLogger.java
```diff
@@ -26,6 +89,23 @@ public void filter(ContainerRequestContext containerRequestContext) {
...
+    private void updateMetrics(int status, long latencyMs) {
+        if (status < 0 || status >= MAX_CODE) {
+            return;
+        }
+        statusCodeCounter.incrementAndGet(status);
+
+        for (int i = 0; i < NUM_BUCKETS; i++) {
+            if (latencyBucketLimits[i] == null || latencyMs < latencyBucketLimits[i]) {
+                latencyBucketsCount.incrementAndGet(i);
+            }
```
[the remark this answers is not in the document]
lulf: According to the prometheus spec, the histograms are cumulative. https://prometheus.io/docs/concepts/metric_types/#histogram

api-server/src/main/java/io/enmasse/api/server/RequestLogger.java
lulf: Good point, I will add them as labels

lulf: Actually, I'm a bit worried about the dimentionality here. I think the metrics are useful even without the labels. They indicate that there is an issue, and if one needs to look into the details, they should be in the log.

api-server/src/main/java/io/enmasse/api/server/RequestLogger.java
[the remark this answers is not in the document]
k-wall: > do you not want to break when you've found the right bucket?

ok- ignore this.. I understand now the buckets are labelled 'le' (less than equal), so your coding is appropriate.

api-server/src/main/java/io/enmasse/api/server/RequestLogger.java
k-wall: If an app enters a phase where same it causes say lots of more creation events, response latencies will rise as PUTs normaly take longer than GETs .  This would be normal.  The human operator looking a Prometheus won't have the information to percieve that and will probably just flag the spike in response times as abnormal.

Is dimentionality actually an issue?  I think we might be talking about no more than 3  endpoints (address-space, address, user) * 4 verbs.  This seems to fall within Prometheus best practice and within the cardinality warnings within the docs.

I'm being influenced by what I see here:
https://github.com/kubernetes/apiserver/blob/master/pkg/endpoints/metrics/metrics.go#L93

k-wall: OTOH as we know EnMasse api-server's days are numbered, maybe I should worry less :)

api-server/src/main/java/io/enmasse/api/server/RequestLogger.java
```diff
@@ -4,19 +4,158 @@
...
+            10000L,
+            null};
+
+    private final Map<String, Map<String, ServiceMetrics>> serviceMetrics = new HashMap<>();
+
+    private static class ServiceMetrics {
+        private final String service;
+        private final String method;
+        private final AtomicLongArray statusCodeCounter = new AtomicLongArray(MAX_CODE);
+        private final AtomicLong latencySum = new AtomicLong(0);
```
k-wall: `latencySum' isn't incremented.

api-server/src/main/java/io/enmasse/api/server/RequestLogger.java
```diff
@@ -4,19 +4,158 @@
...
+            statusCodeCounter.incrementAndGet(status);
+
+            for (int i = 0; i < NUM_BUCKETS; i++) {
+                if (latencyBucketLimits[i] == null || latencyMs < latencyBucketLimits[i]) {
+                    latencyBucketsCount.incrementAndGet(i);
+                }
+            }
+        }
+
+        public List<MetricValue> getStatusValues() {
```
k-wall: unused method

lulf:

lulf:
@enmasse-ci run tests testcase=MonitoringTest

PR #3392, opened 2019-10-25
Handle multiple API server requests in parallel

lulf:
### Type of change

- Enhancement 

### Description

Create multiple instances of HTTP servers while keeping the service
handlers and request filters singletons.

Set the number of HTTP server instances based on the number of CPUs,
optionally via an environment variable.

Fixes #3392 

Co-authored-by: Keith Wall <kwall@apache.org>

### Checklist

- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/api-server/src/main/java/io/enmasse/api/server/ApiServer.java
+++ b/api-server/src/main/java/io/enmasse/api/server/ApiServer.java
@@ -107,9 +107,12 @@ public class ApiServer extends AbstractVerticle {
         Metrics metrics = new Metrics();
 
         HTTPHealthServer httpHealthServer = new HTTPHealthServer(options.getVersion(), metrics);
-        HTTPServer httpServer = new HTTPServer(addressSpaceApi, schemaProvider, authApi, userApi, options, clientCa, requestHeaderClientCa, clock, authenticationServiceRegistry, apiHeaderConfig, metrics);
 
-        vertx.deployVerticle(httpServer, new DeploymentOptions().setWorker(true), result -> {
+        ResteasyDeploymentFactory resteasyDeploymentFactory = new ResteasyDeploymentFactory(addressSpaceApi, schemaProvider, authApi, userApi, clock, authenticationServiceRegistry, apiHeaderConfig, metrics, options.isEnableRbac());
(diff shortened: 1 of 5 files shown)


lulf:

lulf:

api-server/src/main/java/io/enmasse/api/server/ApiServer.java
```diff
@@ -107,9 +107,12 @@ public void start(Future<Void> startPromise) throws Exception {
...
 
         HTTPHealthServer httpHealthServer = new HTTPHealthServer(options.getVersion(), metrics);
-        HTTPServer httpServer = new HTTPServer(addressSpaceApi, schemaProvider, authApi, userApi, options, clientCa, requestHeaderClientCa, clock, authenticationServiceRegistry, apiHeaderConfig, metrics);
 
-        vertx.deployVerticle(httpServer, new DeploymentOptions().setWorker(true), result -> {
+        ResteasyDeploymentFactory resteasyDeploymentFactory = new ResteasyDeploymentFactory(addressSpaceApi, schemaProvider, authApi, userApi, clock, authenticationServiceRegistry, apiHeaderConfig, metrics, options.isEnableRbac());
+        String finalRequestHeaderClientCa = requestHeaderClientCa;
+        String finalClientCa = clientCa;
+        vertx.deployVerticle(() -> new HTTPServer(options, resteasyDeploymentFactory, finalClientCa, finalRequestHeaderClientCa),
+                new DeploymentOptions().setWorker(true).setInstances(options.getNumWorkerThreads()), result -> {
```
k-wall: Maybe log options.getNumWorkerThreads() at info level too.

lulf: All options (including this) are printed at info level during startup

lulf:
@k-wall On the thread name, I'm a bit divided since it would break the uniform logging pattern we have across all components. If possible, I would prefer if one could enable such a pattern for the debug log level, if this is possible with logback.

k-wall:
I take your point about the uniformity of the component logs.   On reflection I’m happy with the patch as it stands.  Merge it!

PR #3398, opened 2019-10-28
Update keycloak image ref (#3342)

lulf:


lulf:
@k-wall Good catch, I didn't actually spot the version change here.

PR #3409, opened 2019-10-30
Fix missing role permission for configmaps

lulf:


templates/csv/enmasse.clusterserviceversion.yaml
```diff
@@ -411,7 +411,7 @@ spec:
           resources: [ "deployments" ]
           verbs: [ "get", "list", "watch", "update", "create", "patch", "delete" ]
         - apiGroups: [ "" ]
-          resources: [ "configmaps", "secrets", "persistentvolumeclaims", "services" ]
+          resources: [ "configmaps", "configmaps/finalizers", "secrets", "persistentvolumeclaims", "services" ]
```
k-wall: enmasse-operator doesn't actually require this permission and would be inconsistent with 020-Role-enmasse-operator.yaml

lulf: Fixed

PR #3424, opened 2019-11-05
Update checkstyle

obabec:
### Type of change

- Enhancement / new feature

### Checklist

- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/.checkstyle/checkstyle.xml
+++ b/.checkstyle/checkstyle.xml
@@ -18,5 +18,63 @@
         </module>
         <module name="StringLiteralEquality"/>
     </module>
+    <module name="TreeWalker">
+        <!-- Checks for Naming Conventions.                  -->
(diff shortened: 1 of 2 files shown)


obabec:
> I think LineLength, MethodLength etc. should be removed (and code removed from supressions).
> 
> I don't know specifically how the other checks will actually look like, but I think any changes to checkstyle should also include the fixes to exiting code and not use supressions.

MethodLenght and LineLenght have to be discussed. About fixing code. I have already cleaned systemtests with this rules in another PR...

lulf:
@obabec I think maybe an approach that doesn't have to make such a massive change to the whole code base in order to be applied, is to start with adding just one or a few rules, fix the code base to adhere to those, and then raise a PR for each of those additions. Otherwise, it is difficult to understand what the implications of the rule is and what the result will look like.

PR #3425, opened 2019-11-05
Fix go deps (again)

ctron:
Limit prometheus dependency to 1.1.x as 1.2.x
introduces a new dependency on github.com/cespare/xxhash/v2
which only works with go modules, which we can't use
because we still need older dependencies that don't work
with go modules.


ctron:

lulf:
Any idea the list of additions is so big compared to list of deletions?

ctron:
> Any idea the list of additions is so big compared to list of deletions?

Not really. It looks like the addition of "honnef" contributed to that. But honestly, I've given up on go dependencies.

PR #3459, opened 2019-11-13
Olm manifest image

lulf:
### Type of change

- Enhancement / new feature

### Description

Introduce an image containing the OLM manifests. This image can in turn be used to do an install without using operator hub.

### Checklist

- [ ] Update documentation
- [ ] Update CHANGELOG.md


lulf:

olm-manifest/src/main/resources/latest/enmasse.package.yaml~
```diff
@@ -0,0 +1,4 @@
+packageName: ${env.PACKAGE_NAME}
```
ctron: I guess this file shouldn't be here.

.travis.yml
```diff
@@ -28,7 +28,6 @@ jobs:
   include:
   - stage: build
     before_install:
-    - ./.travis/find-qdrouterd-base-versions.sh
```
ctron: This step saved us from having different router versions in the past. It should ensure that we don't have a different router image as a base for e.g. Dockerfile the "qdr configurator" is based on, or the router image we run.

I think we still need some kind of validation step to achieve that.

lulf: Its a good point, because the check has already failed to detect that they are different :) 

https://github.com/EnMasseProject/enmasse/blob/master/iot/iot-proxy-configurator/Dockerfile#L6
https://github.com/EnMasseProject/enmasse/blob/master/Makefile.env.mk#L23

I think the approach we should take here is to make sure the ${ROUTER_IMAGE} env is used in the FROM for the proxy configurator Dockerfile. That should ensure everything comes from a single source.

I will make that modification in this PR if thats ok. 

(The original reason I remove the script, was that it barfed at the old OLM manifests)

lulf:

.travis.yml
lulf: I've made the modification to templatize the iot-proxy-configurator Dockerfile so that the FROM gets set to the same value.

lulf:

PR #3462, opened 2019-11-14
Fix permission typo

lulf:


lulf:
Closing, as this has no effect on master where configmaps are no longer used.

PR #3469, opened 2019-11-15
Prevent continuous updates when project is creating

ctron:
As the map key order is randomized, a change will cause a new order in
the message field, which will cause an update on the object, which will
cause another change, …

### Type of change

- Bugfix

### Checklist

- [ ] Update documentation
- [ ] Update CHANGELOG.md


ctron:

PR #3476, opened 2019-11-18
Run systemtests testsuite installing enmasse via BUNDLE or OLM

famartinrh:
### Type of change
- Enhancement / new feature
### Description

This PR uses some recent changes added to ease OLM testing and adds the concept of enmasse install type (env var `INSTALL_TYPE` ) to be able to run the test suite choosing how to install enmasse. The work done is:
- Add `OperatorLifecicleManagerTest` to acceptance profile 
- Add `OLMRequired` junit test condition to check if OLM is available in current cluster
- Special treatment of "infra namespace" in order to make test suite work. In bundle installation it is `enmasse-infra` but in OLM installation it becomes `openshift-operators` in ocp or `operators` in minikube env
- Install example plans ,roles and authservices when installing via OLM, in order to make test suite work.
- Add profile acceptance-no-iot because a the moment of testing iot functionality installing enmasse via OLM leads into some errors, so this profile is needed to ease testing the new systemtests feature
- Moves infinispan deployment to a separated namespace because an error found deploying infinispan to openshift-operators

### Checklist

- [ ] Update documentation
- [ ] Update CHANGELOG.md


famartinrh:
@enmasse-ci run tests profile=acceptance ocp4 test_only

famartinrh:
@enmasse-ci run tests profile=acceptance ocp4 test_only OLM

famartinrh:
@enmasse-ci run tests profile=acceptance ocp4 test_only OLM

famartinrh:
@enmasse-ci run tests profile=acceptance ocp4 test_only

famartinrh:
@enmasse-ci run tests profile=acceptance ocp4 test_only

famartinrh:
@enmasse-ci run tests profile=acceptance ocp4 test_only OLM

famartinrh:
@enmasse-ci run tests profile=olm ocp4 test_only OLM

famartinrh:
@enmasse-ci run tests testcase=OLM** ocp4 test_only

systemtests/src/main/java/io/enmasse/systemtest/Environment.java
```diff
@@ -36,6 +37,8 @@
     public static final String TAG_ENV = "TAG";
     public static final String PRODUCT_NAME_ENV = "PRODUCT_NAME";
     public static final String APP_NAME_ENV = "APP_NAME";
+    public static final String INSTALL_TYPE = "INSTALL_TYPE";
+    public static final String OLM_INSTALL_TYPE = "OLM_INSTALL_TYPE";
```
lulf: Is this redundant?

famartinrh: I think it's not redundant, OLM_INSTALL_TYPE is a kind of subtype of OLM INSTALL_TYPE because as you know with OLM we can install in the default namespace (ie: openshift-operators) or in the namespace `enmasse-infra`

famartinrh:
good point @lulf , I would like to have most tests to work with both install types in order to get the best test coverage of OLM installation. However I agree this concept of installation type could be a requirement specific at test-class level but again I would like to get most of tests working in both install modes so maybe putting annotations everywhere is a bit overkill.
Anyway I think this discussion is pretty much needed in order to all agree and stablish a path to improve our testing of enmasse installed via OLM, so @kornys do you have something to say?

famartinrh:
@enmasse-ci run tests testcase=OLM** ocp4 test_only

famartinrh:
thinking of it @lulf I kind of agree with your idea of the annotation to specify supported install types in tests classes because altough I want most tests working in OLM, for sure not all will work on both kinds of installation. 
I would keep the install type so tests are run twice but in different jobs (you maybe were already saying that), will be better to have separate jobs for each install type to avoid possible environment issues and to avoid test execution time to become too long.
So I would suggest an annotation like 
```
@SupportedInstallTypes({EnmasseInstallType.BUNDLE})
```
Also I would make the absence of the annotation mean that test it's assumed to work on both installation types

lulf:
@famartinrh good point, agreed we probably want separate jobs, but the ability to select in PR as well for instance. I like it!

kornys:
Ok but keep in mind that with downstream we dont have capacity to run 6jobs for cpaas for instance

famartinrh:
@enmasse-ci run tests profile=acceptance ocp4 test_only OLM

famartinrh:
@enmasse-ci run tests profile=acceptance ocp4 test_only OLM

famartinrh:
we still have bugs that prevent us from successfully deploy enmasse via OLM, at least in our CI env using CRC, however I think this PR is ready for merge in order to let developers easily investigate the issues. I'll run tests installing using bundle just to make sure, again, that our classic workflow is still working

famartinrh:
@enmasse-ci run tests profile=acceptance ocp4

famartinrh:
@enmasse-ci run tests profile=acceptance

famartinrh:
@enmasse-ci run tests profile=acceptance

.jenkins/Jenkinsfile-pr
```diff
@@ -124,6 +127,39 @@ pipeline {
             }
 
         }
+        stage('workaround olm-manifest step 1') {
```
lulf: I wonder if a more uniform step would be to edit the Makefiles to Allow setting an INTERNAL_DOCKER_REGISTRY (which by default == DOCKER_REGISTRY) and override that for the olm-manifest image. Then these steps should be unnecessary and we'll save issues in other DSL/pipelines as well. I can share a pr of what i mean.

famartinrh: hmm, that PR is welcome @lulf :) , note that `make docker_build` and `make docker_tag` have to be run with different DOCKER_REGISTRY values in order to push to external registry but having CSV pointing to internal registry

lulf: I tried to modify the makefiles, but there wasn't any clean/easy way to do that. However, I recommend instead to

* Build with internal registry
* Run `make DOCKER_REGISTRY=${PUBLIC_REGISTRY} templates docker_tag docker_push`

famartinrh:
@enmasse-ci run tests testcase=OLM** ocp4

famartinrh:
@enmasse-ci run tests testcase=OLM** ocp4

Issue #3456, opened 2019-11-12
Addresses (broker pods) are not fully deleted 


kornys:
**Describe the bug**
During running shared test suite we can see that broker pods in infra namespaces are not deleted when addresses are removed. 

It caused fails in our systemtests suite because pods consumes all resources on nodes so another tests cannot create another address (pods)

**Additional context**
oc get events
```
2019-11-12T11:18:06Z   2019-11-12T10:03:20Z   331       standard-controller.1070108445                       Address       <none>    Warning   BrokerDeleteFailed   standard-controller      Error deleting broker cluster broker-b41e7da-pj2k: Failure executing: PATCH at: https://172.30.0.1/apis/apps/v1/namespaces/enmasse-infra/statefulsets/broker-b41e7da-pj2k. Message: StatefulSet.apps "broker-b41e7da-pj2k" is invalid: spec: Forbidden: updates to statefulset spec for fields other than 'replicas', 'template', and 'updateStrategy' are forbidden.. Received status: Status(apiVersion=v1, code=422, details=StatusDetails(causes=[StatusCause(field=spec, message=Forbidden: updates to statefulset spec for fields other than 'replicas', 'template', and 'updateStrategy' are forbidden., reason=FieldValueForbidden, additionalProperties={})], group=apps, kind=StatefulSet, name=broker-b41e7da-pj2k, retryAfterSeconds=null, uid=null, additionalProperties={}), kind=Status, message=StatefulSet.apps "broker-b41e7da-pj2k" is invalid: spec: Forbidden: updates to statefulset spec for fields other than 'replicas', 'template', and 'updateStrategy' are forbidden., metadata=ListMeta(_continue=null, remainingItemCount=null, resourceVersion=null, selfLink=null, additionalProperties={}), reason=Invalid, status=Failure, additionalProperties={}).
```

kornys:
@lulf also admin pods are not deleted

k-wall:
Is there a particular test that is known to show this issue?

kornys:
@k-wall nope, issue appears after many test in shared space. You can check logs in master jenkins shared job

kornys:
@k-wall check this test io.enmasse.systemtest.shared.standard.TopicTest.testShardedTopic

lulf:
@kornys @k-wall  I'll have a look, its probably related to CRD refactor

lulf:
[...]

PR #3481, opened 2019-11-19
Fabric8 kubernetes-client race when deleting a deployment may leave pods behind

k-wall:
Fix ##3480

### Type of change

- Bugfix

### Description

Works around an defect in the Fabric8 kubernetes-client.

### Checklist

- [ ] Update documentation
- [ ] Update CHANGELOG.md


k-wall:

lulf:
Nice catch! I suspect this is the same as https://github.com/EnMasseProject/enmasse/issues/3456

k-wall:
> @k-wall Nice catch! I suspect this is the same as #3456 , or do you think this applies to Deployments only?

This one is deployments only, so it would explain broker pods being left behind in brokered but not standard.

k-wall:

PR #3485, opened 2019-11-20
Supress download logging maven

kornys:


kornys:

kornys:

lulf:
@kornys I didn't realize this would affect running tests locally as well. I think we shouldn't require everyone to upgrade to maven 3.6 in order to run tests at all.

kornys:
@lulf hmmm, I think it is ok to require latest tool version, anyway I will open an PR

kornys:
#3488

lulf:
@kornys IMO the rule of thumb is what is available in latest package managers :) maybe it can be detected by script?

lulf:
@kornys alternatively add an environment variable that CI can pass?

gemmellr:
You shouldn't hammer the Apache archive by downloading maven from it on every build.

You could use the travis cache (https://docs.travis-ci.com/user/caching/) to avoid re-downloading it every time (which could similarly apply to deps themselves even without this PR).

You could also get it from maven central (http://repo2.maven.org/maven2/org/apache/maven/apache-maven/3.6.2/apache-maven-3.6.2-bin.zip).

Or best of all, both.

kornys:
@gemmellr thanks! will fix that

PR #3486, opened 2019-11-20
systemtests log pods on test error

famartinrh:


famartinrh:
because just doing `oc get pods` gives us perfect logging with pods status, number of restarts ... I think it's useful @kornys

lulf:
Is there not already a `oc describe pods` in `saveKubernetesState` which captures all pod information?

famartinrh:
yes it is @lulf , however in this PR I only had in mind to improve log output, I just was thinking on improve the logging and change that loop we had that only logs pod names and change it with the standard output of `oc get pod`

kornys:
You can get all these information from fabric8 client I guess

kornys:
Anyway lgtm

PR #3491, opened 2019-11-21
Travis use maven from maven central

kornys:


lulf:
@kornys You can add in travis.yml to the list under 'cache': ${HOME}/.dependencies . then have the script check if the tarball exists in that folder before downloading it. That should save time both downloading and avoiding http request.

kornys:
@lulf sure will do

PR #3493, opened 2019-11-21
Delete crds first

lulf:
### Type of change

- Bugfix

### Description
If address spaces exist, uninstall procedure will hang because clusterroles are deleted before CRDs, which will wait forever for finalizer to be updated. However, the remove clusterroles prevents the address-space-controller from making the update.

### Checklist

- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/address-space-controller/src/main/java/io/enmasse/controller/AddressFinalizerController.java
+++ b/address-space-controller/src/main/java/io/enmasse/controller/AddressFinalizerController.java
@@ -38,14 +38,19 @@ public class AddressFinalizerController extends AbstractFinalizeController {
         final String addressSpaceNamespace = addressSpace.getMetadata().getNamespace();
         final AddressApi addressApi = addressSpaceApi.withAddressSpace(addressSpace);
 
-        // If API does not exist, the address CRD does not exist so we drop the finalizer
-        if (!addressApi.exists()) {
-            return Result.completed(addressSpace);
-        }
-
         ContinuationResult<Address> list = null;
         do {
-            list = addressApi.listAddresses(addressSpaceNamespace, BATCH_SIZE, list, null);
+            try {
+                list = addressApi.listAddresses(addressSpaceNamespace, BATCH_SIZE, list, null);
+            } catch (KubernetesClientException e) {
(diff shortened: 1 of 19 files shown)


address-space-controller/src/main/java/io/enmasse/controller/AddressFinalizerController.java
```diff
@@ -39,7 +40,17 @@ public Result processFinalizer(final AddressSpace addressSpace) {
 
         ContinuationResult<Address> list = null;
         do {
-            list = addressApi.listAddresses(addressSpaceNamespace, BATCH_SIZE, list, null);
+            try {
+                list = addressApi.listAddresses(addressSpaceNamespace, BATCH_SIZE, list, null);
+            } catch (KubernetesClientException e) {
+                // If not found, the address CRD does not exist so we drop the finalizer
+                if (e.getCode() == 404) {
+                    return Result.completed(addressSpace);
```
k-wall: is there a simple way that we can establish with certainty that the 404 refers to to the absence of the crd (rather than listAddresses returning 404 for some other reason).  I know list returns an empty list of there are none.  My concern is a minor one - and the complication of the additional check may not be worth while.

lulf: I'm not sure really. One could check if the resource is listed by querying the /apis/enmasse.io/v1beta1 as an additional check, and failing that mark it as finalized. 

Worst case is that addresses would be orphaned.

lulf: @k-wall Added an attempt to query the API checking if resource exists. Not sure if I like it or not :)

k-wall: Yeah that quite a lot of ugliness..  How about as an alternative simple log a WARN when we go down the 404 branch, that way if we ever go down that path as runtime, they'll be a clue in the log.

lulf: Ok, I've added a warning and removed the ugliness.

PR #3495, opened 2019-11-21
Maven from mvn central

kornys:


kornys:
@lulf seems travis does not want to trigger job :D

PR #3499, opened 2019-11-22
Console server backend

k-wall:
New console server backend that exposes a GraphQL API to the console application.


cmd/console-server/main.go
```diff
@@ -0,0 +1,384 @@
...
+	"time"
+)
+
+/*
+
+For development purposes, you can run the console backend outside Kubernetes:
+
+export DISABLE_PROPAGATE_BEARER=1
+export ENMASSE_ADDRESS_SPACE_TYPE_FILE=.../console//console-server/src/main/resources/addressSpaceTypes.json
+ENMASSE_ADDRESS_TYPE_FILE=.../console//console-server/src/main/resources/addressTypes.json
```
lulf: Consider adding this to a 'run' target in the makefile in console/console-server

k-wall: I've got a few ideas to improve how to run the console-server outside the kubernetes environment for development purposes, I'll try to incorporate this idea too.

k-wall: @lulf With 307a953, for development purposes, you'll run an oc route to expose the agent(s) outside the container so that the console-server can contact them.  This behaviour is all controlled by passing the `--developmentMode=true` command line argument.  I've also added the `run` target to the Makefile, as you suggest.  Comments welcome.

pkg/consolegraphql/cache/cache.go
```diff
@@ -0,0 +1,294 @@
...
+	Delete(objs ...interface{}) error
+	DeleteByPrefix(idxName, keyPrefix string) error
+
+	Get(idxName string, keyPrefix string, filter ObjectFilter) ([]interface{}, error)
+	Dump() error
+
+	GetKeyCreator(idxName string) (KeyCreator, error)
+}
+
+type MemdbCache struct {
```
lulf: I think ideally we should be able to set a size limitation on this cache so that we don't run into OOM situations in systems with many addresses.

Also, if bounded, there should be a replacement policy.

k-wall: @lulf unfortunately memdb doesn't have the ability to constrain it own size, nor the ability to demand load.   With the current implementation console-server has to be sized sufficiently large to accommodate addresss+connections+links.  

my idea (long term) is to swap out memdb with another key value store implementation with the ability to page to disk and give console-server a non-persistent transient volume.  That way it would spill out to disk if/when it needed to and this would be transparent to the application.  I know BDB has suitable properties. There may be more modern implementations with better Go bindings.

for now, console-server's memory needs to be configurable for its CR.

k-wall:
> Reviewed filter parser. @k-wall Maybe this is there somewhere already, but it would be nice to have some documentation on the format (i.e. EBNF)

Agreed.  For the next release, the GraphQL interface won't be considered a public API so we could defer adding to the docs until we've had a chance to mature it.   As you'll have seen, the filter specification is cut down SQL-92 where clause parser.   JSONPath expressions don't seem to be formalised into a EBNF specification.  I side stepped this issue by insisting that they are enclosed in backticks, so they may be treated opaquely.

k-wall:
> @k-wall Regarding metric names. Is it expected that these names (enmasse_accepted etc.) will be used as the prometheus metrics? In that case, they should probably follow the convention from https://prometheus.io/docs/practices/naming/
> 
> I.e. enmasse_accepted -> enmasse_accepted_total. etc.

Yes, that's a good point.  I want to leave open the possibility that console-server pull metrics from a Prometheus server, so adopting the naming conventions would be sensible.

kornys:
@k-wall could you please rebase branch on actual master? And maybe also squash commits. Thank you

kornys:

lulf:
We should also add a changelog entry, but there are probably other things that needs fixing, and I'd prefer to get this to master now first. Great work and thanks to everyone for the new console!

PR #3504, opened 2019-11-25
Try different statefulset workaround

lulf:
### Type of change

- Bugfix

### Description

Attempt different workaround the statefulset deletion error.

### Checklist

- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/standard-controller/src/main/java/io/enmasse/controller/standard/KubernetesHelper.java
+++ b/standard-controller/src/main/java/io/enmasse/controller/standard/KubernetesHelper.java
@@ -196,22 +196,10 @@ public class KubernetesHelper implements Kubernetes {
         }
 
         for (HasMetadata resource : resources.getItems()) {
-            int maxRetries = 10;
-            int retry = 0;
-            while (true) {
-                try {
-                    client.resource(resource).cascading(true).delete();
-                    break;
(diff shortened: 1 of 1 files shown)


lulf:

lulf:

lulf:
@k-wall I'm not sure actually, I was hoping it would avoid patching parts of the object other than replicas (the error it gets is that it tries to PATCH fields which cannot be patched on a statefulset). However, I noticed that I forgot to set cascading, so the behavior would not be the same. 

However, it looks from the PR job that there are no leftover broker logs in for the failed tests, so it might have fixed the issue, but I don't understand why the pods are not orphaned

lulf:
@k-wall From #3506 I see in the log that deleting tries to apply the following patch:

```
[{"op":"replace","path":"/spec/replicas","value":0},{"op":"add","path":"/spec/template/metadata/namespace","value":"enmasse-infra"},{"op":"add","path":"/spec/volumeClaimTemplates/0/metadata/namespace","value":"enmasse-infra"}]
```

However, the last part of that patch touches the volumeClaimTemplates. It seems to me that the volume claim templates is modified by kubernetes-client somehow.

The fix in this PR seems to suffice as a workaround.

PR #3519, opened 2019-11-27
WIP: External test clients

lulf:
External clients for testing EnMasse API and messaging service and recording anomalies.

Currently only api client is implemented, I will work on adding the messaging client as well.

I think this should ideally be in a separate repo, but I'm just raising this PR to make everyone aware of its existence, and would like comments on the test client itself.


external-test-clients/src/main/java/io/enmasse/test/ApiClient.java
```diff
@@ -0,0 +1,209 @@
...
+                    long errorTime = errorEnd - errorStart;
+                    errorHist.recordValue(errorTime);
+                    sumOutage.addAndGet(errorTime);
+                }
+                return ret;
+            } catch (KubernetesClientException e) {
+                failures.compute(e.getCode(), (key, oldValue) -> oldValue == null ? 0 : oldValue + 1);
+                errorStart = System.nanoTime();
+                Thread.sleep(500);
+            }
```
k-wall: Handle other RuntimeExceptions, just in case.

external-test-clients/src/main/java/io/enmasse/test/ProbeClient.java
```diff
@@ -0,0 +1,191 @@
...
+                .setTrustAll(true)
+                .setHostnameVerificationAlgorithm("");
+
+        ProtonClient client = ProtonClient.create(vertx);
+        client.connect(protonClientOptions, host, port, connectResult -> {
+            if (connectResult.succeeded()) {
+                ProtonConnection connection = connectResult.result();
+                ProtonReceiver receiver = connection.createReceiver(address);
+                receiver.handler((protonDelivery, message) -> {
+                    startPromise.complete();
```
k-wall: Organise to fail the promise if the message isn't received in a reasonable time period.

lulf:
Moved to https://github.com/EnMasseProject/external-test-clients

PR #3524, opened 2019-11-29
Add delegate cache to handle caching of global objects

lulf:
This allows for using both namespaced and global watches based on object
type.

### Type of change

- Enhancement / new feature

### Description

An implementation of controller-runtime Cache interface that delegates lookups and watches to either a local namespaced cache, or a global cache, depending on the type. The type mapping is registered at manager creation time.

The intention is to use this to allow the enmasse-operator deployment to watch both local and global objects.

### Checklist

- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/cmd/controller-manager/main.go
+++ b/cmd/controller-manager/main.go
@@ -16,6 +16,7 @@ import (
 	enmassescheme "github.com/enmasseproject/enmasse/pkg/client/clientset/versioned/scheme"
 	"k8s.io/client-go/kubernetes/scheme"
 
+	"github.com/enmasseproject/enmasse/pkg/cache"
 	"github.com/enmasseproject/enmasse/pkg/controller"
 	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
 	"sigs.k8s.io/controller-runtime/pkg/client/config"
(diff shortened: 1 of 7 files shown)


lulf:

kornys:
@enmasse-ci run tests profile=shared ocp4

kornys:
@lulf I see weir behaviour, it is not related to this PR (but it is a reason why on ocp4 second test failed), when I create address for example standard.address-pepa, delete it and then create again immediately, CR is marked with status Active but broker pod is in state terminating.

lulf:
@kornys I am not totally surprised that this causes issues, because we don't have finalizers for addresses yet, so address is gone even though brokers have still not shut down. So I think the fix for this is to add address finalizer support, which I think we should do soon. In the mean time I would just use a different address.

kornys:
@lulf yup that makes sense, and create one test (disabled) just for finalizers of addresses. I will take care about it.

lulf:
Example usage: https://github.com/lulf/enmasse/blob/convert-messaging-user-crd/cmd/controller-manager/main.go#L49

PR #3529, opened 2019-12-02
Enable caching of token and subject reviews so we can avoid hitting k…

dejanb:
…ube api for every registry call

### Type of change

- Enhancement / new feature

### Description

Device registry management API is using Kubernetes API for authentication and authorization. That means that every call ends up with two calls to the Kube API. To avoid this we can cache auth details for a certain period of time.


dejanb:

iot/iot-device-registry-infinispan/src/main/java/io/enmasse/iot/registry/infinispan/util/DeviceRegistryTokenAuthProvider.java
```diff
@@ -36,17 +43,104 @@
...
 
         final String token = authInfo.getString(TOKEN);
 
+        performTokenReview(token, tokenResult -> {
+            if (tokenResult.failed()) {
+                resultHandler.handle(Future.failedFuture(tokenResult.cause()));
+            }
+            TokenReview review = tokenResult.result();
+            if (review.isAuthenticated()) {
+                String role = getRole(authInfo, review, authApi);
```
ctron: I think if getRole throws an exception there, then the result handler will never be called.

dejanb:

pkg/apis/iot/v1alpha1/types_config.go
```diff
@@ -172,6 +176,7 @@ type FileBasedDeviceRegistry struct {
 
 type InfinispanDeviceRegistry struct {
 	Server              InfinispanServer `json:"server"`
+	Management          InfinispanRegistryManagement	`json:"management"`
```
ctron: It looks like it misses a call to `gofmt`

pkg/controller/iotconfig/infinispan_device_registry.go
```diff
@@ -119,6 +119,7 @@ func (r *ReconcileIoTConfig) reconcileInfinispanDeviceRegistryDeployment(config
 
 			{Name: "HONO_REGISTRY_SVC_SIGNING_SHARED_SECRET", Value: *config.Status.AuthenticationServicePSK},
 			{Name: "ENMASSE_IOT_REGISTRY_AMQP_NATIVE_TLS_REQUIRED", Value: strconv.FormatBool(nativeTls)},
+			{Name: "ENMASSE_IOT_REGISTRY_REST_AUTH_TOKEN_CACHE_EXPIRATION", Value: service.Infinispan.Management.AuthTokenCacheExpiration},
```
ctron: I am not sure what will happen in the case it is unset. Will it keep the default, or reset it to "null" in Spring?

iot/iot-device-registry-infinispan/src/main/java/io/enmasse/iot/registry/infinispan/util/DeviceRegistryTokenAuthProvider.java
```diff
@@ -36,17 +41,55 @@
...
 
         final String token = authInfo.getString(TOKEN);
+        tokens.computeIfAbsentAsync(token, review -> authApi.performTokenReview(token))
+                .thenAccept(tokenReview -> {
+                    if (tokenReview != null && tokenReview.isAuthenticated()) {
+                        String role = getRole(authInfo);
+                        if (role != null) {
+                            RbacSecurityContext securityContext = new RbacSecurityContext(tokenReview, authApi, null);
+                            authorizations
+                                    .computeIfAbsentAsync(role, authorized -> securityContext.isUserInRole(role))
```
ctron: This can again throw an Exception, in which case the handler wouldn't get called. Also, to my understanding, the method can block as it does a remote call. I think that would block the case. You could avoid that with `thenAcceptAsync`.

iot/iot-device-registry-infinispan/src/main/java/io/enmasse/iot/registry/infinispan/util/DeviceRegistryTokenAuthProvider.java
```diff
@@ -36,17 +41,55 @@
...
+        tokens = cacheManager.getCache("tokens");
+        authorizations = cacheManager.getCache("subjects");
+
     }
 
     @Override
     public void authenticate(JsonObject authInfo, Handler<AsyncResult<User>> resultHandler) {
 
         final String token = authInfo.getString(TOKEN);
+        tokens.computeIfAbsentAsync(token, review -> authApi.performTokenReview(token))
```
ctron: The `performTokenReview` might block, I think we need an async version, or put it on a worker thread pool.

pkg/apis/iot/v1alpha1/types_config.go
dejanb: Fixed. Thanks

pkg/controller/iotconfig/infinispan_device_registry.go
dejanb: It keeps a default value of 60 sec

iot/iot-device-registry-infinispan/src/main/java/io/enmasse/iot/registry/infinispan/util/DeviceRegistryTokenAuthProvider.java
dejanb: As discussed offline, `computeIfAbsentAsync()` will execute blocking call async. And the exception will be handled by `exceptionally()` now.

iot/iot-device-registry-infinispan/src/main/java/io/enmasse/iot/registry/infinispan/util/DeviceRegistryTokenAuthProvider.java
dejanb: Same as above

ctron:

ctron:

ctron:

PR #3532, opened 2019-12-02
Create github actions workflows

obabec:
- Smoke tests
- Unit tests

### Type of change

- Enhancement / new feature

### Description
 Proposal for github actions. Possible replacement for Travis.

### Checklist

- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/.github/scripts/build.sh
+++ b/.github/scripts/build.sh
@@ -1,0 +1,19 @@
+#!/bin/bash
+set -e
+echo "Make"
+make SKIP_TESTS=true
+
+echo "Build"
+make docker_build
+
+if [[ -v RELEASE ]]
+then
+    echo "Logging in to Docker Hub"
+    docker login -u ${REGISTRY_USER} -p ${REGISTRY_PASS} ${DOCKER_REGISTRY} 
+fi
+
+echo "Push to registry"
+make -j 4 docker_tag docker_push
+
+echo "Generate templates"
+make templates
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,127 @@
+name: Release build
+env:
+  TAG: "latest"
+  DOCKER_ORG: enmasse
+  KUBERNETES_NAMESPACE: "enmasse-infra"
+  REGISTRY_USER: "enmasse+ci "
+
+on:
+  push:
+    branches:
+      - master
+      - actions
+    tags:
+      - 'v*'
+
+jobs:
+  unit:
+    name: Unit tests
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@master
+
+    - name: Set up JDK 11
+      uses: actions/setup-java@v1
+      with:
+        java-version: 11.0.2
+    
+    - name: Setup Ruby environment with ruby-build
+      uses: clupprich/ruby-build-action@v1.0.0
+
+    - name: Install dependencies
+      run: ./.github/scripts/install_dep.sh
+
+    - name: Run unit tests
+      run: make
+
+    - name: Collect logs
+      if: failure()
+      run: ./.github/scripts/collectLogs.sh
+
+    - name: Upload artifacts
+      if: failure()
+      uses: actions/upload-artifact@v1.0.0
+      with:
+        name: unit-logs
+        path: artifacts
+
+
+  smoke:
+    name: Smoke tests and Release
+    needs: unit
+    env:
+      DOCKER_REGISTRY: "quay.io"
+      TAG: "latest"
+      RELEASE: true
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@master      
+    
+    - name: setup-docker
+      uses: docker-practice/actions-setup-docker@0.0.1
+    
+    - name: Minikube setup with registry
+      uses: EnMasseProject/minikube-setup@V1.0.0
+      with:
+        k8s-version: 1.15.0
+    
+    - name: Set up JDK 11
+      uses: actions/setup-java@v1
+      with:
+        java-version: 11.0.2
+
+    - name: Setup Ruby environment with ruby-build
+      uses: clupprich/ruby-build-action@v1.0.0
+  
+    - name: Install dependencies
+      run: ./.github/scripts/install_dep.sh
+    
+    - name: Setup OLM
+      run: ./.travis/setup-olm.sh
+
+    - name: Build
+      run: |
+        export REGISTRY_PASS=${{ secrets.REGISTRY_PASS }}
+        echo $REGISTRY_PASS
+        ./.github/scripts/build.sh
+
+    - name: Smoke tests
+      run: ./.github/scripts/test.sh
+
+    - name: Collect logs
+      if: failure()
+      run: ./.github/scripts/collectLogs.sh
+
+    - name: Upload artifacts
+      if: failure()
+      uses: actions/upload-artifact@v1.0.0
+      with:
+        name: smoke-logs
+        path: artifacts
+
+    - name: Create Release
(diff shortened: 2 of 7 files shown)


.github/workflows/smoke.yml
```diff
@@ -0,0 +1,71 @@
...
+    - name: Generate templates
+      run: make templates
+
+    - name: Smoke tests
+      run: ./.travis/test.sh
+
+    - uses: actions/upload-artifact@v1
+      with:
+        name: test-logs
+        path: artifacts/
```
kornys: should be /tmp/testlogs

kornys: and also systemtests/target/surefire-reports should be archived

.github/workflows/unit.yml
```diff
@@ -0,0 +1,33 @@
...
+        sudo apt-get update
+        sudo apt-get install golang-1.12
+
+    - name: Set up JDK 11
+      uses: actions/setup-java@v1
+      with:
+        java-version: 11.0.2
+
+    - name: Run unit tests
+      run: make
```
kornys: you should archive all **/target/surefire-report xml files

.github/workflows/smoke.yml
obabec: Deploying artifacts is not fully implemented yet. I will try to implement this feature soon as possible.

lulf:
Great work @obabec! I think we should switch to github actions especially as it looks we can get more debug info for failing builds, a killer feature IMO. I also like that you've split build into unit and smoke/systemtests. A couple of things that I think needs to be in place before we can switch:

* Set envs so that images are pushed to quay.io for master builds and release builds (only if both unit and systemtests have passed)
* Publishing release zip on github release page (only if both builds have succeeded)
* Cleanup/move setup actions etc. into EnMasseProject org to avoid relying on personal user repos where possible

obabec:
@lulf About personal repos. My plan is to release Minikube setup action as part of Enmasse project, so everyone can contribute and change things ad hoc. I will do little bit investigation about releasing. I hope there is some more elegant way for publishing if builds passed. One update from my side. I have implemented uploading artifacts, looks great :). Thanks for your point :)

kornys:
@lulf WDYT?

obabec:
@lulf I got your point with scripts, but I also think it's good to have these steps separated. because if one of them fail we will know which one immediately, if script fails we have to list logs and find what happened. Maybe we should figure out some good compromise. WDYT?

.github/workflows/release.yml
```diff
@@ -0,0 +1,154 @@
...
+      
+    - name: Build images
+      id: build
+      run: make docker_build
+      
+    - name: Tag images
+      run: make docker_tag
+      
+    - name: Push images
+      run: make -j 4 docker_tag docker_push
```
lulf: No need to tag images twice

.github/workflows/release.yml
```diff
@@ -0,0 +1,154 @@
...
+      
+    - name: Generate templates
+      run: make templates
```
lulf: This has already been done as part of build

.github/workflows/release.yml
```diff
@@ -0,0 +1,154 @@
...
+        echo "::set-env name=DOCKER_REGISTRY::quay.io"
+
+    - name: Build clean images
+      if: success()
+      run: |
+        make clean
+        make SKIP_TESTS=true
+        make docker_build
+        make docker_tag
+        make templates
```
lulf: I think that we should have the images pushed before the test (I know I may have stated otherwise, but this just occured to me). Otherwise, I'm worried that we're releasing templates we have not really tested. Its better to push images prematurely and have tests fail, than to push bad templates without knowing it.

.github/workflows/smoke.yml
```diff
@@ -0,0 +1,75 @@
...
+      run: make SKIP_TESTS=true
+      
+    - name: Build images
+      run: make docker_build
+      
+    - name: Tag images
+      run: make docker_tag
+      
+    - name: Push images
+      run: make -j 4 docker_tag docker_push
```
lulf: No point in tagging twice.

.github/workflows/smoke.yml
```diff
@@ -0,0 +1,75 @@
...
+      run: sudo apt install maven
+  
+    - name: Install golang
+      run: |
+        sudo add-apt-repository ppa:longsleep/golang-backports
+        sudo apt-get update
+        sudo apt-get install golang-1.12
+
+    - name: Setup OLM
+      run: ./.travis/setup-olm.sh
```
lulf: This is missing from the release.yaml

.github/workflows/release.yml
obabec: I may just tag images from local registry to quay instead of building new.

lulf: The problem is with the templates, as they would have to be regenerated. IMO we should not regenerate them after testing.

obabec: Ye I should skip whole this sections and just re-tag from localhost:*** to quay.io***. Is this acceptable ?

.github/workflows/release.yml
```diff
@@ -0,0 +1,106 @@
+name: Release build
+env:
+  IMAGE_REGISTRY: "localhost:5000"
+  TAG: "latest"
+  DOCKER_ORG: enmasse
+  KUBERNETES_NAMESPACE: "enmasse-infra"
+  SYSTEMTEST_PROFILE: smoke
+  EXTERNAL_REGISTRY: "quay.io"
+  REGISTRY_USER: "obabec+actions"
```
lulf: Should be updated to the enmasse+ci user

.github/workflows/release.yml
```diff
@@ -0,0 +1,106 @@
...
+        ./.github/scripts/retag_images.sh
+        echo "::set-env name=IMAGE_REGISTRY::quay.io"
+
+    - name: Release binaries
+      if: success()
+      id: release
+      run: |
+        ./.github/scripts/release.sh
+        echo "::set-env name=TAG::$(grep "release.version" pom.properties| cut -d'=' -f2)"
+        echo ::set-output name=tag::$TAG
```
lulf: I still thing retagging is not safe. If something goes wrong in the script, we may have a broken release template without knowing. I think we should just simplify this action and push images to quay and test with non-replaced templates.

.github/scripts/build.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+echo "Setup OLM"
+$HOME/.travis/setup-olm.sh
```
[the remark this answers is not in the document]
obabec: I don't think minikube action if good fit for this. Because if we left minikube action as it is, other teams like streams,.. can use it as well :) I will just get this outside of build step.

.github/workflows/smoke.yml
```diff
@@ -0,0 +1,57 @@
...
+      run: ./.github/scripts/install_dep.sh
+    
+    - name: Setup OLM
+      run: ./.travis/setup-olm.sh
+
+    - name: Build
+      run: ./.github/scripts/build.sh
+
+    - name: Smoke tests
+      run: ./.travis/test.sh
```
lulf: Perhaps copy this to .github/scripts (leave the original as is for now..) so that its all in the same place?

.github/workflows/release.yml
```diff
@@ -0,0 +1,126 @@
...
+        path: artifacts
+
+
+  smoke:
+    name: Smoke tests
+    needs: unit
+    env:
+      DOCKER_REGISTRY: "quay.io"
+      TAG: "latest"
+      SYSTEMTEST_PROFILE: smoke
```
lulf: No need to define this (it feels wrong to define this when test.sh tests both smoke tests and OLM

PR #3540, opened 2019-12-04
Allow changing authentication service of an address space

lulf:


k-wall:
Just recording my comment made in another channel:

Is the comparison of the authservice name sufficient?  what if I updated say a host/port number on an external authservice? I don't think that would get detected/propagated.

lulf:
@k-wall See https://github.com/EnMasseProject/enmasse/pull/3543

Issue #3542, opened 2019-12-04
Addresses stuck in configuring after kubernetes api server restart


lulf:
**Describe the bug**
During Kubernetes API server restarts, the admin/agent pod seems to fail reconnecting its watch. Until restarted, new addresses are not provisioned. 

The steps to reproduce should be considered a working hypothesis for reproducing the issue and have not been confirmed.

**To Reproduce**
1. Create address space and wait until ready
2. Restart openshift api server(s)
3. Look for watch failure in agent log
3. Try to create new addresses and see them not getting provisioned

**Expected behavior**
Address is provisioned

**Additional context**
[agent.log](https://github.com/EnMasseProject/enmasse/files/3922879/agent.log)

lulf:
I was able to trigger a similar issue locally. However, this time it was the standard-controller that never got notified of address creation after the restart, until it reached the resync point where the watch is recreated.

I think we should add a similar logic to the agent, i.e. recreate any watches at a specific resync interval (every 5 or 10 minutes) to recover from this kind of scenario.

k-wall:
I'm also wondering if we should have the `error` event in agent's `Watcher.watch`restart the watch. The coding would need to take care not to start two watches if both `end` and `closed` get fired.

k-wall:
The underlying reason that the watch 'stuck' remains unclear, however #3544 means that a similarly stuck watch will be automatically aborted and restarted by the period resync

PR #3544, opened 2019-12-05
Add resync period for agent watches

lulf:
### Type of change

- Bugfix

### Description

This adds periodic resync intervals to agent watches similar to what
exists for other controller processes.

On watch error, abort current request and reconnect.

Issue #3542

### Checklist

- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/agent/lib/kubernetes.js
+++ b/agent/lib/kubernetes.js
@@ -57,6 +57,7 @@ function get_options(options, path) {
     return {
         hostname: options.host || process.env.KUBERNETES_SERVICE_HOST,
         port: options.port || process.env.KUBERNETES_SERVICE_PORT,
+        resyncInterval: options.resyncInterval || process.env.RESYNC_INTERVAL,
         rejectUnauthorized: false,
         path: options.path || path,
         headers: {
@@ -238,7 +239,14 @@ Watcher.prototype.watch = function () {
             }
         });
     });
-    request.on('error', function(e) { log.error('error on watch: %s', e); });
+    request.on('error', function(e) {
+        log.error('error on watch: %s', e);
+    });
+    if (opts.resyncInterval !== undefined) {
+        request.setTimeout(opts.resyncInterval * 1000, function () {
+            request.abort();
+        });
+    }
 };
 
 function matcher(object) {
--- a/agent/lib/kubernetes.js
+++ b/agent/lib/kubernetes.js
@@ -232,7 +232,7 @@ Watcher.prototype.watch = function () {
         response.on('data', watch_handler(self));
         response.on('end', function () {
             if (!self.closed) {
-                log.debug('response ended; reconnecting...');
+                log.debug('response %s ended; reconnecting...', opts.path);
                 self.list();
             } else {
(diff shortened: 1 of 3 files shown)


lulf:

agent/lib/kubernetes.js
```diff
@@ -238,7 +239,22 @@ Watcher.prototype.watch = function () {
             }
         });
     });
-    request.on('error', function(e) { log.error('error on watch: %s', e); });
+    request.on('error', function(e) {
+        log.error('error on watch: %s', e);
+        if (!self.closed) {
```
grs: The error event will generally be immediately followed by an end event which will immediately trigger the list. This code will also schedule a delayed list in that case, which to me at least is confusing.

agent/lib/kubernetes.js
```diff
@@ -238,7 +239,22 @@ Watcher.prototype.watch = function () {
...
+            request.abort();
+            setTimeout(self.list.bind(self), self.delay * 1000);
+            self.delay = Math.min(30, self.delay + 1);
+        }
+    });
+    if self.options.resyncInterval !== undefined {
+        setTimeout(function () {
```
grs: I think setting a timeout on the request itself (i.e. request.setTimeout(self.options.resyncInterval * 1000, function () {self.abort(); }); ) might be cleaner. The watch is generally ended by the server anyway (and would be ended by any network failure), but adding the client timeout would ensure the client set a maximum idle time, which I assume is what is wanted here.

lulf: I was thinking to follow the same pattern of periodic re-watch as done in kubernetes controllers, as it seems more robust in cases where the expected events not to arrive (either due to a bug in the code or underlying dependencies or even bugs in the OS network layer).

agent/lib/kubernetes.js
lulf: We have seen it to be the case that end might not be called, as there are no new watches set (see https://github.com/EnMasseProject/enmasse/issues/3542). This is intended to catch that case.

lulf:

agent/lib/kubernetes.js
grs: Understood, but your code will run even if end *is* called, resulting in a bifurcation of sorts. If you have a timeout on the request, then whether end is called or not, the request can be aborted.

agent/lib/kubernetes.js
grs: All I'm suggesting is that instead of doing your own timeout independent of the state of the request, you use the inbuilt timeout on the http agent. To me that seems more predictable and I don't as yet see any way in which it is less robust.

Simply a suggestion however.

lulf: That makes sense, I will change it.

agent/lib/kubernetes.js
lulf: Good point, maybe leave this out  for now and rely on the client timeout for now then.

lulf:

PR #3552, opened 2019-12-06
Temp workaround for OLM tests

kornys:


kornys:
@enmasse-ci run tests profile=acceptance testcase=olm.** ocp4 test_only OLM

PR #3553, opened 2019-12-06
Prevents NetworkPolicyTestStandard leaving behind its infraconfig

k-wall:
NetworkPolicyTestStandard was leaving its standardinfraconfig resource behind.  This change ensures that it gets cleaned up during tear down.


kornys:
@k-wall nice catch!

PR #3561, opened 2019-12-06
Add protocol adapter options for tenant timeout and max payload

ctron:
### Type of change

- Enhancement / new feature

### Description

This primarily adds a way to set the "tenant timeout", and provides a default value. It also allows to set the max payload size.

### Checklist

- [ ] Update documentation
- [ ] Update CHANGELOG.md


ctron:

ctron:

pkg/controller/iotconfig/adapter.go
```diff
@@ -203,6 +225,10 @@ func AppendHonoAdapterEnvs(config *iotv1alpha1.IoTConfig, container *corev1.Cont
 
 }
 
+func appendAdapterEnvVar(container *corev1.Container, a adapter, key string, value string) {
```
lulf: Is it really appending or is applying more correct?

ctron: Good question … I know the underlying method is named "apply or remove" … however the result will be that a new env-var entry will be added to the list of existing  env-vars.

Also is the statement contained in the method `AppendHonoAdapterEnvs` and the next calls which follows is named  `AppendTrustStores`.

PR #3564, opened 2019-12-09
Use existing constant

ctron:
### Type of change

- Refactoring

### Description

Use constant instead of plain string.

### Checklist

- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/systemtests/src/main/java/io/enmasse/systemtest/utils/IoTUtils.java
+++ b/systemtests/src/main/java/io/enmasse/systemtest/utils/IoTUtils.java
@@ -8,6 +8,7 @@ import io.enmasse.address.model.AddressSpace;
 import io.enmasse.iot.model.v1.AdapterConfig;
 import io.enmasse.iot.model.v1.AdaptersConfig;
 import io.enmasse.iot.model.v1.IoTConfig;
+import io.enmasse.iot.model.v1.IoTCrd;
 import io.enmasse.iot.model.v1.IoTProject;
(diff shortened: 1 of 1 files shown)


ctron:

systemtests/src/main/java/io/enmasse/systemtest/utils/IoTUtils.java
```diff
@@ -186,7 +187,7 @@ private static void waitForIoTProjectDeleted(Kubernetes kubernetes, IoTProject p
     }
 
     public static boolean isIoTInstalled(Kubernetes kubernetes) {
-        return kubernetes.getCRD("iotprojects.iot.enmasse.io") != null;
+        return kubernetes.getCRD(IoTCrd.project().getMetadata().getName()) != null;
```
lulf: I believe this check might not be accurate (depending on what its used for), because the iot project CRD is installed even if IoT is not installed: https://github.com/EnMasseProject/enmasse/tree/master/templates/crds

ctron: Good point. But then I think @famartinrh or @kornys should take a look. I was simply trying to re-use an existing constant, rather than adding a new string.

kornys: good point I will look at it

kornys: it will be fixed in different PR

PR #3568, opened 2019-12-10
Add status subresources

lulf:


lulf:

lulf:
@enmasse-ci run tests profile=smoke testcase=standard.SmokeTest

k-wall:
@lulf what's the issue we are solving by this PR?  Please link to an issue containing some background.   Will this apply to other CRDs too e.g. address spaces?

lulf:
@k-wall This is an experiment with status subresources.  Adding subresources support will allow controllers to update only the status field through a separate URL path on the resource ("/status"), and allow creating permissions so that only controllers can be permitted to update the status field.

I added this to the new messaginguser CRD already, and thought it would probably be good if we added it to the address and addressspace CRDs now while we have not yet released them (in case there are some gotchas with adding them after the fact).

See https://book-v1.book.kubebuilder.io/basics/status_subresource.html for more details.

lulf:
@k-wall Also, an update to status will not change metadata.generation, avoiding triggering the reconciliation loop after an update.

k-wall:
> @k-wall This is an experiment with status subresources. Adding subresources support will allow controllers to update only the status field through a separate URL path on the resource ("/status"), and allow creating permissions so that only controllers can be permitted to update the status field.
> 

Ok, it always sense to follow least privilege

> I added this to the new messaginguser CRD already, and thought it would probably be good if we added it to the address and addressspace CRDs now while we have not yet released them (in case there are some gotchas with adding them after the fact).
> 
> Also, an update to status will not change metadata.generation, avoiding triggering the reconciliation loop after an update.
> 
> See https://book-v1.book.kubebuilder.io/basics/status_subresource.html for more details.
> 
> However, it seems like the status subresource is not supported by kubernetes-client.

Yes, I'd noticed some defect reports from the fabric8 project.

lulf:
Closing this, as it will probably be a while until we can enable status subresource

PR #3581, opened 2019-12-11
Update log levels

lulf:
### Type of change

- Refactoring

### Description

* Use permitted log levels.
* Allow changing level using env var.

Future work would involve to add a mechanism for modifying the logLevel
variable dynamically (for instance via http).

### Checklist

- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/templates/enmasse-operator/050-Deployment-enmasse-operator.yaml
+++ b/templates/enmasse-operator/050-Deployment-enmasse-operator.yaml
@@ -33,6 +33,8 @@ spec:
         image: ${CONTROLLER_MANAGER_IMAGE}
         imagePullPolicy: ${IMAGE_PULL_POLICY}
         env:
+        - name: LOG_LEVEL
+          value: debug
         - name: POD_NAME
           valueFrom:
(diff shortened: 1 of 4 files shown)


templates/enmasse-operator/050-Deployment-enmasse-operator.yaml
```diff
@@ -33,6 +33,8 @@ spec:
         image: ${CONTROLLER_MANAGER_IMAGE}
         imagePullPolicy: ${IMAGE_PULL_POLICY}
         env:
+        - name: LOG_LEVEL
```
ctron: I think the default should stay "info".

lulf: I was thinking to leave it at debug for a few days in case there are more issues popping up in CI, and then move it back to info once it is stable.

ctron: … and then to forget about it :grin:

PR #3626, opened 2020-01-07
Use plugins provided by base image

lulf:
### Type of change

- Enhancement / new feature
- Refactoring

### Description

Use the plugins bundled with the Artemis base image.

Remove jmx_exporter now that base image includes prometheus plugin.

### Checklist

- [ ] Update documentation
- [ ] Update CHANGELOG.md


broker-plugin/plugin/src/main/resources/bootstrap.xml
```diff
@@ -34,6 +34,9 @@
        <app url="console" war="console.war"/>
    </web>
 
+   <web bind="http://0.0.0.0:8080" path="web">
```
k-wall: How does the path element `web` work?   

This doesn't appear to match what is advertised via the deployment/statefulset broker annotations?

         prometheus.io/path: /metrics
          prometheus.io/port: '8080'
          prometheus.io/scrape: 'true'

prometheus.io/path

Makefile.env.mk
```diff
@@ -40,7 +40,7 @@ ALERTMANAGER_IMAGE ?= prom/alertmanager:v0.15.2
 GRAFANA_IMAGE ?= grafana/grafana:5.3.1
 APPLICATION_MONITORING_OPERATOR_IMAGE ?= quay.io/integreatly/application-monitoring-operator:0.0.27
 KUBE_STATE_METRICS_IMAGE ?= quay.io/coreos/kube-state-metrics:v1.4.0
-BROKER_IMAGE ?= quay.io/enmasse/artemis-base:2.10.1
+BROKER_IMAGE ?= quay.io/enmasse/artemis-base:5a47cee4
```
k-wall: Is using a SHA here just a temporary measure?

broker-plugin/plugin/src/main/resources/bootstrap.xml
lulf: Its actually the path in the filesystem for looking up the .war file (which I discovered eventually :D)

Makefile.env.mk
lulf: Yes, temporary until the base image PR is merged (but using this hash now so that i'm able to run tests).

broker-plugin/plugin/src/main/resources/bootstrap.xml
k-wall: Ah..  that kind of path!   So I guess this means the plugin just knows to bind to /metrics by default?

lulf: Thats the 'url' field in the app element. I bound it to / , but I can bind it to /metrics instead.

lulf:
@robshelly Could you have a look at the monitoring configuration? I'm not sure about the naming (health-tls) for TLS endpoints etc. Ideally it should trust some CA, but we don't have a service-wide CA right to trust right now.

lulf:

lulf:

PR #3631, opened 2020-01-09
Collect info about pv and pvc

kornys:


kornys:
@k-wall I agree, fixed in last commit

PR #3634, opened 2020-01-09
Remove debug to check if finalizer issue is fixed

lulf:


kornys:
@lulf can we also remove verbose level for delete command?

lulf:
@kornys I reduced verbosity, but would like to keep the current level just in case we run into the bug

PR #3636, opened 2020-01-10
Add debug option for infinispan device registry

jbtrystram:
### Type of change

- Bugfix
- Enhancement / new feature
- Refactoring
- Documentation

### Description

Enable debug output in the infinispan device registry with :
```
kind: IoTConfig
spec:
  services:
    deviceRegistry:
      infinispan:
        java:
          debug: true
```

### DIFF
--- a/pkg/controller/iotconfig/utils.go
+++ b/pkg/controller/iotconfig/utils.go
@@ -62,21 +62,13 @@ func AppendEnvVarValue(container *corev1.Container, name string, value string) {
 	})
 }
 
-func AppendStandardHonoJavaOptionsToEnv(container *corev1.Container, debug bool) {
+func AppendStandardHonoJavaOptions(container *corev1.Container) {
 
 	opts := "-Djava.net.preferIPv4Stack=true -Dvertx.logger-delegate-factory-class-name=io.vertx.core.logging.SLF4JLogDelegateFactory"
 
-	if debug {
-		opts += " -Ddebug"
-
-	}
 	AppendEnvVarValue(
-			container,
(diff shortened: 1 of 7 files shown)


jbtrystram:

pkg/controller/iotconfig/utils.go
```diff
@@ -62,12 +62,21 @@ func AppendEnvVarValue(container *corev1.Container, name string, value string) {
...
+	if debug {
+		opts += " -Ddebug"
+
+	}
 	AppendEnvVarValue(
-		container,
-		"JAVA_APP_OPTS",
-		"-Djava.net.preferIPv4Stack=true -Dvertx.logger-delegate-factory-class-name=io.vertx.core.logging.SLF4JLogDelegateFactory",
-	)
+			container,
```
ctron: I think you need to run `go fmt` against the changed files.

ctron:
I am struggling a bit with this PR. I think it is implemented well! IIRC we would only also need an additional port mapping.

The question is just, should it be that simple to enable this internal functionality. Having that enabled, would allow you direct access, with not authentication, to the credentials database. Which I would say is something that shouldn't be enabled that easily.

jbtrystram:
@ctron I agree with you, the original idea was to enable debug output to facilitate tests debugging ! The controller should alter the logback config file, but I wanted to see if that worked with an extra java argument.. And then I looked at the logs and realized that this was not the effect I expected :) 
That why there is no port mapping

jbtrystram:

jbtrystram:

PR #3652, opened 2020-01-13
Remove IoT project finalizers if delete fails

dejanb:
### Type of change

- Enhancement / new feature

### Description

Attempt to stabilze IoT tests, but making sure IoT project resources are removed

### DIFF
--- a/systemtests/src/main/java/io/enmasse/systemtest/utils/IoTUtils.java
+++ b/systemtests/src/main/java/io/enmasse/systemtest/utils/IoTUtils.java
@@ -18,6 +18,7 @@ import io.enmasse.systemtest.iot.MessageSendTester;
 import io.enmasse.systemtest.iot.MessageType;
 import io.enmasse.systemtest.logs.CustomLogger;
 import io.enmasse.systemtest.model.addressplan.DestinationPlan;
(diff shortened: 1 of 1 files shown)


dejanb:

lulf:
@dejanb @famartinrh Would it be possible instead to ensure that the device registry is deleted after the iotprojects etc. during cleanup?

dejanb:
@lulf I think there's still work to make sure tests clean up properly. This fix is just to make sure we don't mess the whole build/cluster when that happen. I put warn log in there, so we can track when this is happening and further investigate.

famartinrh:

famartinrh:
@lulf @dejanb so the issue here is that sometimes the iotproject fails to be deleted, I want to have a deeper look at this to try to determine why. This causes the infinispan deployment to not been deleted  and then causes more tests to fail, so I have a workaround for that here https://github.com/EnMasseProject/enmasse/tree/systemtests-deploy-infinispan-workaround , do you guys think it's a good idea to merge that? so we can get test runs with less noise and hopefully be able to find what happens to iotproject deletion

dejanb:
This is now included in #3683

PR #3664, opened 2020-01-14
Add uninstall procedure for OLM uninstall

lulf:
### Type of change

- Documentation

### Description

Documents the uninstall procedure when EnMasse is installed using operator hub.

### Checklist

- [x] Update documentation
- [ ] Update CHANGELOG.md


documentation/modules/proc-uninstalling-using-olm.adoc
```diff
@@ -0,0 +1,56 @@
...
+{cmdcli} login -u system:admin
+----
+endif::[]
+
+. Remove all IoTProject and AddressSpace instances:
++
+[options="nowrap",subs="+quotes,attributes"]
+----
+{cmdcli} delete iotprojects -A --all
+{cmdcli} delete addressspaces -A --all
```
famartinrh: what is this -A? it doesn't work for me

documentation/modules/proc-uninstalling-using-olm.adoc
```diff
@@ -0,0 +1,56 @@
...
+----
+{cmdcli} delete crd -l app=enmasse
+{cmdcli} delete oauthclients -l app=enmasse
+----
+
+. Delete the namespace where {ProductName} was installed:
++
+[options="nowrap",subs="+quotes,attributes"]
+----
+{cmdcli} delete namespace _{ProductNamespace}_
```
famartinrh: maybe some disclaimer to not apply this in case of installing into openshift-operators namespace?

documentation/modules/proc-uninstalling-using-olm.adoc
lulf: @famartinrh from oc delete --help:

```
  -A, --all-namespaces=false: If present, list the requested object(s) across all namespaces. Namespace in current
context is ignored even if specified with --namespace.
```

Which version of oc/kubectl are you using? Could be I need to change it to --all-namespaces=true, but I think its safe to require client >= 4.x

documentation/modules/proc-uninstalling-using-olm.adoc
lulf: Agreed, will add it.

documentation/modules/proc-uninstalling-using-olm.adoc
famartinrh: my oc version is 3.11, I agree on requiring client >= 4 having that this is for OLM

lulf:
@famartinrh Updated with console variant

PR #3668, opened 2020-01-15
[IoT] Prevent tenant cleaner to update caches schema

dejanb:
### Type of change

- Bugfix

### Description

Cache schema check is not consistent when the schema is updated multiple times. The schema is the same, just with the reordered elements. The quick fix is to prevent tenant cleaner job to update the schema (as I think it shouldn't do anyways). But we might want to revisit how we do this check in the future anyways.

### Checklist

- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/iot/iot-service-base/src/main/java/io/enmasse/iot/service/base/AbstractProjectBasedService.java
+++ b/iot/iot-service-base/src/main/java/io/enmasse/iot/service/base/AbstractProjectBasedService.java
@@ -109,8 +109,8 @@ public abstract class AbstractProjectBasedService extends AbstractKubernetesBase
 
     protected void stopWatcher() {
         if (this.factory != null) {
-            this.factory = null;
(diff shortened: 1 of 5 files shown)


dejanb:

iot/iot-service-base/src/main/java/io/enmasse/iot/service/base/AbstractProjectBasedService.java
```diff
@@ -109,8 +109,8 @@ public void onAdd(final IoTProject obj) {
 
     protected void stopWatcher() {
         if (this.factory != null) {
-            this.factory = null;
```
ctron: :man_facepalming: Good catch!

ctron:

ctron:
It looks like something fails to clean up properly … I don't think this is related to the changes in this PR, but another issue.

PR #3672, opened 2020-01-16
Created mutations for add, delete, and patch

vbusch:
Created mutations for Addresses & Address Spaces.  Implement Create, Patch, and Delete

### Type of change

- Bugfix
- Enhancement / new feature
- Refactoring
- Documentation

### Checklist

- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/cmd/console-server/main.go
+++ b/cmd/console-server/main.go
@@ -81,8 +81,17 @@ func authHandler(next http.Handler) http.Handler {
 			return
 		}
 
+		coreClient, err := enmassev1beta1.NewForConfig(config)
+		if err != nil {
+			log.Printf("Failed to build client set : %v", err)
+			rw.WriteHeader(500)
+			return
+		}
+
+
 		requestState := &server.RequestState{
 			UserInterface: userclientset.Users(),
+			EnmasseV1beta1Client: coreClient,
 		}
 
 		ctx := server.ContextWithRequestState(requestState, req.Context())
@@ -113,9 +122,16 @@ func developmentHandler(next http.Handler) http.Handler {
 			return
 		}
 
-		requestState := &server.RequestState{
-			UserInterface: userclientset.Users(),
+		coreClient, err := enmassev1beta1.NewForConfig(config)
+		if err != nil {
+			log.Printf("Failed to build client set : %v", err)
+			rw.WriteHeader(500)
+			return
 		}
+
+		requestState := &server.RequestState{
+			UserInterface:        userclientset.Users(),
+			EnmasseV1beta1Client: coreClient}
 
 		ctx := server.ContextWithRequestState(requestState, req.Context())
 		next.ServeHTTP(rw, req.WithContext(ctx))
--- a/pkg/consolegraphql/resolvers/resolver_address.go
+++ b/pkg/consolegraphql/resolvers/resolver_address.go
@@ -12,6 +12,7 @@ import (
 	"github.com/enmasseproject/enmasse/pkg/apis/enmasse/v1beta1"
 	"github.com/enmasseproject/enmasse/pkg/consolegraphql"
 	"github.com/enmasseproject/enmasse/pkg/consolegraphql/cache"
+	"github.com/enmasseproject/enmasse/pkg/consolegraphql/server"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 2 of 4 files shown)


pkg/apis/enmasse/v1beta1/types.go
```diff
@@ -206,7 +206,7 @@ type Address struct {
 	metav1.ObjectMeta `json:"metadata,omitempty"`
 
 	Spec   AddressSpec   `json:"spec"`
-	Status AddressStatus `json:"status,omitempty"`
+	Status *AddressStatus `json:"status,omitempty"`
```
k-wall: Why is this change necessary?   Having the status field a pointer isn't the common pattern for either our own objects (AddressSpace, AddressPlan etc), or for those of the Kubernetes core (Namespace, Pod etc).  Why deviate here?

cmd/console-server/main.go
```diff
@@ -67,7 +67,7 @@ Major items that are unfinished:
 
 
 */
-func authHandler(next http.Handler) http.Handler {
+func authHandler(next http.Handler, coreClient *enmassev1beta1.EnmasseV1beta1Client) http.Handler {
```
k-wall: Passing the existing coreClient argument to this function is wrong.  The coreClient used for the mutations needs to be associated with the bearer token of the web user in order that the actions are carried out by the user (rather than the console-service service account(.  The authHandler needs to create its own coreClient.

pkg/apis/enmasse/v1beta1/types.go
vbusch: I see this pattern a fair bit in k8s.io/api/core/v1/types.go, and occassionally in the current file.
type EndpointSpec struct {
	Name        string           `json:"name"`
	Service     string           `json:"service"`
	Certificate ***CertificateSpec** `json:"cert,omitempty"`
	Expose      ***ExposeSpec**      `json:"expose,omitempty"`
}

My golang knowledge isn't great.  But without using a pointer, it is difficult to tell the difference between an empty object and an unset object in go.  As a result, you end up with empty objects, instead of null objects.

When creating a new address we want the following to be sent:
```
apiVersion: enmasse.io/v1beta1
kind: Address
metadata:
    name: myspace.myqueue
spec:
    address: myqueue
    type: queue
    plan: standard-small-queue
```
But instead we end up with:
```
apiVersion: enmasse.io/v1beta1
kind: Address
metadata:
    name: myspace.myqueueess
spec:
    address: myqueueess
    type: queue
    plan: standard-small-queue
status:
   planStatus:
      resources: {}
```

lulf: Are we just missing an omitempty in the type annotation?

vbusch: Address type already had the omitempty for AddressStatus
```
type Address struct {
	metav1.TypeMeta   `json:",inline"`
	metav1.ObjectMeta `json:"metadata,omitempty"`

	Spec   AddressSpec   `json:"spec"`
	Status AddressStatus `json:"status,omitempty"`
}
```
Is there another type, that I should look at?

lulf: I think this is fixed on master now @vbusch by making the planStatus nullable. Actually, the same as your second change in this PR. So you can remove the pointer on AddressStatus I think, rebase  on master, and it should hopefully work.

pkg/apis/enmasse/v1beta1/types.go
```diff
@@ -236,7 +236,7 @@ type AddressStatus struct {
 	Phase          string            `json:"phase,omitempty"`
 	Messages       []string          `json:"messages,omitempty"`
 	BrokerStatuses []BrokerStatus    `json:"brokerStatus,omitempty"`
-	PlanStatus     AddressPlanStatus `json:"planStatus,omitempty"`
+	PlanStatus     *AddressPlanStatus `json:"planStatus,omitempty"`
```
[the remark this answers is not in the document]
lulf: I made the same change in #3689

k-wall:
As @lulf suggested, I rebased console-server against master.

k-wall:
@vbusch I took the liberty of rebasing your branch and incorporating the change to user the logged on user's coreClient

PR #3676, opened 2020-01-16
Add controller run target

lulf:


controller-manager/Makefile
```diff
@@ -3,6 +3,9 @@ include ../Makefile.go.mk
 
 build_deps: build/operatorImageMap.yaml
 
+run:
```
ctron: I think you should either add a source file (maybe both ./build/operatorImageMap.yaml ./build/controller-manager) or add the target to `.PHONY`.

PR #3678, opened 2020-01-16
Add ids to all elements for testing

anukritijha:
This PR includes ids in filters, buttons, drop-downs and input fields in all the source files to enable testing.


anukritijha:
@kornys Added ids to all elements.

kornys:
@k-wall can we merge it into console-server?

PR #3685, opened 2020-01-16
Fix: searchbar auto-populate issue

rkpattnaik780:
The filters now use 'LIKE' instead of '='. The filters have been modified in all of the 5 lists.

ToDos : 
* In some cases redundant elements are encountered in the dropdowns.


k-wall:
@rkpattnaik780   If the UI knows that it wants an exact match then an equality expression (=) should be used.  Use the LIKE clause only when it needs the pattern matching ability as it is more costly on the backend.

rkpattnaik780:
Okay @k-wall , the way I see it is that I should be using "LIKE" when the text is manually typed and use equality operation when text is received through a dropdown select. Is it?

k-wall:
> Okay @k-wall , the way I see it is that I should be using "LIKE" when the text is manually typed and use equality operation when text is received through a dropdown select. Is it?

Yes, that approach sounds reasonable.

PR #3695, opened 2020-01-17
Improve IoT project reconciliation; 

dejanb:
### Type of change

- Enhancement / new feature

### Description

Update status only if there are changes

### Checklist

- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/pkg/controller/iotproject/iotproject_controller.go
+++ b/pkg/controller/iotproject/iotproject_controller.go
@@ -8,6 +8,7 @@ package iotproject
 import (
 	"context"
 	"fmt"
+	"reflect"
 
 	corev1 "k8s.io/api/core/v1"
 
@@ -104,9 +105,9 @@ type ReconcileIoTProject struct {
 	scheme *runtime.Scheme
 }
 
-func (r *ReconcileIoTProject) updateProjectStatus(ctx context.Context, project *iotv1alpha1.IoTProject, currentError error) (reconcile.Result, error) {
+func (r *ReconcileIoTProject) updateProjectStatus(ctx context.Context, originalProject *iotv1alpha1.IoTProject, reconciledProject *iotv1alpha1.IoTProject, currentError error) (reconcile.Result, error) {
 
-	newProject := project.DeepCopy()
+	newProject := reconciledProject.DeepCopy()
 
 	// get conditions for ready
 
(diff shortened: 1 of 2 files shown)


dejanb:

dejanb:
With this change I don't see events being created. But now there's an error like

```2020-01-17T14:27:40.543Z WARN  [DefaultChannelPipeline] An exceptionCaught() event was fired, and it reached at the tail of the pipeline. It usually means the last handler in the pipeline did not handle the exception.
io.netty.channel.unix.Errors$NativeIoException: syscall:read(..) failed: Connection reset by peer
	at io.netty.channel.unix.FileDescriptor.readAddress(..)(Unknown Source)```

File: pkg/controller/iotproject/iotproject_controller.go [hunk 1]
```diff
@@ -261,9 +263,12 @@ func (r *ReconcileIoTProject) Reconcile(request reconcile.Request) (reconcile.Re
 		rc.Process(func() (result reconcile.Result, e error) {
 			return r.reconcileManaged(ctx, &request, project)
 		})
-		rc.Process(func() (result reconcile.Result, e error) {
-			return r.updateProjectStatus(ctx, project, rc.Error())
-		})
+		if !reflect.DeepEqual(project, original) {
+			reqLogger.Info("Project changed, updating status")
+			rc.Process(func() (result reconcile.Result, e error) {
```
ctron: Shouldn't the change be made on:

~~~go
	// update status
	err := r.client.Status().Update(ctx, newProject)
~~~

There are multiple paths leading to a call to `updateProjectStatus` and now I don't think we would properly update the status any longer.

lulf:
@dejanb FWIW, a similar change might be needed for IoTConfig controller, as it has the same symptoms at least.

pkg/controller/iotproject/iotproject_controller.go
dejanb: @ctron yeah ... This is the first thing that moved the needle in the local testing ... I want to see how it affects tests ... And will see how to better refactor all this if effective

dejanb:
@lulf sure thing. Will keep looking

dejanb:

dejanb:

dejanb:

PR #3698, opened 2020-01-20
Disruption budget broker

### LINKED ISSUE #3637: Add support for pod disruption budgets
**Description**

To prevent router downtime during node upgrades, the router standard infra config should allow configuring a pod distruption budget: https://kubernetes.io/docs/tasks/run-application/configure-pdb/

**Design**

The original syntax for PDB is:

```
apiVersion: policy/v1beta1
kind: PodDisruptionBudget
metadata:
  name: r-pdb
spec:
  minAvailable: 2
  selector:
    matchLabels:
      role: router
      infraUuid: 1234
```

However, exposing the selector in the infra config could be problematic, as the matchlabels would be different depending on which address space is created. Therefore, adding just 

`.spec.router.minAvailable` to the standard infra config should result in creating the pod distruption budget with the correct labels.

**Tasklist**

Enumerate the sub-tasks required to complete the task:

* [ ] Extend standard infraconfig CRD with minAvailable
* [ ] Create PodDistruptionBudget in address-space-controller with correct labels when minAvailable is specified.

lulf:
### Type of change

- Enhancement / new feature

### Description

Allow creating poddisruptionbudgets for brokers, and prefer to schedule brokers on separate hosts. This is mainly useful in combination with partitioned queues in order to get HA during node upgrades.

Fixes #3637 

### Checklist

- [ ] Update documentation
- [x] Update CHANGELOG.md

### DIFF
--- a/address-space-controller/src/main/java/io/enmasse/controller/PodDistruptionBudgetController.java
+++ b/address-space-controller/src/main/java/io/enmasse/controller/PodDistruptionBudgetController.java
@@ -1,0 +1,137 @@
+/*
+ * Copyright 2020, EnMasse authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+package io.enmasse.controller;
+
+import io.enmasse.address.model.AddressSpace;
+import io.enmasse.admin.model.v1.InfraConfig;
+import io.enmasse.admin.model.v1.StandardInfraConfig;
+import io.enmasse.config.AnnotationKeys;
+import io.fabric8.kubernetes.api.model.IntOrString;
+import io.fabric8.kubernetes.api.model.policy.PodDisruptionBudget;
+import io.fabric8.kubernetes.api.model.policy.PodDisruptionBudgetBuilder;
+import io.fabric8.kubernetes.client.KubernetesClientException;
+import io.fabric8.kubernetes.client.NamespacedKubernetesClient;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class PodDistruptionBudgetController implements Controller {
+    private static final Logger log = LoggerFactory.getLogger(PodDistruptionBudgetController.class);
+
+    private final NamespacedKubernetesClient client;
+    private final String namespace;
+
+    public PodDistruptionBudgetController(NamespacedKubernetesClient client, String namespace) {
+        this.client = client;
+        this.namespace = namespace;
+    }
+
+    public AddressSpace reconcileActive(AddressSpace addressSpace) throws Exception {
+        InfraConfig infraConfig = InfraConfigs.parseCurrentInfraConfig(addressSpace);
+
+        if (infraConfig instanceof StandardInfraConfig) {
+            RouterSet routerSet = RouterSet.create(namespace, addressSpace, client);
+            reconcileRouterPodDisruptionBudget(addressSpace, routerSet, (StandardInfraConfig) infraConfig);
+            reconcileBrokerPodDisruptionBudget(addressSpace, (StandardInfraConfig) infraConfig);
+        }
+        return addressSpace;
+    }
+
+    private static boolean needsUpdate(IntOrString existing, Integer updated) {
+        return updated != null && (existing == null || !updated.equals(existing.getIntVal()));
+    }
+
+
+    private void reconcileRouterPodDisruptionBudget(AddressSpace addressSpace, RouterSet routerSet, StandardInfraConfig infraConfig) {
+        if (infraConfig.getSpec() != null && infraConfig.getSpec().getRouter() != null && (infraConfig.getSpec().getRouter().getMinAvailable() != null || infraConfig.getSpec().getRouter().getMaxUnavailable() != null)) {
+            String name = String.format("enmasse.%s.%s.qdrouterd", addressSpace.getMetadata().getNamespace(), addressSpace.getMetadata().getName());
+            try {
+                boolean changed = false;
+                PodDisruptionBudget podDisruptionBudget = client.inNamespace(namespace).policy().podDisruptionBudget().withName(name).get();
+                if (podDisruptionBudget == null) {
+                    podDisruptionBudget = new PodDisruptionBudgetBuilder()
+                            .editOrNewMetadata()
+                            .withName(name)
+                            .addToLabels("app", "enmasse")
+                            .endMetadata()
+                            .editOrNewSpec()
+                            .endSpec()
+                            .build();
+                    changed = true;
+                }
+
+                Integer minAvailable = infraConfig.getSpec().getRouter().getMinAvailable();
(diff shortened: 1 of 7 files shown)


lulf:

rgodfrey:
@lulf how much does pod disruption budget actually help for brokers?

If we have 10000 queues, each using 0.002 broker, with each queue having a min partitions of 2, we'd presumably have 20 brokers...  Depending on how the queues were distributed the maximum disruption through which we could maintain 100% availability could be 10, or it could only be 1 (and in fact the latter case is probably the most likely).  However, if the maximum number of brokers we can ever guarantee to support loss of while remaining HA is 1, then we also need to make the pod anti affinity mandatory (and probably set the broker CPU/memory/storage requirements higher to "fill" a node).

lulf:
@rgodfrey I think it depends on the system, but I think the disruption budget only makes sense if you have brokers assigned to separate nodes and you have maxUnavailable: $numpartitionsinqueues - 1. The PR also contains a commit to prefer scheduling brokers on separate nodes.

rgodfrey:
@lulf Yeah - ultimately I think we probably need to provide more detailed guidance on setting up HA-ish instances.  Basically I think it probably never makes sense to have more than one broker for the same address space being on the same node.  The other, more complicated case, is trying to set up HA in the case of a full availability zone failure - for this we need the smarts in the queue allocation algorithm to ensure partitions are distributed.

lulf:
@rgodfrey Agreed. I wonder if its best to focus that effort for shared infra though.

address-space-controller/src/main/java/io/enmasse/controller/PodDistruptionBudgetController.java
```diff
@@ -0,0 +1,137 @@
...
+        this.namespace = namespace;
+    }
+
+    public AddressSpace reconcileActive(AddressSpace addressSpace) throws Exception {
+        InfraConfig infraConfig = InfraConfigs.parseCurrentInfraConfig(addressSpace);
+
+        if (infraConfig instanceof StandardInfraConfig) {
+            RouterSet routerSet = RouterSet.create(namespace, addressSpace, client);
+            reconcileRouterPodDisruptionBudget(addressSpace, routerSet, (StandardInfraConfig) infraConfig);
+            reconcileBrokerPodDisruptionBudget(addressSpace, (StandardInfraConfig) infraConfig);
```
davidffrench: @lulf Could PDBs also be created for brokers in a BrokeredInfraConfig?

address-space-controller/src/main/java/io/enmasse/controller/PodDistruptionBudgetController.java
```diff
@@ -0,0 +1,137 @@
...
+                            .addToMatchLabels("role", "broker")
+                            .addToMatchLabels("infraType", "standard")
+                            .addToMatchLabels("infraUuid", infraUuid)
+                            .endSelector()
+                            .endSpec()
+                            .build();
+                    changed = true;
+                }
+
+                Integer minAvailable = infraConfig.getSpec().getRouter().getMinAvailable();
```
davidffrench: @lulf I believe this should be `.getBroker()` instead of `.getRouter()`

address-space-controller/src/main/java/io/enmasse/controller/PodDistruptionBudgetController.java
```diff
@@ -0,0 +1,137 @@
...
+                if (needsUpdate(podDisruptionBudget.getSpec().getMinAvailable(), minAvailable)) {
+                    podDisruptionBudget.getSpec().setMinAvailable(new IntOrString(minAvailable));
+                    changed = true;
+                }
+
+                Integer maxUnavailable = infraConfig.getSpec().getRouter().getMaxUnavailable();
```
davidffrench: @lulf I believe this should be `.getBroker()` instead of `.getRouter()`

address-space-controller/src/main/java/io/enmasse/controller/PodDistruptionBudgetController.java
```diff
@@ -0,0 +1,137 @@
...
+        if (infraConfig.getSpec() != null && infraConfig.getSpec().getRouter() != null && (infraConfig.getSpec().getRouter().getMinAvailable() != null || infraConfig.getSpec().getRouter().getMaxUnavailable() != null)) {
+            String name = String.format("enmasse.%s.%s.qdrouterd", addressSpace.getMetadata().getNamespace(), addressSpace.getMetadata().getName());
+            try {
+                boolean changed = false;
+                PodDisruptionBudget podDisruptionBudget = client.inNamespace(namespace).policy().podDisruptionBudget().withName(name).get();
+                if (podDisruptionBudget == null) {
+                    podDisruptionBudget = new PodDisruptionBudgetBuilder()
+                            .editOrNewMetadata()
+                            .withName(name)
+                            .addToLabels("app", "enmasse")
```
davidffrench: I might be missing something here, it doesn't look like there is any selectors specific to the router pods for a specific address space, like there is for the router PDB below?

address-space-controller/src/main/java/io/enmasse/controller/PodDistruptionBudgetController.java
lulf: No, it doesn't make sense there as you can never have more than 1 broker.

address-space-controller/src/main/java/io/enmasse/controller/PodDistruptionBudgetController.java
lulf: It gets set on line 76

address-space-controller/src/main/java/io/enmasse/controller/PodDistruptionBudgetController.java
lulf: Ouch, that is correct. Thanks!

address-space-controller/src/main/java/io/enmasse/controller/PodDistruptionBudgetController.java
lulf: You're right, thanks again.

address-space-controller/src/main/java/io/enmasse/controller/PodDistruptionBudgetController.java
davidffrench: Ok, perfect. That makes perfect sense. I wasn't as familiar with brokered address spaces and didn't realise it could only create one broker.

address-space-controller/src/main/java/io/enmasse/controller/PodDistruptionBudgetController.java
davidffrench: Ah sorry, missed that. Thank you

lulf:
@davidffrench No worries, always happy to accept additional reviews (and you caught an important bug as well). We'll be adding systemtests for this feature next.

davidffrench:
> @davidffrench No worries, always happy to accept additional reviews (and you caught an important bug as well). We'll be adding systemtests for this feature next.

Always happy to help out when I can. Thanks for answers

PR #3699, opened 2020-01-20
Fix uploading release assets

obabec:
### Type of change

- Bugfix

### Checklist

- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -109,7 +109,7 @@ jobs:
           path: artifacts
 
       - name: Create Release
-        if: job.status == 'success' && contains(github.ref, 'refs/tags/0.')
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -109,7 +109,7 @@ jobs:
           path: artifacts
 
       - name: Create Release
-        if: job.status == 'success' && contains(github.ref, 'refs/tags/v')
+        if: job.status == 'success' && contains(github.ref, 'refs/tags/0.')
```
lulf: What happens when we go to 1.0? :)

obabec: Fixed, master push has different ref so there is no need at all to check last section of that.

PR #3701, opened 2020-01-20
Let the tenant cleaner upload, but not override the protobuf schema

ctron:
This change tries to fix the issue when the tenant cleaner tries to
clean up a tenant from an infinispan cluster, which never got
the protobuf schema uploaded by the device registry.

In change 4824dbb7ae73eda147fccf9cb0b19553ed8f6699 the tenant cleaner
was changes to never upload the protobuf schema. But this may cause
the problem, that when an IoTProject is being created and deleted
without the device registry properly uploading the schema, then the
tenant cleaner will fail.

While in production the argument could be, that the device registry
should be brought up online, before starting to clean up IoTProjects,
in testing this might be an issue, especially when the test environment
gets broken by that situation.

So with this change, when the device registry is configured to upload
the schema, it will overwrite it (as before). While the tenant cleaner
will only set the schema if it does not exists.

### Type of change

- Bugfix
- Enhancement / new feature
- Refactoring
- Documentation

### Checklist

- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/iot/iot-infinispan-base/src/main/java/io/enmasse/iot/infinispan/cache/AbstractCacheProvider.java
+++ b/iot/iot-infinispan-base/src/main/java/io/enmasse/iot/infinispan/cache/AbstractCacheProvider.java
@@ -16,6 +16,7 @@ import org.infinispan.client.hotrod.RemoteCacheManager;
 import org.infinispan.client.hotrod.configuration.ConfigurationBuilder;
 import org.infinispan.client.hotrod.configuration.ServerConfigurationBuilder;
 import org.infinispan.configuration.cache.Configuration;
+import org.infinispan.query.remote.client.ProtobufMetadataManagerConstants;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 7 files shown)


ctron:

jbtrystram:

ctron:

ctron:

iot/iot-infinispan-base/src/main/java/io/enmasse/iot/infinispan/cache/AbstractCacheProvider.java
```diff
@@ -82,6 +83,25 @@ public void close() throws Exception {
...
+        if (!this.properties.isUploadSchema()) {
+            return;
+        }
+
+        var cache = this.remoteCacheManager
+                .getCache(ProtobufMetadataManagerConstants.PROTOBUF_METADATA_CACHE_NAME);
+
+        if (this.properties.isOverrideSchema()) {
+            log.info("Uploading (put) schema - name: {}, schema: {}", fileName, schema);
+            cache.put(fileName, schema);
```
dejanb: It'd be good to log the result of the operations for the easier debugging (as the future improvement)

PR #3704, opened 2020-01-20
Safer cleanup in case GOPATH points to the wrong place

lulf:
Does it make sense? I don't fully understand how a clean could go wrong, but I am wondering if using some uncommon make version could cause it.


ctron:
I don't think we need `GOPATH` any longer … since we switched to go modules.

PR #3707, opened 2020-01-20
remove address space specific console endpoint

k-wall:
also:

* switched agent from the (now redundant) external console certificate to the internal agent/admin certificate.
* remove a few bits of dead implementation/config


k-wall:
One thing that is missing is the removal of the console route in the upgrade case.  I think this has to be a responsibility of the io.enmasse.controller.CreateController#reconcileActive to remove any orphaned infra resources.  Does this sound like the right approach? @lulf

address-space-controller/src/main/java/io/enmasse/controller/CreateController.java
```diff
@@ -273,4 +284,29 @@ private InfraConfig getInfraConfig(AddressSpace addressSpace) {
     public String toString() {
         return "CreateController";
     }
+
+    private Set<HasMetadata> getExistingResources(String infraUuid) {
+        Set<HasMetadata> existingResources = Collections.emptySet();
+        if (infraUuid != null) {
+            existingResources = kubernetes.getResources(infraUuid).stream().filter(m -> m.getMetadata().getAnnotations() == null || !m.getMetadata().getAnnotations().containsKey(AnnotationKeys.CLUSTER_ID)).collect(Collectors.toSet());
```
lulf: Wouldn't this accidentally delete resources that are not created in the createInfraResources? I.e. RouterConfigController, NetworkPolicyController, AuthController etc. all create resources with infra uuid.

k-wall: Yeah, you are right.  I failed to consider those.   I was fixating on the broker resources which are identifiable by their annotation.   I don't see a clean way to identify only those the were created from the infra template in order to handle this generally.

lulf: Ok, so maybe for now, we just identify which resources was created that are no longer needed. Is it the client secret for the console that is removed?

k-wall: No, the oauth client secret remains required.    The resources I need to delete are the console-xxxxxxx service and the console-xxxxxx route.

address-space-controller/src/main/java/io/enmasse/controller/EndpointController.java
```diff
@@ -30,22 +33,142 @@
...
+
+                CertSpec cert = endpoint.getCert();
+                if (cert != null && "selfsigned".equals(cert.getProvider()) && cert.getSecretName() != null) {
+                    Secret secret = client.secrets().inNamespace(namespace).withName(cert.getSecretName()).get();
+                    if (secret != null && secret.getMetadata().getLabels() != null && infraUuid.equals(secret.getMetadata().getLabels().get(LabelKeys.INFRA_UUID))) {
+                        client.secrets().inNamespace(namespace).delete(secret);
+                    }
+                }
+
+                if (isOpenShift) {
```
lulf: I believe that in the case of !isOpenShift, there is an console-$uuid-external service that should be removed as well.

k-wall:
In the end, I gave all the responsibilities for removing the console endpoint and its resource to the EndpointController.   I realise have it delete the service is slightly odd, as it did create it (the CreateController did), but couldn't find a way to put the responsibility there without making the code a mess.   I also experimented with giving some of the responsibilities to the operator, but that felt worse.

address-space-controller/src/main/java/io/enmasse/controller/EndpointController.java
k-wall: Ah.. I am not handling loadbalancer case.. is this what you refer to?  I'll fix.

lulf: Yes

k-wall: Hopefully resolved by my last push.  I had to give responsibilities back to the CreateController as the remainder of the controller needed certificate references.

I've manually tested the upgrade case for both standard/brokered and confirmed that the console resources are removed.

PR #3715, opened 2020-01-21
Change test data to prevent test fails in CI mode

anukritijha:
### Type of change

- Bugfix

### Description

Changes test data based on the changed schema, to prevent react test fails in CI mode.


anukritijha:
@k-wall fyi, this should fix the issue in react tests (d36668b)

PR #3717, opened 2020-01-21
Start from empty object when extracting spec

lulf:
### Type of change

- Bugfix

### Description

This prevents the original object from being modified and cause issues
for other address comparisons happening when addresses are updated. This
happened when objects were created from Go-based code where status field
was initialized.

Remove check for unused annotation.

### Checklist

- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/agent/lib/internal_address_source.js
+++ b/agent/lib/internal_address_source.js
@@ -26,16 +26,9 @@ function extract_spec(def) {
     if (def.spec === undefined) {
         console.error('no spec found on %j', def);
     }
-    var o = myutils.merge(def.spec, {status:def.status});
+    var o = myutils.merge({}, def.spec, {status:def.status});
     o.name = def.metadata ? def.metadata.name : def.address;
-    if (def.metadata && def.metadata.annotations && def.metadata.annotations['enmasse.io/broker-id']) {
-        var broker_id = def.metadata.annotations['enmasse.io/broker-id'];
-        var cluster_id = def.metadata.annotations['cluster_id'];
-        if (cluster_id === undefined) {
-            cluster_id = broker_id;
-        }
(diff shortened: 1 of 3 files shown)


agent/lib/internal_address_source.js
```diff
@@ -26,16 +26,9 @@ function extract_spec(def) {
     if (def.spec === undefined) {
         console.error('no spec found on %j', def);
     }
-    var o = myutils.merge(def.spec, {status:def.status});
+    var o = myutils.merge(myutils.merge({}, def.spec), {status:def.status});
```
grs: Just fyi, you can call merge with an arbitrary number of arguments, so you could repalce the two calls with two args with one call of three args if you want here. (Not that it is wrong as is)

agent/lib/internal_address_source.js
```diff
@@ -26,16 +26,9 @@ function extract_spec(def) {
...
     o.name = def.metadata ? def.metadata.name : def.address;
-    if (def.metadata && def.metadata.annotations && def.metadata.annotations['enmasse.io/broker-id']) {
-        var broker_id = def.metadata.annotations['enmasse.io/broker-id'];
-        var cluster_id = def.metadata.annotations['cluster_id'];
-        if (cluster_id === undefined) {
-            cluster_id = broker_id;
-        }
-        o.allocated_to = [{clusterId: cluster_id, containerId: broker_id, state: 'Active'}];
-    }
+
```
k-wall: it looks like the enmasse.io/broker-id dealt with some backward compatibility.   Is it certain this can be safely removed/

lulf: Yes, it is backward compatability version older than 0.26. We have already removed the compatibility in standard-controller too.

lulf:

PR #3722, opened 2020-01-22
Fix - expose creation time in connection list and connection detail page

drcoolsanjeev:
### Description
Fix - The connection list and connection detail should expose the connection's timestamp to the user


anukritijha:
@drcoolsanjeev Can you please resolve the conflicts?

PR #3732, opened 2020-01-22
Limit objects that the user may see by filtering using cached SelfSubjectAccessReviews

k-wall:
Limits the namespace, address space, address objects visible to the logged on console user to those that the user has GET permission.   Connections are permissions by considering the permissions of the address space to which they belong.

### DIFF
--- a/pkg/consolegraphql/accesscontroller/access_controller.go
+++ b/pkg/consolegraphql/accesscontroller/access_controller.go
@@ -1,0 +1,116 @@
+/*
+ * Copyright 2020, EnMasse authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ *
+ */
+
+package accesscontroller
+
+import (
+	"github.com/enmasseproject/enmasse/pkg/consolegraphql/cache"
+	authv1 "k8s.io/api/authorization/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/client-go/kubernetes"
+)
+
+type AccessController interface {
+	CanRead(obj interface{}) (bool, error)
+	ViewFilter() cache.ObjectFilter
+}
+
+type AccessControlledResource interface {
+	GetControllingResourceAttributes() *authv1.ResourceAttributes
+}
+
+type allowAll struct {
+}
+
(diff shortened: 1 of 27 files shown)


k-wall:
@lulf @vbusch this is the new access controller that will filter the objects that the user can see. The access controller itself is complete with the exception of handling cache expiry.  This will be added tomorrow.  I also need to keep the access controller in the user's HTTP session so that access decisions are retained across requests.  comments appreciated,

pkg/consolegraphql/accesscontroller/access_controller.go
```diff
@@ -0,0 +1,116 @@
...
+
+func (a *allowAll) ViewFilter() cache.ObjectFilter {
+	return func(obj interface{}) (bool, bool, error) {
+		return true, true, nil
+	}
+}
+
+type kubernetesRbac struct {
+	clientSet         kubernetes.Interface
+	cachedPermissions map[authv1.ResourceAttributes]bool
```
lulf: Needs a sync.Mutex or RWLock to guard this as you mention in the TODO.

k-wall: Yeah - that's my idea.  The AccessController instance will be per HTTP session so concurrency through this object instance will be low (user agents can make concurrent HTTP requests for the same browsing session).  I this a Mutex will suffice.

k-wall:
I used https://github.com/alexedwards/scs for HTTP session to pass the access controller's state between requests from a single user.  The state is serialised by the library to a byte array and stored in memory, and reused by the next request.  The session is time bound for both length of session and idle timeout.  This will present sessions accumulating in memory and means RBAC decisions are periodically discarded.

There is a negative consequence of serialisation. If two requests for a single session race,  the results of one's SSAR may be lost.  The worst this could mean is the SSAR is request is repeated by a later call, although as requests are likely to be about the same namespace, this is actually unlikely.


My initial idea was https://www.gorillatoolkit.org but this wasn't a good fit.  It uses the cookie for storage (size limited) or an external storage provider (which we don't want).

pkg/consolegraphql/accesscontroller/access_controller.go
k-wall: AccessController ended up being single threaded with state injected into them.  I believe this answers the threading concerns.

PR #3735, opened 2020-01-23
Use correct config values

lulf:
CC @davidffrench

### DIFF
--- a/address-space-controller/src/main/java/io/enmasse/controller/PodDistruptionBudgetController.java
+++ b/address-space-controller/src/main/java/io/enmasse/controller/PodDistruptionBudgetController.java
@@ -114,13 +114,13 @@ public class PodDistruptionBudgetController implements Controller {
                     changed = true;
                 }
 
-                Integer minAvailable = infraConfig.getSpec().getRouter().getMinAvailable();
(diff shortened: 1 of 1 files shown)


address-space-controller/src/main/java/io/enmasse/controller/PodDistruptionBudgetController.java
```diff
@@ -114,13 +114,13 @@ private void reconcileBrokerPodDisruptionBudget(AddressSpace addressSpace, Stand
                     changed = true;
                 }
 
-                Integer minAvailable = infraConfig.getSpec().getRouter().getMinAvailable();
```
famartinrh: not directly related with this, but we should change the types of minAvailable and maxUnavailable to IntOrString, to get all the expressivity that allows the underliying PodDistruptionBudget, I already was prepairing a PR with that, so tell me please if I can continue and create my PR

lulf: @famartinrh Please continue :)

lulf:
Closing as it will be part of @famartinrh's PR

PR #3736, opened 2020-01-23
Add null checks for planStatus

anukritijha:
### Type of change

- Bugfix

### Description

Checks if planStatus is null and accordingly assigns partitions a value.


anukritijha:
@kornys fyi, this will fix the issue #1635

kornys:
@anukritijha ok, thx, merge it into console-server please

PR #3737, opened 2020-01-23
PodDistruptionBudget use correct config values

famartinrh:
Bugfix of PodDistructionBudget functionalities


templates/crds/standardinfraconfigs.crd.yaml
```diff
@@ -75,9 +75,9 @@ spec:
               type: object
               properties:
                 minAvailable:
-                  type: integer
+                  type: int-or-string
```
lulf: This is missing for the .spec.router minAvailable and maxUnavailable as well.

PR #3740, opened 2020-01-23
Add partitioned queues

davidffrench:
### Type of change
- Enhancement / new feature

### Description
Adding new example plans for partitioned/sharded queues. These are for the small, medium and large queues with updated description and partitions of 3. 

image

### Checklist
- [ ] Update documentation
- [x] Check RBAC rights for Kubernetes / OpenShift roles
- [x] Update CHANGELOG.md

### DIFF
--- a/templates/example-plans/020-AddressPlan-standard-large-partitioned-queue.yaml
+++ b/templates/example-plans/020-AddressPlan-standard-large-partitioned-queue.yaml
@@ -1,0 +1,16 @@
+apiVersion: admin.enmasse.io/v1beta2
+kind: AddressPlan
+metadata:
+  name: standard-large-partitioned-queue
+  labels:
+    app: enmasse
+spec:
+  displayName: Large HA Queue
+  displayOrder: 6
+  shortDescription: Creates a large HA queue sharing underlying brokers with other queues.
+  longDescription: Creates a large HA queue sharing underlying brokers with other queues. The queue is sharded accross multiple brokers for HA and improved performance. A sharded queue no longer guarantees message ordering.
+  addressType: queue
+  resources:
+    router: 0.1
+    broker: 1.0
+  partitions: 3
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## 0.31.0
-*  #3686: Add metrics endpoint to controller-manager
+*  Adding example partitioned/sharded queue example plans
+*  #3686: Add metrics endpoint to controller-managerxs
 *  #2714: Allow setting security context of pods using persistent volumes
 *  #3622: Manage iot projects in enmasse-operator and remove iot-operator deployment
 *  #3407: AddressSpace, Address and MessagingUser resources are now backed by Custom Resource Definitions (CRD) rather than api-server:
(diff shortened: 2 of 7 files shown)


templates/example-plans/020-StandardInfraConfig-default-with-mqtt.yaml
```diff
@@ -16,7 +16,7 @@ spec:
       storage: 2Gi
     addressFullPolicy: FAIL
   router:
-    minReplicas: 2
+    minReplicas: 3
```
lulf: This seems like an unrelated change. The minReplicas of routers does not affect queue partitions. The partitions setting will only scale up the broker, which is handled by the operator.

templates/example-plans/020-StandardInfraConfig-default.yaml
```diff
@@ -14,7 +14,7 @@ spec:
       storage: 2Gi
     addressFullPolicy: FAIL
   router:
-    minReplicas: 2
+    minReplicas: 3
```
lulf: Same as above.

templates/example-plans/020-StandardInfraConfig-default-with-mqtt.yaml
davidffrench: It is, was easier to include it in this PR but should really be a separate PR. would you like me to split it out?

templates/example-plans/020-AddressPlan-standard-large-partitioned-queue.yaml
```diff
@@ -0,0 +1,16 @@
...
+kind: AddressPlan
+metadata:
+  name: standard-large-partitioned-queue
+  labels:
+    app: enmasse
+spec:
+  displayName: Large HA Queue
+  displayOrder: 6
+  shortDescription: Creates a large HA queue backed by a dedicated broker.
+  longDescription: Creates a large HA queue backed by a dedicated broker. The queue is sharded accross multiple brokers for HA and improved performance. A sharded queue no longer guarantees message ordering.
```
rgodfrey: This "backed by a dedicated broker" isn't true, is it... If the total size is 1.0, but it uses 3 partitions, it will only be using 0.3333 of a broker 3 times

templates/example-plans/020-StandardInfraConfig-default-with-mqtt.yaml
rgodfrey: what's the rationale for having a different config with a minimum router replica of three anyway... why not 4, 5, 17... etc

lulf: I don’t see the point in increasing minreplicas beyond 2 @davidffrench

davidffrench: These are good points, 2 routers are sufficient for HA in these sample plans. I have removed this commit.

templates/example-plans/020-AddressPlan-standard-large-partitioned-queue.yaml
davidffrench: Good spotting @rgodfrey . I have updated the text now to 
```  
shortDescription: Creates a large HA queue sharing underlying brokers with other queues.
longDescription: Creates a large HA queue sharing underlying brokers with other queues. The queue is sharded accross multiple brokers for HA and improved performance. A sharded queue no longer guarantees
```

lulf:

davidffrench:
Thank you again for the quick review @lulf

Issue #3437, opened 2019-11-06
Allow monitoring resources to be created when using enmasse operator


david-martin:
**Overview**

Currently in Integreatly on OpenShift 3, Enmasse is installed via ansible bundle. For Integreatly on OpenShift 4 we are using the enmasse operator to install & manage it. Integreatly depends on monitoring resources (Alerts & Dashboards) that are specific to Enmasse. In the ansible bundle, there is a ‘monitoring’ var to instruct the installer to create these monitoring resources. However, in the enmasse operator there is no equivalent. This means Integreatly is missing valuable monitoring when running on OpenShift 4.

**Design**

TBD

Ideally there is a field in a CR (e.g. ConsoleService) that can be set accordingly so the operator creates these resources.

**Tasklist**

TBD

david-martin:
cc @lulf

lulf:
I think we should support installing the monitoring CRs in some way via the operator. Ideally, I think we should have an "enmasse" CRD where the current ConsoleService, IoTConfig, monitoring schema/settings and more could live. 

An alternative shorter term could perhaps be to add MONITORING_ENABLE environment variable to enmasse-operator.

wdyt @k-wall @ctron @vbusch @robshelly  ?

ctron:
I don't think having an über CRD makes sense. Different parts might evolve in different paces, having a single big CRD would make it more complex updating that.

And for IoT, we might end up at some point having multiple IoT configurations available.

Also I don't really see the benefit of having a big CRD, compared to having a modularized configuration.

lulf:
> I don't think having an über CRD makes sense. Different parts might evolve in different paces, having a single big CRD would make it more complex updating that.

That they evolve at different pace is fine, as long as things are released and tested together. Having it in a single CRD could make it easier to coordinate changes affecting multiple parts of the schema. I.e. common configuration applied to all enmasse components, even iot, could be kept in the same place. I can understand how IoT might be a bit different to things like console though.

> Also I don't really see the benefit of having a big CRD, compared to having a modularized configuration

I see the main benefit as making EnMasse simpler to understand for a new user or a service admin. Having to wrap your head around all our CRDs might make sense to us as developers who know this very well, but I think its harder for everyone else. IMO sometimes we should focus on user experience, not what makes sense to us from a code point of view.

What would be your preferred alternative to the uber CRD for monitoring?

ctron:
[...]

PR #3759, opened 2020-01-26
Update operator to install monitoring resources

obrienrobert:
### Type of change

- Enhancement / new feature

### Description

https://github.com/EnMasseProject/enmasse/issues/3437

### Checklist

- [ ] Update documentation
- [ ] Update CHANGELOG.md


cmd/controller-manager/templateHelper.go
```diff
@@ -0,0 +1,104 @@
+/*
+ * Copyright 2018-2019, EnMasse authors.
```
lulf: Update copyright to 2020, and move file to pkg/util for instance.

cmd/controller-manager/templateHelper.go
```diff
@@ -0,0 +1,104 @@
...
+// by the user in the custom resource
+func NewTemplateHelper(extraParams map[string]string) *TemplateHelper {
+
+	log.Info("Installing Monitoring Resources")
+
+	params := Parameters{Params: extraParams}
+
+	templatePath := "./templates/"
+	if _, err := os.Stat(templatePath); os.IsNotExist(err) {
+		templatePath = "../templates/build/enmasse-latest/install/bundles/monitoring"
```
lulf: Having this hardcoded seems wrong to me. If its needed for testing locally, I would prefer if the template path was exposed as an environment variable and set in the make run target in the controller-manager Makefile.

cmd/controller-manager/templateHelper.go
```diff
@@ -0,0 +1,104 @@
...
+	return &TemplateHelper{
+		TemplatePath: templatePath,
+		TemplateList: GetTemplateList(),
+		Parameters:   params,
+	}
+}
+
+func GetTemplateList() []string {
+	templateList := []string{
+		"010-PrometheusRules-kube-metrics.yaml",
```
lulf: Consider getting rid of the number prefix (that is only for ordering when installing via bundles), and moving them to controller-manager/src/main/resources/templates instead. I think it belongs there now.

templates/enmasse-operator/020-ClusterRole-enmasse-operator.yaml
```diff
@@ -21,3 +21,9 @@ rules:
     resources: [ "apiservices" ]
     resourceNames: [ "v1alpha1.enmasse.io", "v1beta1.enmasse.io", "v1alpha1.user.enmasse.io", "v1beta1.user.enmasse.io" ]
     verbs: [ "create", "update", "patch", "get", "list", "watch", "delete" ]
+  - apiGroups: [ "monitoring.coreos.com" ]
+    resources: [ "prometheusrules", "servicemonitors" ]
```
lulf: The same permissions must be added to olm-manifest/src/main/resources/latest/enmasse.clusterserviceversion.yaml so that the OLM operator works.

cmd/controller-manager/main.go
```diff
@@ -93,6 +94,17 @@ func main() {
...
 	}
 
+	serverClient, err := client.New(cfg, client.Options{})
+
+	// Install monitoring resources
+	monitoring := os.Getenv("ENABLE_MONITORING")
+	if monitoring == "true" {
+		err = installMonitoring(ctx, serverClient)
+		if err != nil {
+			fmt.Print(err)
```
lulf: Use logger instead.

cmd/controller-manager/main.go
```diff
@@ -266,4 +289,33 @@ func addMonitoringKeyLabelToOperatorService(ctx context.Context, cfg *rest.Confi
...
+			}
+		}
+	}
+
+	return nil
+}
+
+func hasApplicationMonitoring(config *rest.Config) (bool, error) {
+	dc := discovery.NewDiscoveryClientForConfigOrDie(config)
+	apiVersion := "applicationmonitoring.integreatly.org/v1alpha1"
```
lulf: I think checking for the ApplicationMonitoring CRD is a bit too narrow. Ideally it would install the service monitor and rules if just the prometheus operator CRDs were installed. 

Same with grafana I guess. Maybe just applying and ignoring the error is more flexible than checking for specific resources?

lulf:
@obrienrobert The ENABLE_MONITORING setting must also be set for the enmasse-operator deployment in ```olm-manifest/src/main/resources/latest/enmasse.clusterserviceversion.yaml```

lulf:
@obrienrobert Here is the docs PR: https://github.com/obrienrobert/enmasse/pull/1

lulf:
@obrienrobert One thing I noticed when writing the documentation is that if enmasse is installed before the monitoring stack, then one need to restart the enmasse-operator pod. I wonder if it would be good to periodically attempt to apply the resources, say every 5 minutes or so. Then also lower error to warning level. Wdyt?

documentation/modules/proc-enable-monitoring.adoc
```diff
@@ -0,0 +1,30 @@
+// Module included in the following assemblies:
+//
+// assembly-monitoring.adoc
+
+[id='deploy-monitoring-bundle-{context}']
+= Enable monitoring
+
+The simplest way to deploy monitoring is to enable the monitoring environment variable on the
+`enmasse-operator` deployment. This is only needed if you are not using a default install
```
jenmalloy:
```suggestion
`enmasse-operator` deployment. 
```

documentation/modules/proc-enable-monitoring.adoc
```diff
@@ -0,0 +1,30 @@
+// Module included in the following assemblies:
+//
+// assembly-monitoring.adoc
+
+[id='deploy-monitoring-bundle-{context}']
+= Enable monitoring
+
+The simplest way to deploy monitoring is to enable the monitoring environment variable on the
```
jenmalloy:
```suggestion
If you are not using a default installation configuration, the simplest way to deploy monitoring is to enable the monitoring environment variable on the
```

documentation/modules/proc-enable-monitoring.adoc
```diff
@@ -0,0 +1,30 @@
...
+`enmasse-operator` deployment. This is only needed if you are not using a default install
+configuration.
```
jenmalloy:
```suggestion
```

documentation/modules/proc-enable-monitoring.adoc
```diff
@@ -0,0 +1,30 @@
...
+.Procedure
+
+. Label the _{ProductNamespace}_ namespace:
++
+[options="nowrap",subs="+quotes,attributes"]
+----
+{cmdcli} label namespace _{ProductNamespace}_ monitoring-key=middleware
+----
+
+. Enable monitoring on the operator
```
jenmalloy:
```suggestion
. Enable monitoring on the operator:
```

documentation/modules/proc-enable-monitoring.adoc
```diff
@@ -0,0 +1,30 @@
+// Module included in the following assemblies:
+//
+// assembly-monitoring.adoc
+
+[id='deploy-monitoring-bundle-{context}']
+= Enable monitoring
```
jenmalloy:
```suggestion
= Enabling monitoring
```

obrienrobert:
@jenmalloy, thank you for your feedback. I've updated the doc as per your suggestions in the latest commit.

cmd/controller-manager/main.go
```diff
@@ -101,11 +102,20 @@ func main() {
...
 
 	if monitoringEnabled == "true" {
-		log.Info("Installing monitoring resources")
-
+		// Attempt to install monitoring resources on operator start
 		err = installMonitoring(ctx, serverClient)
+
+		ticker := time.NewTicker(5 * time.Second)
+
+		// Attempt to install monitoring resources every 5 minutes
```
lulf: Just a minor improvement: how about only creating the ticker if installMonitoring fails and complete/cancel it once it succeeds (inside the ticker loop)?

obrienrobert: @lulf can do. I've just noticed that the ticker time also needs to updated to 5 minutes.

kornys:
@lulf I think systemtests should be also fixed before we merge this PR

lulf:

lulf:
@enmasse-ci run tests profile=isolated testcase=MonitoringTest

lulf:
@enmasse-ci run tests profile=isolated testcase=MonitoringTest

PR #3787, opened 2020-01-28
Implemented the command queries

vbusch:
- returned the yaml required for running the commands on command-line
- Added authentication service type to the create addressspace

### Type of change

- Bugfix
- Enhancement / new feature
- Refactoring
- Documentation

### Checklist

- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/pkg/consolegraphql/resolvers/resolver_address.go
+++ b/pkg/consolegraphql/resolvers/resolver_address.go
@@ -187,5 +187,15 @@ func (r *mutationResolver) PurgeAddress(ctx context.Context, input metav1.Object
 }
 
 func (r *queryResolver) AddressCommand(ctx context.Context, input v1beta1.Address) (string, error) {
-	return "oc apply blah", nil
+
+	return `apiVersion: enmasse.io/v1beta1
+kind: Address
+metadata:
+  namespace: `+ input.Namespace +`
+  name: `+ input.Name +`
+spec:
+  address: `+ input.Spec.Address +`
+  addressSpace: `+ input.Spec.AddressSpace +`
+  type: `+ input.Spec.Type +`
+  plan: `+ input.Spec.Plan, nil
 }
--- a/pkg/consolegraphql/resolvers/resolver_address.go
+++ b/pkg/consolegraphql/resolvers/resolver_address.go
@@ -187,8 +187,8 @@ func (r *mutationResolver) PurgeAddress(ctx context.Context, input metav1.Object
 }
 
 func (r *queryResolver) AddressCommand(ctx context.Context, input v1beta1.Address) (string, error) {
-
-	return `apiVersion: enmasse.io/v1beta1
+	return  `oc apply -f - << EOF
+apiVersion: enmasse.io/v1beta1
 kind: Address
 metadata:
   namespace: `+ input.Namespace +`
@@ -197,5 +197,6 @@ spec:
   address: `+ input.Spec.Address +`
   addressSpace: `+ input.Spec.AddressSpace +`
   type: `+ input.Spec.Type +`
-  plan: `+ input.Spec.Plan, nil
+  plan: `+ input.Spec.Plan +`
+EOF`, nil
 }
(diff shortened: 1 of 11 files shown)


pkg/consolegraphql/resolvers/resolver_address.go
```diff
@@ -187,5 +187,15 @@ func (r *mutationResolver) PurgeAddress(ctx context.Context, input metav1.Object
 }
 
 func (r *queryResolver) AddressCommand(ctx context.Context, input v1beta1.Address) (string, error) {
-	return "oc apply blah", nil
+
+	return `apiVersion: enmasse.io/v1beta1
```
k-wall: I thought the intent was to provide a complete UNIX shell command, which could be copied and pasted straight to the terminal?  I was expecting a here document.

oc apply -f << EOF
...
EOF

pkg/consolegraphql/resolvers/resolver_address.go
```diff
@@ -187,5 +187,15 @@ func (r *mutationResolver) PurgeAddress(ctx context.Context, input metav1.Object
...
+kind: Address
```
k-wall: what is the reason to not use yaml serialisation of the object?

pkg/consolegraphql/resolvers/resolver_address.go
vbusch: @k-wall that was what I was expecting to do.  But then I checked the wireframes and saw.  https://marvelapp.com/i1igj54/screen/58093530 .  And ended up just providing the yaml.
I do agree that being able to cut and paste would be more useful.  I can quickly add the commands.

pkg/consolegraphql/resolvers/resolver_address.go
vbusch: @k-wall I will provide a sample.  It was giving me a number of empty fields, that made it look more complicated.  I'll see if there is a way to remove them.

vbusch: @k-wall  I haven't been able to find a way to get the yaml marshaller to remove the k8s.io.ObjectMeta objects empty field.  If I could tell it to look at the json's tags (instead of yaml's tags), it should work. 
```kind: AddressSpace
  apiversion: enmasse.io/v1beta1
  metadata:
    name: mynamespace
    generatename: ""
    namespace: myaddressspace
    selflink: ""
    uid: 7ba0337f-9037-42e6-b00a-2831cf51a77e
    resourceversion: ""
    generation: 0
    creationtimestamp: "0001-01-01T00:00:00Z"
    deletiontimestamp: null
    deletiongraceperiodseconds: null
    labels: {}
    annotations: {}
    ownerreferences: []
    initializers: null
    finalizers: []
    clustername: ""
    managedfields: []
  spec:
    plan: standard
    type: standard-unlimited
    authenticationservice:
      name: none-authentication```

Comment (k-wall) [reply pkg/consolegraphql/resolvers/resolver_address.go]: There might be a more elegant solution, but I think the following process would produec yaml with the field names following the JSON annotations:

1) Marshall Object to JSON (byte array)
2) Unmarshal the byte array to map[interface{}]interface{}
3) Marshall the map to yaml

Comment (vbusch) [reply pkg/consolegraphql/resolvers/resolver_address.go]: I have swtiched to use JSON -> MAP -> YAML marshalling.  Let me know, if you think it is better this way or before.

File: pkg/consolegraphql/resolvers/resolver_address_test.go [hunk 3]
```diff
@@ -336,6 +336,24 @@ func TestQueryAddressMetrics(t *testing.T) {
 	assert.Equal(t, float64(100), value, "Unexpected stored metric value")
 }
 
+func TestAddressCommand(t *testing.T) {
+	r,ctx := newTestAddressResolver(t)
+	namespace := "mynamespace"
+	addr := createAddress(namespace, "myaddrspace.myaddr")
```
k-wall: You don't need to add the object to cache.

PR #3797, opened 2020-01-29
Implement address purge

k-wall:
Agent now has separate channels, one for commands and one for the flow of address and connection updates.  Console-server authenticates to the agent separately when the user requests an operation such as address purge using the access token of the logged on user.

Internally, the AgentCollector got generalised into an AgentDelegate that now has a factory method CommandDelegate which produces a command delegate for a given access token.  It is this object that the rest of the resolvers use to cause the mutation operation to occur.


pkg/consolegraphql/agent/agent.go
```diff
@@ -0,0 +1,534 @@
...
+		AddressSpace:          aad.addressSpace,
+		InfraUuid:             aad.infraUuid,
+		Type:                  AgentEventTypeRestart,
+	}
+	err = aad.handler(restart)
+	if err != nil {
+		return err
+	}
+
+	housekeepInterval := 1 * time.Minute // TODO Make configurable
```
lulf: Just so I understand this correctly,this is the 'status collection interval' right?

pkg/consolegraphql/agent/agent.go
```diff
@@ -0,0 +1,534 @@
...
+}
+
+func (aad *amqpAgentDelegate) housekeepCommandDelegates() {
+	findExpiredCommandDelegates := func() []CommandDelegate {
+		aad.delegateMux.Lock()
+		defer aad.delegateMux.Unlock()
+
+		expired := make([]CommandDelegate, 0)
+
+		horizon := time.Now().Add(time.Minute * -5)
```
lulf: Is this how long the collector outstanding commants are kept?

pkg/consolegraphql/agent/agent.go
k-wall: I'll give it a better name.   The agent command delegates belong to a logged (human) user.  They are retained in a cache in a connected state just in case the user invokes another command.  The housekeeper interval refers to the scheduling of a go routine that closes command delegates that have been unused for more than 5 minutes.

lulf: Thanks for the clarification

pkg/consolegraphql/agent/agent.go
```diff
@@ -0,0 +1,534 @@
...
+	}
+}
+
+func buildTlsConfig(devMode bool) *tls.Config {
+	tlsConfig := &tls.Config{}
+	if devMode {
+		tlsConfig.InsecureSkipVerify = true
+	} else {
+		// TODO fix me - we need a CA that is the issuer of all the admin server certs.
+		tlsConfig.InsecureSkipVerify = true
```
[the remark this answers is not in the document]
k-wall: That's addressed in 0f056db.   Unfortunately there's no easy way to turn off host name validation whilst leaving the chain validation in place, hence the ugly boilerplate.

pkg/consolegraphql/agent/agent.go
k-wall: It is renamed `commandDelegateExpiryPeriod` and controllable via an environment variable.  When command delegates expire, their AMQP connections are closed.

pkg/consolegraphql/agent/agent.go
k-wall: Resolved above.

pkg/consolegraphql/agent/agent.go
```diff
@@ -134,6 +134,7 @@ func (aad *amqpAgentDelegate) doCollect() error {
 		amqp.ConnSASLXOAUTH2("unused", aad.bearerToken, amqpOverrideSaslFrameSize),
 		amqp.ConnServerHostname(aad.host),
 		amqp.ConnProperty("product", "console-server"),
+		amqp.ConnConnectTimeout(time.Second * 10),
```
lulf: Should this be tunable as well? I remember we had to do that for the standard-controller <-> router connection .

PR #3798, opened 2020-01-29
Add support for specifying annotations in pod template

lulf:


lulf:

api-model/src/main/java/io/enmasse/address/model/KubeUtil.java
```diff
@@ -175,6 +175,14 @@ public static void applyPodTemplate(PodTemplateSpec actual, PodTemplateSpec desi
             actual.getMetadata().setLabels(labels);
         }
 
+        if (desired.getMetadata() != null && desired.getMetadata().getAnnotations() != null) {
+            Map<String, String> annotations = new HashMap<>(desired.getMetadata().getAnnotations());
```
k-wall: Should the pod template be allowed to specify annotations that begin enmasse.io/?

lulf: Its using the same 'rules' as for labels. I think as long as we override any conflicting entries with our own own (see the lines below), then we should be fine.

PR #3708, opened 2020-01-21
Changed mock console server and client to use patch path 'spec/plan'

vbusch:
Signed-off-by: Vanessa <vbusch@redhat.com>

### Type of change

- Bugfix
- Enhancement / new feature
- Refactoring
- Documentation

### Checklist

- [ ] Update/write design documentation in `./documentation/design`
- [ ] Write tests and make sure they pass
- [ ] Update documentation
- [ ] Check RBAC rights for Kubernetes / OpenShift roles
- [ ] Try your changes from Pod inside your Kubernetes and OpenShift cluster, not just locally
- [ ] Reference relevant issue(s) and close them after merging
- [ ] Update CHANGELOG.md

PR #3811, opened 2020-01-31
Do Not Merge: Update graphql schema to match kubernetes objects to simplify patch.

vbusch:
Switched casing to lower case.  
ToDo:
- Order by and filter are still using golang capitalization


vbusch:
Included the changes in: https://github.com/EnMasseProject/enmasse/pull/3708

PR #3819, opened 2020-02-02
Fix #3763: Docs - include the topic field in the address example with…

### LINKED ISSUE #3763: Better documentation for creating addresses of the subscription type
**Description**

When creating a subscription the user needs to specify the topic's address in the spec.topic field, however I don't see this documented.     This ought to be called out in the main documentation and the swagger/CRD.

k-wall:
### Type of change

- Documentation

### Description

Minimal improvement to the documentation to cover the use of the topic: field in the address resource.

### Checklist

- [ ] Update documentation
- [ ] Update CHANGELOG.md


documentation/common/address-example1.yaml
```diff
@@ -6,3 +6,4 @@ spec:
     address: myqueue <2>
     type: queue <3>
     plan: standard-small-queue <4>
+    topic: topic-address <5>
```
lulf: I find this example a bit confusing mentioning only queue type and plan, and then making a reference to a topic. I feel this should be in a separate example.

k-wall: That's a good point, I think separate examples would be an improvement.

jenmalloy: @k-wall, @lulf, I will create a new issue to add this content in a new, separate example. Thanks!

k-wall:
@jenmalloy @lulf I pulled out a separate example for topic/subscription.  How does this look?

documentation/modules/ref-address-example-topic-and-sub.adoc
```diff
@@ -0,0 +1,20 @@
+// Module included in the following assemblies:
+//
+// assembly-managing-addresses.adoc
+
+[id='ref-address-example-topic-and-sub-{context}']
+== Topic and Subscription Address Example
```
jenmalloy:
```suggestion
== Topic and subscription address example
```

documentation/modules/ref-address-example-topic-and-sub.adoc
```diff
@@ -0,0 +1,20 @@
...
+
+When using topics and subscriptions, the subscription references the topic to which it belongs using the `topic:` field, as illustrated below.
```
jenmalloy:
```suggestion
When using topics and subscriptions, the subscription references the topic to which it belongs using the `topic:` field, as shown in the following example.
```

k-wall:
> LGTM, but does the tooling allow for these includes to be used @jenmalloy or do the examples have to be inline in the doc for it to work?

My intent was to follow the existing practice with regards to file inclusions etc so there should be no new demands on the tooling.  Did I make a mistake?

lulf:
@k-wall I can't remember fully, but iirc we started move includes inline for some reason it was easier to keep in sync for @jenmalloy . I'm fine with using includes if it works though.

jenmalloy:
@k-wall @lulf seems like the includes are ok, so reapproving. Thanks for the discussion!

PR #3838, opened 2020-02-08
Default address resource name from address

k-wall:
To simplify the UI, the server should default the address resource name from the address if it is not specified.

### DIFF
--- a/pkg/consolegraphql/resolvers/resolver_address.go
+++ b/pkg/consolegraphql/resolvers/resolver_address.go
@@ -27,8 +27,6 @@ var separators = []string{"_", ".", "-"}
 
 const maxKubeName = 253
 
-const addressSpaceNameAnnotationKey = "addressSpace"
-
 func (r *Resolver) Address_consoleapi_enmasse_io_v1beta1() Address_consoleapi_enmasse_io_v1beta1Resolver {
 	return &addressK8sResolver{r}
 }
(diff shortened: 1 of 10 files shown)


k-wall:
@lulf I noticed that previously address were chopped at 63 when forming the address resource name.  It was not clear to me why taking only the first 60 was important.   In my implementation I made sure that the name fits with the limit stated here: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/

const MAX_KUBE_NAME = 63/*max allowed*/ - 3/*needed for kube to add stateful set qualifier*/;

lib/utils.js:121

pkg/consolegraphql/resolvers/resolver_address.go
```diff
@@ -272,3 +297,47 @@ func tokenizeAddress(name string) ([]string, error) {
...
 	return addressToks, nil
 }
+
+func defaultResourceNameFromAddress(input *v1beta1.Address) error {
+	if input.Spec.Address == "" {
+		return fmt.Errorf("address is undefined, cannot default resource name from address '%s'",
+			input.Spec.Address)
+	}
+	if input.Annotations != nil {
+		if as, present := input.Annotations[addressSpaceNameAnnotationKey]; present {
```
lulf: Is this required to be set by the console and always guaranteed to be correct?

k-wall: As things stand now, the console ui will be required to set the annotation in order to let the backend know which address space the new address belongs.

On reflection, maybe it would be better to avoid exposing the implementational detail (the annotation is really a private concern of the controller) and change the GraphQL API (createAddress, addressCommand) to accept an addressSpaceName parameter.  The parameter would be required if you want the server to default the address resource name.

@lulf @vbusch  WDYT?

lulf: @k-wall Ok, although putting it in the request is nice, I'm fine with leaving it as annotation too, mainly because we would not have to change the API once we remove address spaces.

vbusch: Having createAddress & addressCommand take the addressSpaceName seems more straight forward.
Soon, with shared infra, the addressSpace will no longer be required, as we will only be using the namespace.

PR #3842, opened 2020-02-10
Rename PodDistruptionBudget -> PodDisruptionBudget

lulf:


address-space-controller/src/main/java/io/enmasse/controller/PodDisruptionBudgetController.java
```diff
@@ -159,4 +159,9 @@ private void reconcileBrokerPodDisruptionBudget(AddressSpace addressSpace, Stand
             deleteIfExists(name);
         }
     }
+
+    @Override
+    public String toString() {
+        return "PodDisruptionBudgetController";
```
dejanb: Maybe add a `namespace` value here for more details

address-space-controller/src/main/java/io/enmasse/controller/PodDisruptionBudgetController.java
```diff
@@ -17,13 +17,13 @@
...
-public class PodDistruptionBudgetController implements Controller {
-    private static final Logger log = LoggerFactory.getLogger(PodDistruptionBudgetController.class);
+public class PodDisruptionBudgetController implements Controller {
+    private static final Logger log = LoggerFactory.getLogger(PodDisruptionBudgetController.class);
 
     private final NamespacedKubernetesClient client;
     private final String namespace;
 
-    public PodDistruptionBudgetController(NamespacedKubernetesClient client, String namespace) {
+    public PodDisruptionBudgetController(NamespacedKubernetesClient client, String namespace) {
```
dejanb: It looks like namespace is tied to the client? If so, then `namespace` parameter seems unnecessary.

address-space-controller/src/main/java/io/enmasse/controller/PodDisruptionBudgetController.java
lulf: The namespace should be the same as where infra is deployed, so I'd like to follow the convention for the other controllers added to the chain, otherwise the log will look strange.

address-space-controller/src/main/java/io/enmasse/controller/PodDisruptionBudgetController.java
lulf: For some reason, kubernetes-client tends to complain about operations missing the namespace if it is not supplied. I could've also used client.getNamespace() I guess, but I'd rather change all our controllers in one go for that.

lulf:

PR #3847, opened 2020-02-10
Clean cache and perform update earlier

lulf:


lulf:

PR #3852, opened 2020-02-11
Pass -mod=vendor flag when generating graphql code

k-wall:
Hopefully this allows code generation to work successfully without having first installed gqlgen into the `$GOPATH`.


k-wall:
My original intention was to incorporate the running of console codegen steps into update/verify codegen.  Since @ctron nice refactoring, this turns out to be quite hard.  The real pain point is  gqlgen  - its tooling doesn't seem to like running outside of the project tree and appears to demand that all the types references are available.  My attempts to make it play nicely failed.
For the now I am content to leave run-console-codegen.sh separate.

PR #3856, opened 2020-02-11
Implement a finer grained check for the existince of APIs

ctron:
### Type of change

- Bugfix

### Checklist

- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/pkg/util/openshift.go
+++ b/pkg/util/openshift.go
@@ -1,68 +1,95 @@
 /*
- * Copyright 2019, EnMasse authors.
+ * Copyright 2019-2020, EnMasse authors.
  * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
  */
 package util
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"net/url"
 	"os"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	"strings"
 	"time"
 
 	"sigs.k8s.io/controller-runtime/pkg/client/config"
 
+	routeapiv1 "github.com/openshift/api/route/v1"
 	routev1 "github.com/openshift/client-go/route/clientset/versioned/typed/route/v1"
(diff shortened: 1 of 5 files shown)


ctron:

ctron:

k-wall:
Running the unit tests with `make test_go` panics with this change unless the environment variable NAMESPACE is not set.  I don't think the average developer would have this in their environment.

panic: infrastructure namespace is not set in 'NAMESPACE'

```
goroutine 1 [running]:
github.com/enmasseproject/enmasse/pkg/util.init.0()
	/home/keith/go/src/github.com/enmasseproject/enmasse/pkg/util/env.go:101 +0x111

```

ctron:
> @ctron Running the unit tests with `make test_go` panics with this change unless the environment variable NAMESPACE is not set. I don't think the average developer would have this in their environment.
> 
> panic: infrastructure namespace is not set in 'NAMESPACE'
> 
> ```
> goroutine 1 [running]:
> github.com/enmasseproject/enmasse/pkg/util.init.0()
> 	/home/keith/go/src/github.com/enmasseproject/enmasse/pkg/util/env.go:101 +0x111
> ```

Same for the CI *sigh* …

pkg/util/openshift.go
```diff
@@ -1,68 +1,90 @@
...
-	value, ok := os.LookupEnv("ENMASSE_OPENSHIFT")
+func detectApi(gvk schema.GroupVersionKind) bool {
+
+	name := fmt.Sprintf("ENMASSE_HAS_API_%s_%s_%s", strings.ToUpper(gvk.Group), strings.ToUpper(gvk.Version), strings.ToUpper(gvk.Kind))
+
+	value, ok := os.LookupEnv(name)
 	if ok {
-		log.Info("Set by env-var 'ENMASSE_OPENSHIFT': " + value)
+		log.Info("Set by env-var '" + name + "': " + value)
 		return strings.ToLower(value) == "true"
```
k-wall: Should we be using `strconv.ParseBool()` (and treating error as false)?  A user familiar to Go conventions could pass one of the other Go truthy values (1, T, 0, F)?

ctron:

pkg/util/openshift.go
ctron: I kept that line as it was, but I agree, that might be a nice change.

ctron: Just checked on the API, that would require us to handle `error` was well … which is different than the Java side

PR #3866, opened 2020-02-12
Wait for the expected number of connectors to be present

lulf:


kornys:
@enmasse-ci run tests testcase=bridging.** profile=isolated test_only ocp4

PR #3869, opened 2020-02-12
Example infinispan deployment remove cpu limits

famartinrh:
Currently our tests fail to deploy our infinispan example deployment in Openshift 4.4 because of the limits on cpu we have on that deployment.
Just removing the limits work, I also tried to reduce the limits to 0.25 and also work.


famartinrh:
@enmasse-ci run tests profile=iot OCP4

PR #3870, opened 2020-02-12
Add support for specifying maxConsumers of a subscription

### LINKED ISSUE #3868: Allow configuring maxConsumers for durable subscriptions
**Description**

At present the number of consumers for a durable subscription is hardcoded to 1. Making this configurable would allow for shared durable subs.

**Design**

Add a field .spec.subscription.maxConsumers to the Address type. The agent will read this field when configuring the sub.

**Tasklist**

Enumerate the sub-tasks required to complete the task:

* [ ] Add to model
* [ ] Configure in agent.

lulf:
### Type of change

- Enhancement / new feature

### Description

This allows configuring the max consumers setting of the subscription as
part of the address definition. At present, this setting cannot be
changed without recreating the subscription.

Fixes #3868

### Checklist

- [ ] Update documentation
- [x] Reference relevant issue(s) and close them after merging
- [x] Update CHANGELOG.md


lulf:
@enmasse-ci run tests testcase=TopicTest

lulf:
@enmasse-ci run tests testcase=standard.TopicTest

lulf:

lulf:

systemtests/src/test/java/io/enmasse/systemtest/shared/standard/TopicTest.java
```diff
@@ -388,6 +391,95 @@ void testDurableSubscriptionOnPooledTopic() throws Exception {
...
+        resourcesManager.setAddresses(topic, subscription);
+
+        // Start two consumers and check that the second consumer is disconnected
+        AmqpClient client = getAmqpClientFactory().createTopicClient();
+        Future<List<Message>> c1 = client.recvMessages(AddressUtils.getQualifiedSubscriptionAddress(subscription), 0);
+        Future<List<Message>> c2 = client.recvMessages(AddressUtils.getQualifiedSubscriptionAddress(subscription), 0);
+
+        // Second consumer should get an error
+        try {
+            c2.get(10, TimeUnit.SECONDS);
```
kornys: @lulf I think better is assertThrow instead of try/catch fail

systemtests/src/test/java/io/enmasse/systemtest/shared/standard/TopicTest.java
```diff
@@ -46,6 +46,9 @@
 import static io.enmasse.systemtest.TestTag.NON_PR;
 import static org.hamcrest.CoreMatchers.is;
 import static org.hamcrest.MatcherAssert.assertThat;
+import static org.junit.Assert.assertEquals;
```
kornys: alsoplease do not use asserts from junit4, use junit5 asserts from org.junit.jupiter

templates/crds/addresses.crd.yaml
```diff
@@ -85,6 +85,13 @@ spec:
             topic:
               type: string
               description: "Address of topic that this subscription refers to (only applicable to subscription types)."
+            subscription:
+              type: object
+              description: "Properties that can be set for subscription type addresses."
+              properties:
+                maxConsumers:
+                  type: integer
+                  description: "Maximum number of consumers attached to this subscription."
```
k-wall:
```suggestion
                  description: "Maximum number of concurrent consumers that can be attached to this subscription. If unspecified, 1 consumer is allowed per subscription"
```

k-wall:
@lulf couple of further thoughts:

* "At present, this setting cannot be changed." this needs to be reflected in the documentation.
* The actual maxConsumers setting is invisible to the end user.  I wonder if the address should have new subscription section in the status field that reveals its applied value.
* I notice that Artemis `org.apache.activemq.artemis.api.core.management.ActiveMQServerControl#updateQueue(String, String, Integer, Boolean, Boolean, String)` does support the modification of the attribute, so I think we could support updates to the field (not tested).
* Should we add the field to the create subscription address workflow in the console?

lulf:
@k-wall Agreed with your thoughts! 

> @lulf couple of further thoughts:
> 
>     * "At present, this setting cannot be changed." this needs to be reflected in the documentation.
> 
Agreed, I will add that to the subscription type section.

>     * The actual maxConsumers setting is invisible to the end user.  I wonder if the address should have new subscription section in the status field that reveals its applied value.
>

I agree, it makes sense if it can be changed dynamically. If its static, I think it's a matter of documenting the default.

>     * I notice that Artemis `org.apache.activemq.artemis.api.core.management.ActiveMQServerControl#updateQueue(String, String, Integer, Boolean, Boolean, String)` does support the modification of the attribute, so I think we could support updates to the field (not tested).
> 

I agree that would be nice. The major part of that work would be to refactor the agent code to support more fine-grained desired vs. actual state of queue settings. At present it only deals with the existence/absence of queues, so this would require significant more work compared to the currently limited approach.

>     * Should we add the field to the create subscription address workflow in the console and expose on the subscription address detail page.   Adding this should be straight forward.

Agreed, I can try to add that now that console server is merged.

lulf:
@k-wall addressed the docs and status section. I have raised https://github.com/EnMasseProject/enmasse/issues/3879 for the console improvement.

lulf:
@enmasse-ci run tests profile=standard.TopicTest

lulf:
@enmasse-ci run tests profile=standard.TopicTest

lulf:
@enmasse-ci run tests testcase=standard.TopicTest

templates/crds/addresses.crd.yaml
```diff
@@ -136,6 +143,13 @@ spec:
                       - Active
                       - Migrating
                       - Draining
+            subscription:
+              type: object
+              description: "Applied poroperties for subscription type addresses."
```
k-wall:
```suggestion
              description: "Applied properties for subscription type addresses."
```

lulf:
@enmasse-ci run tests testcase=standard.TopicTest

PR #3894, opened 2020-02-14
Add wakeup event to trigger resync on outdated resource version

lulf:
### Type of change

- Bugfix

### Description

If the resourceVersion of objects may change between list and watch, the
watch may be closed immediately after creation. To avoid waiting for the
next resync, trigger a new resync if the exception has the 410 Gone
status.

### Checklist

- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/k8s-api/src/main/java/io/enmasse/k8s/api/cache/EventCache.java
+++ b/k8s-api/src/main/java/io/enmasse/k8s/api/cache/EventCache.java
@@ -8,6 +8,7 @@ import static io.enmasse.k8s.api.cache.EventCache.EventType.Added;
 import static io.enmasse.k8s.api.cache.EventCache.EventType.Deleted;
 import static io.enmasse.k8s.api.cache.EventCache.EventType.Sync;
 import static io.enmasse.k8s.api.cache.EventCache.EventType.Updated;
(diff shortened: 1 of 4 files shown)


k8s-api/src/main/java/io/enmasse/k8s/api/cache/EventCache.java
```diff
@@ -53,7 +55,7 @@ public EventCache(FieldExtractor<T> fieldExtractor) {
     @Override
     public void pop(Processor<T> processor, long timeout, TimeUnit timeUnit) throws Exception {
         Event<T> firstEvent = queue.poll(timeout, timeUnit);
-        if (firstEvent == null) {
+        if (firstEvent == null || Wakeup.equals(firstEvent.eventType)) {
```
ctron: With enums you can do `==`

ctron: I also think that later on (line 72) you need to handle `Wakeup` as well.

lulf:

lulf:
Merging this to get some test runs on master during the weekend.

PR #3895, opened 2020-02-14
add tests verify broker remains operable after full queue

famartinrh:
### Type of change

- Test

### Description

Test described in ENTMQMAAS-1040

### Checklist

- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/systemtests/src/test/java/io/enmasse/systemtest/isolated/CommonTest.java
+++ b/systemtests/src/test/java/io/enmasse/systemtest/isolated/CommonTest.java
@@ -10,6 +10,20 @@ import io.enmasse.address.model.Address;
 import io.enmasse.address.model.AddressBuilder;
 import io.enmasse.address.model.AddressSpace;
 import io.enmasse.address.model.AddressSpaceBuilder;
+import io.enmasse.admin.model.v1.AddressPlan;
+import io.enmasse.admin.model.v1.AddressSpacePlan;
+import io.enmasse.admin.model.v1.AddressSpacePlanBuilder;
+import io.enmasse.admin.model.v1.BrokeredInfraConfig;
+import io.enmasse.admin.model.v1.BrokeredInfraConfigBuilder;
+import io.enmasse.admin.model.v1.BrokeredInfraConfigSpecAdminBuilder;
+import io.enmasse.admin.model.v1.BrokeredInfraConfigSpecBrokerBuilder;
+import io.enmasse.admin.model.v1.ResourceAllowance;
+import io.enmasse.admin.model.v1.ResourceRequest;
+import io.enmasse.admin.model.v1.StandardInfraConfig;
+import io.enmasse.admin.model.v1.StandardInfraConfigBuilder;
+import io.enmasse.admin.model.v1.StandardInfraConfigSpecAdminBuilder;
+import io.enmasse.admin.model.v1.StandardInfraConfigSpecBrokerBuilder;
+import io.enmasse.admin.model.v1.StandardInfraConfigSpecRouterBuilder;
 import io.enmasse.systemtest.UserCredentials;
 import io.enmasse.systemtest.amqp.AmqpClient;
 import io.enmasse.systemtest.bases.TestBase;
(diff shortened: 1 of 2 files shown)


famartinrh:
@enmasse-ci run tests testcase=CommonTest#testFullBrokerKeepsOperable*

famartinrh:
@enmasse-ci run tests testcase=CommonTest#testFullBrokerKeepsOperable* profile=isolated

famartinrh:
@enmasse-ci run tests testcase=CommonTest#testFullBrokerKeepsOperable* profile=isolated

famartinrh:
@enmasse-ci run tests testcase=CommonTest#testFullBrokerKeepsOperable* profile=isolated

famartinrh:
@enmasse-ci run tests testcase=CommonTest#testFullBrokerKeepsOperableStandard profile=isolated

famartinrh:
@enmasse-ci run tests testcase=CommonTest#testFullBrokerRemainsOperableStandard* profile=isolated

famartinrh:
@enmasse-ci run tests testcase=CommonTest#testFullBrokerRemainsOperableStandard* profile=isolated

kornys:
@enmasse-ci run tests testcase=CommonTest#testFullBrokerRemainsOperableStandard* profile=isolated

famartinrh:
@enmasse-ci run tests testcase=CommonTest#testFullBrokerRemainsOperableStandard* profile=isolated

famartinrh:
@enmasse-ci run tests testcase=CommonTest#testFullBrokerRemainsOperableStandard* profile=isolated

systemtests/src/test/java/io/enmasse/systemtest/isolated/CommonTest.java
```diff
@@ -425,6 +452,258 @@ void testMessagingDuringRestartComponents() throws Exception {
...
+                .withPlan(addressPlan.getMetadata().getName())
+                .endSpec()
+                .build();
+
+        if (createSecondAddressAfter) {
+            resourcesManager.setAddresses(address1);
+        } else {
+            resourcesManager.setAddresses(address1, address2);
+        }
+
```
k-wall: You've got an implicit assumption that the two queues are on the same broker (single). Given the state of the plans, this assumption should be safe, but I think the test should verify that starting condition is true.   That way, if we ever change the way the queue to broker allocation works, this will fail automatically.    In short, I am asking that you verify that the two queues have the same status.brokerStatuses.

systemtests/src/test/java/io/enmasse/systemtest/isolated/CommonTest.java
```diff
@@ -425,6 +452,258 @@ void testMessagingDuringRestartComponents() throws Exception {
         }
     }
 
+    @Test
+    @Tag(NON_PR)
+    @RepeatedTest(value = 3, name = RepeatedTest.LONG_DISPLAY_NAME)
```
k-wall: Why do we chose to repeat this test?  I notice this is not something we generally do.  What make this one special?

systemtests/src/test/java/io/enmasse/systemtest/isolated/CommonTest.java
famartinrh: great suggestion, I'll add it, thanks

systemtests/src/test/java/io/enmasse/systemtest/shared/standard/QueueTest.java
```diff
@@ -96,6 +97,45 @@ public static void runQueueTest(AmqpClient client, Address dest, int countMessag
...
     }
 
+    public static void runQueueTestRetrying(AmqpClient client, Address dest, int countMessages, Duration retryTimeout) throws InterruptedException, TimeoutException, ExecutionException, IOException {
+        List<String> msgs = TestUtils.generateMessages(countMessages);
+        long timeoutMs = countMessages * ClientUtils.ESTIMATE_MAX_MS_PER_MESSAGE;
+
+        TestUtils.waitUntilConditionOrFail(() -> {
+            log.info("Start sending with " + timeoutMs + " ms timeout");
+            Count<Message> predicate = new Count<>(msgs.size());
+            Future<Integer> numSent = client.sendMessages(dest.getSpec().getAddress(), msgs, predicate);
```
k-wall: If you succeed in `numSent > 0 ` and `numSent < msg.length` the send loop will retry.  This will mean that the number of messages of the queue ends up being greater than msg.length, which will make for an odd looking test failure.   You could reduce msg by actual at line 112 to avoid that.

systemtests/src/test/java/io/enmasse/systemtest/isolated/CommonTest.java
famartinrh: I added that annotation here to make sure this test wasn't flaky, the same applies for the test "...WithRetry".
Anyway we may conclude to disable or remove the test "testFullBrokerRemainsOperableStandard" because of being a known defect so I can remove the annotations for repeating test runs at any moment

k-wall: I'm not sure it is fair to say there is a Broker defect.  Presumably, internally, the Broker is having to (asynchronously) remove messages that were on the first queue, freeing up space, and then reset internal state to reflect the fact that it is no longer full.  I don't think it would be reasonable to expect this to be synchronous with the delete of the first queue (after all the queue could be enormous).   So for me, having to wait a period before the colocated queue accepts messages seems fine.

I think we should remove testFullBrokerRemainsOperableStandard and keep only the version that retries.

famartinrh: @k-wall wdyt about "testFullBrokerRemainsOperableStandardCreateSecondAddressAfter" does that test make sense? should it be removed as well?

famartinrh:
@enmasse-ci run tests testcase=CommonTest#testFullBrokerRemainsOperable* profile=isolated

famartinrh:
@enmasse-ci run tests testcase=CommonTest#testFullBrokerRemainsOperable* profile=isolated

systemtests/src/test/java/io/enmasse/systemtest/isolated/CommonTest.java
k-wall: I think I expect scenarios such as testFullBrokerRemainsOperableStandardCreateSecondAddressAfter to be part of the Broker test suite, rather than this one.   I'd be happy to see it removed.

famartinrh:
@enmasse-ci run tests testcase=CommonTest#testFullBrokerRemainsOperable* profile=isolated

famartinrh:
@enmasse-ci run tests testcase=CommonTest#testFullBrokerRemainsOperable* profile=isolated

famartinrh:
@enmasse-ci run tests testcase=CommonTest#testFullBrokerRemainsOperable* profile=isolated

famartinrh:
@enmasse-ci run tests testcase=CommonTest#testFullBrokerRemainsOperable* profile=isolated

PR #3896, opened 2020-02-14
Enable caching of maven and npm repos

lulf:


lulf:
Hmm, it seems caching is only limited to push and pull_request triggered workflows :( https://github.com/actions/cache/issues/63

The build is taking almost 2 hours now, so we should try to improve it somehow.

PR #3899, opened 2020-02-16
Add console-server readiness/liveness probe

k-wall:
### Type of change

- Enhancement / new feature

### Description

Adds a readiness/liveness probe to the console-server to ensure that it is minimally functional.

### Checklist

- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/pkg/controller/consoleservice/consoleservice_controller.go
+++ b/pkg/controller/consoleservice/consoleservice_controller.go
@@ -738,7 +738,27 @@ func applyDeployment(consoleservice *v1beta1.ConsoleService, deployment *appsv1.
 		}
 
 		// TODO use https
-		// TODO add health probe
+
+		probeHandler := corev1.Handler{
+			Exec: &corev1.ExecAction{
+				Command: []string{
+					"sh",
+					"-c",
+					"/probe.sh",
+				},
+			},
+		}
+
+		container.ReadinessProbe = &corev1.Probe{
+			InitialDelaySeconds: 30,
+			Handler:             probeHandler,
+		}
+
+		container.LivenessProbe = &corev1.Probe{
+			InitialDelaySeconds: 30,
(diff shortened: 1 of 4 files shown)


pkg/controller/consoleservice/consoleservice_controller.go
```diff
@@ -738,7 +738,27 @@ func applyDeployment(consoleservice *v1beta1.ConsoleService, deployment *appsv1.
...
+				Command: []string{
+					"sh",
+					"-c",
+					"/probe.sh",
+				},
+			},
+		}
+
+		container.ReadinessProbe = &corev1.Probe{
+			InitialDelaySeconds: 30,
```
lulf: Could it be worth making the init delay configurable in the console service
CR? Maybe not important in the short term.

k-wall: We need a consistent way of exposing this sort of thing for configuration.   In the infraconfigs we've added a `podTemplate` field, but this creates a ugly overlap between it and our CRD's own fields (e.g `podTemplate.container.resources` and `admin.resources`) so I am not sure this is the right pattern.    I wonder about reusing smaller kubernetes types instead.  For instance in this case we could add:

consoleServer.readinessProbe and consoleServer.livenessProbe which internally would reuse the Probe type and let the user override the timeouts.   WDYT?

lulf: Agreed, I think that strikes a good balance between ease of use and reusing kubernetes patterns at the appropriate level of abstraction.

k-wall: Ok, I'll take this approach. I'll merge this as it stands and incorporate the change into my next PR (which is pushes a number of variables to the CRD for configuration).

PR #3901, opened 2020-02-16
Add query time / query error prometheus metrics for the console server itself

k-wall:
### Type of change

- Enhancement / new feature

### Description

Expose query time / query error prometheus metrics for the console server itself.

### Checklist

- [ ] Update documentation
- [ ] Update CHANGELOG.md


controller-manager/src/main/resources/templates/ServiceMonitor-console.yaml
```diff
@@ -0,0 +1,23 @@
...
+  endpoints:
+  - port: metrics
+    path: /metrics
+    metricRelabelings:
+      - sourceLabels: [__name__]
+        targetLabel: __name__
+        replacement: "enmasse_${1}"
+  namespaceSelector:
+    matchLabels:
+      monitoring-key: middleware
```
k-wall: @lulf @robshelly what is the role of the monitoring-key?  Do I need to attach this to the console service as I see done for the address-space-controller.

lulf: I think it might be necessary for it to be picked up by the application monitoring operator, but I'm not sure if it is actually necessary or not anymore.

robshelly: It's still necessary as far as I know

controller-manager/src/main/resources/templates/ServiceMonitor-console.yaml
```diff
@@ -0,0 +1,23 @@
...
+  name: enmasse-console
+  namespace: {{ index .Params "Namespace" }}
+  labels:
+    monitoring-key: middleware
+    app: enmasse
+spec:
+  selector:
+    matchLabels:
+      app: enmasse
+      component: console
```
robshelly: This needs to be `consoleservice` as the service created for the console is labelled `component: consoleservice`

k-wall: That's done.

controller-manager/src/main/resources/templates/ServiceMonitor-console.yaml
k-wall: We agreed offline it wasn't necessary.

k-wall:
> Nice work! Metrics are working correctly but in order for them to be scraped just a few changes are needed:
> 
> The service for the console needs to also expose the metrics port. So need to add something like this [here](https://github.com/EnMasseProject/enmasse/blob/master/pkg/controller/consoleservice/consoleservice_controller.go#L431-L438):
> 
> ```
> {
> 	Port:       8080,
> 	Protocol:   corev1.ProtocolTCP,
> 	TargetPort: intstr.FromString("metrics"),
> 	Name:       "metrics",
> },
> ```
> 
> Also add the enmasse-console service monitor to the [template helper](https://github.com/EnMasseProject/enmasse/blob/master/pkg/util/templatehelper.go#L52-L63) in order for the operator to create it.

Done.

PR #3907, opened 2020-02-17
Add better keywords for trigger PR on crc

kornys:


kornys:
@enmasse-ci run tests profile=smoke test_only crc

PR #3919, opened 2020-02-18
fix leaking watches in systemtests

famartinrh:
This should fix the EOFException we are getting in shared tests jobs because in the case of creating and deleteing lots of address space systemetsts will leak lots of watches


lulf:

PR #3920, opened 2020-02-18
Add k8s loadBalancer instead of route when not on openshift

jbtrystram:
### Type of change

- Bugfix
- Enhancement / new feature

### Description

When not on openshift, a kubernetes loadbalancer should be used. This is a first draft to add this in iot compenents.

@lulf there are still a few types errors here and there but I wanted your feedback on the general approach. WDYT ? 
Aslo, according to [1] we need an Ingress controller. Any opinion on which one we should use ? 

[1]: https://kubernetes.io/docs/concepts/services-networking/ingress-controllers/

### Checklist

- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/pkg/controller/add_authenticationservice.go
+++ b/pkg/controller/add_authenticationservice.go
@@ -1,4 +1,4 @@
-/*
+/*https://github.com/EnMasseProject/enmasse/pull/3920
  * Copyright 2018, EnMasse authors.
  * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
  */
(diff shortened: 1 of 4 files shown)


obabec:
Before testing function of these change this branch should be rebased on newer master with changed deployment of infinispan on kubernetes.

lulf:
Can this be rebased and reviewed?

jbtrystram:
@lulf @obabec it looks like @ctron already added it https://github.com/EnMasseProject/enmasse/blob/0da533e3b0e0d83bfaf8dcb5793b83da44c59bff/pkg/controller/iotconfig/device_registry.go#L175

ctron:
IIRC it was @dejanb 

However, I think the misses the update of the status section

PR #3922, opened 2020-02-18
Allow to enable tracing for IoT

ctron:
I don't expect this to be merged for the next version. I also thinks it makes sense to have some common feature in EnMasse for this. But this helps with current IoT development.

This enables Jaeger tracing with the IoT parts of EnMassse in combination with the Jaeger Operator.


ctron:

pkg/controller/iotconfig/utils.go
```diff
@@ -62,12 +63,92 @@ func AppendEnvVarValue(container *corev1.Container, name string, value string) {
 	})
 }
 
+// block injection of sidecar variables, for containers not using jaeger
+func BlockTracingSidecarConfig(config *iotv1alpha1.IoTConfig, container *corev1.Container) {
+
+	if config.Spec.Tracing.Strategy.Sidecar != nil || config.Spec.Tracing.Strategy.DaemonSet != nil {
```
lulf: I'm a bit confused about the DaemonSet, wouldn't this potentially add a tracing pod to all nodes on the cluster?

ctron: Yes. The jaeger operator allows two modes: sidecar and daemonset.

I didn't want to impose either of them, so you need to align this configuration with your jaeger operator deployment.

ctron:
> Should we consider enabling of tracing to happen from a monitoring CR? Especially with the daemonset, I think that is something that definitely applies to components outside of IoT.

Yes, I absolutely agree. I tried to explain that in one of our calls, right now tracing helps us in developing IoT components. And adding a proper "tracing" CR is on our roadmap.

So my proposal would be, to get started with this PR, and removing the section from the IoTConfig later on, when we have that "tracing" or " monitoring" CRD. The other changes in this PR would be required anyway, as they enable tracing and jaeger in the different IoT components.

iot/pom.xml
```diff
@@ -19,6 +19,8 @@
     <!-- align with Hono compatible versions -->
     <spring.boot.version>2.1.5.RELEASE</spring.boot.version>
     <micrometer.version>1.1.4</micrometer.version>
+    <jaeger-client.version>0.34.0</jaeger-client.version>
+    <opentracing.version>0.31.0</opentracing.version>
```
lulf: Should we move this to the topevel pom? Jus thinking we want to standardize on the same version across enmasse and we could get into the situation where we duplicate this info because we forgot this being here.

ctron: I moved this to the top level.

PR #3928, opened 2020-02-19
Bug fix - prevent address space and address watchers spinning on resource update

k-wall:
### Type of change

- Bugfix

### Description

A defect in the update path meant that *every* update of an address (or address space) resource causes the watcher to restart.  This would cause all resources to be restarted and the watch
re-established.   This would result on unnecessary load on the kubernetes api-server and the console.

### Checklist

- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/pkg/consolegraphql/watchers/resource_watcher.go
+++ b/pkg/consolegraphql/watchers/resource_watcher.go
@@ -32,12 +32,7 @@ type ResourceWatcher interface {
 type WatcherOption func(watcher ResourceWatcher) error
 
 func atomicInc(i *int32) {
-	for {
-		val := atomic.LoadInt32(i)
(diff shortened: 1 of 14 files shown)


pkg/consolegraphql/watchers/resource_watcher.go
```diff
@@ -15,11 +15,61 @@
...
 }
 
 type WatcherOption func(watcher ResourceWatcher) error
+
+func atomicInc(i *int32) {
+	for {
+		val := atomic.LoadInt32(i)
+		if atomic.CompareAndSwapInt32(i, val, val+1) {
+			break
+		}
```
lulf: I think you can use this instead https://golang.org/pkg/sync/atomic/#AddInt32

PR #3934, opened 2020-02-19
Ensure infra uuid is not lost on replace

lulf:
### Type of change

- Bugfix

### Description

This adds an extra annotation with the namespace of the address space to the 'messaging' Service resource that is created for every address space. In the event that an address space is replaced and the annotations are lost, they can be recovered from the service resource in the reconciliation loop.

This fix is a lot simpler than my original which were to introduce status subresource for address and address space, as that would have required upgrading kubernetes-client again. I think we should keep AddressSpace and Address as they are for now.

### Checklist

- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/address-space-controller/src/main/java/io/enmasse/controller/common/KubernetesHelper.java
+++ b/address-space-controller/src/main/java/io/enmasse/controller/common/KubernetesHelper.java
@@ -5,13 +5,6 @@
 
 package io.enmasse.controller.common;
 
-import java.io.File;
-import java.io.IOException;
-import java.util.Map;
-import java.util.Optional;
-import java.util.Set;
-import java.util.stream.Collectors;
-
 import io.enmasse.address.model.AddressSpace;
 import io.enmasse.address.model.KubeUtil;
 import io.enmasse.admin.model.v1.InfraConfig;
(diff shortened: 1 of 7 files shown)


lulf:

lulf:

address-space-controller/src/main/java/io/enmasse/controller/common/KubernetesHelper.java
```diff
@@ -174,6 +175,35 @@ public AppliedConfig getAppliedConfig(AddressSpace addressSpace) throws IOExcept
...
 
+    /**
+     * Attempt to find the infra uuid of an address space by doing a reverse-lookup on the expected services.
+     */
+    @Override
+    public String getInfraUuid(AddressSpace addressSpace) {
+        if (addressSpace.getAnnotation(AnnotationKeys.INFRA_UUID) != null) {
+            return addressSpace.getAnnotation(AnnotationKeys.INFRA_UUID);
+        }
+        ServiceList services = client.services().inNamespace(namespace).list();
```
k-wall: should this restrict by label app=enmasse?

PR #3948, opened 2020-02-21
add a prometheus record to monitor the state of enmasse-operator

jbtrystram:
### Type of change

- Enhancement / new feature

### DIFF
--- a/controller-manager/src/main/resources/templates/PrometheusRules-enmasse.yaml
+++ b/controller-manager/src/main/resources/templates/PrometheusRules-enmasse.yaml
@@ -20,6 +20,9 @@ spec:
     - record: enmasse_component_health
       expr: up{job="address-space-controller"} or on(namespace) (1- absent(up{job="address-space-controller"}) )
 
+    - record: enmasse_component_health
+      expr: up{job="enmasse-operator-metrics"} or on(namespace) (1- absent(up{job="enmasse-operator-metrics"}) )
+
     - alert: ComponentHealth
       annotations:
         description: "{{ "{{" }} $labels.job {{ "}}" }} has been down for over 5 minutes"


controller-manager/src/main/resources/templates/PrometheusRules-enmasse.yaml
```diff
@@ -20,6 +20,9 @@ spec:
     - record: enmasse_component_health
       expr: up{job="address-space-controller"} or on(namespace) (1- absent(up{job="address-space-controller"}) )
 
+    - record: enmasse_component_health
+        expr: up{job="enmasse-operator"} or on(namespace) (1- absent(up{job="enmasse-operator"}) )
```
robshelly: The operator is actually exporting metrics with the label `job="enmasse-operator-metrics"`.

controller-manager/src/main/resources/templates/PrometheusRules-enmasse.yaml
```diff
@@ -20,6 +20,9 @@ spec:
     - record: enmasse_component_health
       expr: up{job="address-space-controller"} or on(namespace) (1- absent(up{job="address-space-controller"}) )
 
+    - record: enmasse_component_health
+        expr: up{job="job="enmasse-operator-metrics""} or on(namespace) (1- absent(up{job="job="enmasse-operator-metrics""}) )
```
robshelly: Test this out and discovered there's an extra tab at the start of the line which is causing the creation of the resource to fail

jbtrystram: @robshelly indeed. I updated

controller-manager/src/main/resources/templates/PrometheusRules-enmasse.yaml
```diff
@@ -20,6 +20,9 @@ spec:
     - record: enmasse_component_health
       expr: up{job="address-space-controller"} or on(namespace) (1- absent(up{job="address-space-controller"}) )
 
+    - record: enmasse_component_health
+      expr: up{job="job="enmasse-operator-metrics""} or on(namespace) (1- absent(up{job="job="enmasse-operator-metrics""}) )
```
robshelly: My bad! I thought I was the indentation causing the error but I just noticed there's another typo. `job="` is in the labels twice

PR #3951, opened 2020-02-21
Address unsafe publication to volatiles used to form the schema

k-wall:
### Type of change

- Bugfix

### Description

The code was reading from currentAddressSpacePlans etc whilst holding the mutex, but the write
was made without.  Other threads (the other watching threads) were seeing stale values and therefore writing incomplete addressspacechema records.

### Checklist

- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@
 * #3873: Require TLSv1.2 or newer for external TLS endpoints
 * #3867: Add addition checks for router mesh status
 #      : Address CRD now validates the spec.address field comprises permitted characters.
+# #3949: Address unsafe publication to volatiles used to form the schema
 
 ## 0.30.2
 *  #2714: Allow setting security context of pods using persistent volumes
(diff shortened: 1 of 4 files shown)


k-wall:
@enmasse-ci run tests profile=acceptance

lulf:
@k-wall , I don't understand how #3951 fixes the issue of you saw of schema resource version increasing when the system is in the steady state. First, wrapping the assignment in a synchronized block doesn't seem to change anything, since the synchronized block is not also wrapping updateSchema.

And even without that fix, it would eventually be in a state where references between plans and infraconfigs would be valid, and once reaching that state I don't see how resourceVersion would change for the schema any less/more frequently than before this fix.

k-wall:
@lulf  the code needs to take the mutex for both reading and writing the shared variable.    previously the code was just taking the mutex on the read.  I was seeing cases where elements of the schema were empty for some invocations of updateSchema().  My conclusion was stale copies of shared variable.  Taking the mutex made this issue disappear.

PR #3958, opened 2020-02-21
Document remove console link step

k-wall:
### Type of change

- Bugfix
- Documentation

### Description

The documentation needs to call out a manual step to remove the consolelink.  This is OpenShift 4 only step.

### Checklist

- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/ansible/playbooks/openshift/uninstall.yml
+++ b/ansible/playbooks/openshift/uninstall.yml
@@ -13,8 +13,13 @@
       shell: oc delete clusterroles -l app=enmasse
     - name: Delete API services
       shell: oc delete apiservices -l app=enmasse
+    - name: Check for console API
+      shell: oc api-resources --namespaced=false --api-group=console.openshift.io --no-headers
+      register: console_api
+      ignore_errors: True
     - name: Delete consolelinks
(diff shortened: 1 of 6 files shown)


k-wall:
Tested manually on oc cluster (3.11) and CRC (4.2.14).

ansible/playbooks/openshift/uninstall.yml
```diff
@@ -13,6 +13,13 @@
       shell: oc delete clusterroles -l app=enmasse
     - name: Delete API services
       shell: oc delete apiservices -l app=enmasse
+    - name: Check for console API
+      shell: oc api-resources --namespaced=false --api-group=console.openshift.io --no-headers
+      register: console_api
+      ignore_errors: True
+    - name: Delete consolelinks
+      shell: oc delete consolelinks -l app=enmasse
+      when: console_api.stdout.find('ConsoleLink') != -1
```
lulf: Would it not be sufficient with only the 'Delete consolelinks' step?

k-wall: My intent was to separate the failure that would result on 3.11 as the resource is unknown from a failure that could occur on OCP 4 if something really went wrong.   Maybe I'm over complicating things?

lulf: Ah, understood. I'm fine leaving this in.

PR #3971, opened 2020-02-24
Require python3 to run script to avoid accidentally using python2

lulf:


lulf:

PR #3973, opened 2020-02-25
Fill kubeDomain based on api url

kornys:
### Type of change

- Enhancement

### Description

Fill kube domain based on api url for local development

### DIFF
--- a/systemtests/src/main/java/io/enmasse/systemtest/Environment.java
+++ b/systemtests/src/main/java/io/enmasse/systemtest/Environment.java
@@ -99,7 +99,7 @@ public class Environment {
             try {
                 this.kubernetesDomain = new URL(url).getHost().replace("api", "apps");
             } catch (Exception ex) {
-                ex.printStackTrace();
+                throw new RuntimeException(ex);
             }
         }
         log.info(debugFormat, TEST_LOG_DIR_ENV, testLogDir);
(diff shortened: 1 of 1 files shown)


kornys:
@enmasse-ci run tests profile=shared testcase=isolated.web.FirefoxConsoleTest#testLoginLogout crc test_only

systemtests/src/main/java/io/enmasse/systemtest/Environment.java
```diff
@@ -94,6 +95,12 @@ private Environment() {
         if (installType == EnmasseInstallType.OLM) {
             log.info(debugFormat, OLM_INSTALL_TYPE, olmInstallType.name());
         }
+        if (url.startsWith("https://api")) {
+            try {
+                this.kubernetesDomain = new URL(url).getHost().replace("api", "apps");
+            } catch (Exception ignored) {
```
lulf: Probably worth printing this exception and at least wrap it in a runtime exeception as I think this would be fatal?

kornys: ok let me fix that

kornys:
@enmasse-ci run tests profile=isolated testcase=isolated.web.FirefoxConsoleTest#testLoginLogout crc test_only

systemtests/src/main/java/io/enmasse/systemtest/Environment.java
kornys: I added printing of stacktrace, it is not easy to throw exception in Envritonment

lulf: use throw new RuntimeException(ex)

PR #3982, opened 2020-02-25
Wait for pods to be removed

lulf:
### Type of change

- Bugfix

### Description
Wait for deployment pods to be deleted
    
This prevents the standard-controller from re-creating resources that  gets deleted after the deployments are deleted.

### Checklist

- [ ] Update documentation
- [x] Update CHANGELOG.md

### DIFF
--- a/address-space-controller/src/main/java/io/enmasse/controller/common/KubernetesHelper.java
+++ b/address-space-controller/src/main/java/io/enmasse/controller/common/KubernetesHelper.java
@@ -140,7 +140,7 @@ public class KubernetesHelper implements Kubernetes {
     public void deleteResources(String infraUuid) throws InterruptedException {
         // Delete and await deployments to be deleted so that the per-address space controllers are not re-creating any resources
         for (Deployment deployment : client.apps().deployments().withLabel(LabelKeys.INFRA_UUID, infraUuid).list().getItems()) {
-            if (!client.apps().deployments().delete(deployment)) {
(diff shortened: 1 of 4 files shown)


lulf:

address-space-controller/src/main/java/io/enmasse/controller/common/KubernetesHelper.java
```diff
@@ -122,12 +122,33 @@ public static boolean isStatefulSet(HasMetadata res) {
...
+            throw new RuntimeException(message);
+        }
+    }
+
     @Override
-    public void deleteResources(String infraUuid) {
+    public void deleteResources(String infraUuid) throws InterruptedException {
+        // Delete and await deployments to be deleted so that the per-address space controllers are not re-creating any resources
+        for (Deployment deployment : client.apps().deployments().withLabel(LabelKeys.INFRA_UUID, infraUuid).list().getItems()) {
+            if (!client.apps().deployments().delete(deployment)) {
```
k-wall: I think it might be safer to maintain `withPropagationPolicy("Background")` in the delete call.  Some parts of fabric8 switched to orphan dependants true (which would leave behind pods).    I find it difficult to remember which, so I think being explicit is the way to safe.  https://github.com/fabric8io/kubernetes-client/issues/1775

lulf: @k-wall Fixed (though a bit clunky due to the api)

lulf:

PR #3983, opened 2020-02-25
quiesce the console-server if there are no logged on users

k-wall:
### Type of change

- Enhancement / new feature

### Description

If the console-server has no logged on user, it should be a quiesced state and not be pulling data from kubernetes or the agents.

### Checklist

- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/cmd/console-server/main.go
+++ b/cmd/console-server/main.go
@@ -6,30 +6,22 @@
 package main
 
 import (
-	"bytes"
 	"context"
-	"crypto/sha256"
 	"flag"
 	"github.com/99designs/gqlgen/graphql"
 	"github.com/99designs/gqlgen/handler"
-	"github.com/alexedwards/scs/v2"
 	adminv1beta2 "github.com/enmasseproject/enmasse/pkg/client/clientset/versioned/typed/admin/v1beta2"
 	enmassev1beta1 "github.com/enmasseproject/enmasse/pkg/client/clientset/versioned/typed/enmasse/v1beta1"
-	"github.com/enmasseproject/enmasse/pkg/consolegraphql/accesscontroller"
 	"github.com/enmasseproject/enmasse/pkg/consolegraphql/cache"
 	"github.com/enmasseproject/enmasse/pkg/consolegraphql/metric"
 	"github.com/enmasseproject/enmasse/pkg/consolegraphql/resolvers"
 	"github.com/enmasseproject/enmasse/pkg/consolegraphql/server"
 	"github.com/enmasseproject/enmasse/pkg/consolegraphql/watchers"
 	"github.com/enmasseproject/enmasse/pkg/util"
-	user "github.com/openshift/client-go/user/clientset/versioned/typed/user/v1"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"github.com/vektah/gqlparser/gqlerror"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/tools/clientcmd"
-	"k8s.io/client-go/tools/clientcmd/api"
 	"log"
 	"net/http"
 	"time"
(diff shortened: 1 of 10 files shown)


k-wall:
I introduced a watcher manager that has responsibility for controlling all of the watchers as a group.   The watcher manager is plugged into session's store and responds to sessionPresent and noSessionsPresent signals.   These signals cause the watcher manager to begin/end watching as appropriate.

Unfortunately there was no extension point in `scs.SessionManager` that allows you to determine how many sessions exist.   My solution to this was to exploit its store plugability.  I've cloned their original memstore implementation and add a thread-safe listener.  It is this that is wired up back to the manager.   (memstore.go credits the original author and maintains the original MIT license).

When the first session connects, the manager will begin the watching.  It is therefore likely that the console UI will first see an empty result set, however, by the time it polls again, it is likely that the cache will be populated.

k-wall:

pkg/consolegraphql/server/session.go
```diff
@@ -0,0 +1,57 @@
...
+type sessionCountingListener struct {
+	sessionCount      int32
+	sessionPresent    func()
+	noSessionsPresent func()
+	metric            prometheus.Gauge
+}
+
+func (l *sessionCountingListener) Added(string) {
+	curr := atomic.AddInt32(&l.sessionCount, 1)
+	if curr == 1 {
```
lulf: Is the assumption that Added is called only once for a session? If not, why is sessionPresent() only called if == 1 and not >= 1? Maybe add some comment to clarify how this works.

pkg/consolegraphql/server/session/memstore.go
```diff
@@ -0,0 +1,176 @@
...
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+/* Copy of https://github.com/alexedwards/scs/tree/master/memstore with the additional of Listener */
```
lulf: What is the reason to copy this into the tree instead of adding a dependency on it?

k-wall: I mentioned that in my comment [above](https://github.com/EnMasseProject/enmasse/pull/3983#issuecomment-590954282).  The provided memstore is not listenable, and there was no reasonable way to add it without cloning the code.  I did consider adaption, but that would not work for sessions that are expired by the store itself.

pkg/consolegraphql/server/session.go
k-wall: `Added` is called each time a new session is started.  `Removed` is called each time a session is removed (or is expired).   The watcher manager needs to be notified when the session store transitions from having no sessions to having the first, and the reverse, from having one to having none so that it can begin and end the watchers.    At the moment the watcher manager is not idempotent, so it is important that the begin and end messages are received exactly once.  To achieve this I used to atomic int in the sessions listeners and trigger when the atomic ints value becomes 0 or 1 accordingly.

lulf: Thanks for the explanation! Maybe put it into a comment?

k-wall: May be making watcher matcher idempotent would make the code clearer and address your comment?

k-wall:

k-wall:

cmd/console-server/main.go
```diff
@@ -340,7 +131,11 @@ http://localhost:` + port + `/graphql
 		Name: "console_query_error_total",
 		Help: "Number of queries that have ended in error",
 	}, []string{"operationName"})
-	prometheus.MustRegister(queryTimeMetric, queryErrorCountMetric)
+	sessionCountMetric := prometheus.NewGauge(prometheus.GaugeOpts{
+		Name: "console_active_session_count",
```
lulf: Should probably be console_active_session_total   (ref https://prometheus.io/docs/practices/naming/)

k-wall:
```suggestion
		Name: "console_active_session_total",
```

PR #3993, opened 2020-02-26
Increase selenium pod memory

kornys:


kornys:
@enmasse-ci run tests profile=systemtests testcase=web.** crc test_only

PR #3994, opened 2020-02-26
add padding in bottom of table to avoid scrolling for kebab

drcoolsanjeev:
### Type of change

- Bugfix
### Description
- FIX - list page row kebab menu not fully visible


k-wall:
Tested manually in Firefox and Chrome on the Mac. The kebab menus on the rows are much more usable.  Thanks @drcoolsanjeev

PR #3997, opened 2020-02-26
Handle kubernetes exceptions gracefully

lulf:
Adds a pause if the resync fails with a kubernetes client exception. This will prevent a lot of logging during shutdown where CRDs are deleted before the controllers.


lulf:

lulf:

PR #4011, opened 2020-02-28
fix-issue delete space modal doesn't pop down

drcoolsanjeev:
### Type of change

- Bugfix

### Description
- FIx - delete address space space dialogue doesn't pop down


k-wall:
LGTM

PR #4015, opened 2020-02-28
Change network error message

anukritijha:
### Type of change

- Bugfix

### Description

This PR changes the alert message in case of network errors and also changes the 'Take me Home' label to 'Sign in'.


k-wall:
I think it would be better to special case 403 errors and display an error message like "Your authenticated session has expired. Please login again".     For other server errors, the behaviour you had before was fine.  Some server errors (such as a server being restarted) will be transient and the user won't need to login again.

anukritijha:
@k-wall Raised a new PR, which addresses your comments, hence closing this one.
https://github.com/EnMasseProject/enmasse/pull/4028

PR #4017, opened 2020-02-28
add top 10 records in dropdown if more than 100 records are present

drcoolsanjeev:
### Type of change

- Bugfix
- Enhancement / new feature
### Description
- fix - Typeahead filter - gives misleading information
- show the first 10 results in dropdown with an indication that there are more available


kornys:
@enmasse-ci run tests testcase=FirefoxConsoleTest

console/console-init/ui/src/pages/AddressSpaceDetail/AddressList/AddressListFilter.tsx
```diff
@@ -156,11 +167,21 @@ export const AddressListFilter: React.FunctionComponent<IAddressListFilterProps>
...
     ) {
       //To display dropdown if fetched records are less than 100 in count.
-      if (response.data.addresses.total > 100) {
-        setNameOptions([]);
+      if (
+        response.data.addresses.total >
+        MAX_ITEM_TO_DISPLAY_IN_TYPEAHEAD_DROPDOWN
+      ) {
+        let list = [];
+        for (let i = 0; i < 10; i++) {
```
anukritijha: I think it's better to store 10 in a const variable in a global file and import it from there, as it's being used in multiple files.

anukritijha: Also, this logic here is repetitive and we can have a common function which just accepts a list of records and push the 10 records in an array

drcoolsanjeev:
@enmasse-ci run tests testcase=FirefoxConsoleTest

drcoolsanjeev:
@enmasse-ci run tests profile=shared testcase=web.**

PR #4026, opened 2020-03-02
systemtests disable dns caching

famartinrh:


famartinrh:
@enmasse-ci run tests profile=acceptance ocp4 test_only

famartinrh:
@enmasse-ci run tests profile=acceptance ocp4 test_only

famartinrh:
@enmasse-ci run tests profile=acceptance ocp4 test_only

systemtests/src/main/java/io/enmasse/systemtest/listener/JunitExecutionListener.java
```diff
@@ -24,6 +24,7 @@
 
     @Override
     public void testPlanExecutionStarted(TestPlan testPlan) {
+        java.security.Security.setProperty("networkaddress.cache.ttl" , "0");
```
k-wall: @famartinrh I'm not certain setting this here will be effective?  (Did you somehow test to make sure here is good?)   I think need to make sure it is set before any networking has occurred.  It might be safer to the system property sun.net.inetaddr.ttl (perhaps in Maven)

https://stackoverflow.com/questions/1256556/how-to-make-java-honor-the-dns-caching-timeout

PR #4036, opened 2020-03-04
Fix some issues in the system tests

ctron:
* If the route is not found, throw a more helpful exception.
* Combine the two setup methods, as they do rely on an order
  but @BeforEach does not honor any order
* Fix a race condition when deleting an IoTProject. It relies
  on an initial set of data, which might already be gone, as the
  deletion is on progress.

### DIFF
--- a/systemtests/src/main/java/io/enmasse/systemtest/utils/IoTUtils.java
+++ b/systemtests/src/main/java/io/enmasse/systemtest/utils/IoTUtils.java
@@ -177,7 +177,15 @@ public class IoTUtils {
         assertThat(project.getStatus().getDownstreamEndpoint().getPort(), not(is(0)));
     }
 
-    private static void waitForIoTProjectDeleted(Kubernetes kubernetes, IoTProject project) throws Exception {
+    public static boolean isIoTInstalled(Kubernetes kubernetes) {
+        return kubernetes.getCRD(IoTCrd.project().getMetadata().getName()) != null;
+    }
+
+    public static void deleteIoTProjectAndWait(Kubernetes kubernetes, IoTProject project) throws Exception {
+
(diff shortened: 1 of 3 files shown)


ctron:
@enmasse-ci run tests profile=iot testcase=io.enmasse.systemtest.iot.isolated.IoTProjectManagedTest.testCreate

kornys:
@enmasse-ci run tests profile=iot testcase=io.enmasse.systemtest.iot.isolated.IoTProjectManagedTest#testCreate

systemtests/src/main/java/io/enmasse/systemtest/utils/IoTUtils.java
```diff
@@ -177,7 +177,15 @@ public static void waitForIoTProjectReady(Kubernetes kubernetes, IoTProject proj
         assertThat(project.getStatus().getDownstreamEndpoint().getPort(), not(is(0)));
     }
 
-    private static void waitForIoTProjectDeleted(Kubernetes kubernetes, IoTProject project) throws Exception {
+    public static boolean isIoTInstalled(Kubernetes kubernetes) {
+        return kubernetes.getCRD(IoTCrd.project().getMetadata().getName()) != null;
```
lulf: This will always be true, the CRD is always created when installing.

ctron: Indeed. That came back by accident. It is removed again.

ctron:
@enmasse-ci run tests

ctron:
@enmasse-ci run tests

ctron:
@enmasse-ci run tests profile=shared testcase=io.enmasse.systemtest.shared.standard.web.FirefoxConsoleTest#testAddressStatus

PR #4046, opened 2020-03-05
add netork-only as fetch policy for all queries

drcoolsanjeev:
### Type of change
- Bugfix

### Description
- add fetchPoilcy as network-only for all graphql query to avoid caching


k-wall:
Corrected milestone from 0.31.0 -> 0.31.1

PR #4048, opened 2020-03-05
Set subscription name based on application bundle prefix

lulf:


kornys:
@enmasse-ci run tests profile=olm olm

PR #4053, opened 2020-03-06
Scale tests init

kornys:
### Type of change

- Enhancement

### Description

Scale tests PR

### Checklist

- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/systemtests/src/test/java/io/enmasse/systemtest/scale/ScaleTest.java
+++ b/systemtests/src/test/java/io/enmasse/systemtest/scale/ScaleTest.java
@@ -1,0 +1,195 @@
+/*
+ * Copyright 2018-2020, EnMasse authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+package io.enmasse.systemtest.scale;
+
+import io.enmasse.address.model.Address;
+import io.enmasse.address.model.AddressBuilder;
+import io.enmasse.address.model.AddressSpace;
+import io.enmasse.address.model.AddressSpaceBuilder;
+import io.enmasse.admin.model.v1.AddressPlan;
+import io.enmasse.admin.model.v1.AddressSpacePlan;
+import io.enmasse.admin.model.v1.AuthenticationService;
+import io.enmasse.admin.model.v1.ResourceAllowance;
+import io.enmasse.admin.model.v1.ResourceRequest;
+import io.enmasse.admin.model.v1.StandardInfraConfig;
+import io.enmasse.admin.model.v1.StandardInfraConfigBuilder;
+import io.enmasse.admin.model.v1.StandardInfraConfigSpecAdminBuilder;
+import io.enmasse.admin.model.v1.StandardInfraConfigSpecBrokerBuilder;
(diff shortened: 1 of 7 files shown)


kornys:
@enmasse-ci run tests profile=scale testcase=ScaleTest#testNumberOfSupportedAddresses test_only

systemtests/src/test/java/io/enmasse/systemtest/scale/ScaleTest.java
```diff
@@ -0,0 +1,224 @@
...
+    void tearDown(ExtensionContext extensionContext) throws Exception {
+        GlobalLogCollector.saveInfraState(TestUtils.getScaleTestLogsPath(extensionContext));
+        kubernetes.deleteNamespace(namespace);
+    }
+
+    @Test
+    void testNumberOfSupportedAddresses() throws Exception {
+        int operableAddresses = 0;
+        int iterator = 0;
+        int failureThreshold = 12000;
```
lulf: This should probably be higher based on your initial tests I think

systemtests/src/test/java/io/enmasse/systemtest/scale/ScaleTest.java
```diff
@@ -0,0 +1,224 @@
...
+
+        while (true) {
+            log.info("Total operable addresses: {}", operableAddresses);
```
lulf: This seems like a very verbose output that is not that useful unless test is failing.

kornys:
> To me, it feels like these scale tests are a bit too specific to one part of the system. I.e. what use is it to test that we can create 12000 addresses without actually using them? If we write the tests with this level of detail, how many test cases will we get, and how long will running the testsuite take?
> 
> The way I would have expected to write this test was to:
> 
>     1. Deploy a system using N addresses where M% are queues
> 
>     2. Attach C clients with L links each spread across the addresses
> 
>     3. Check if system behaves as expected: a) all addresses are ready within a reasonable time. b) no clients have errors attaching and/or sending a single message every minute over some time period. c) no crashes amongst enmasse pods.
> 
> 
> One can run this test varying those parameters, using some expected limit that we want to be able to support. But then we can also potentially reuse that test to check what happens if we push further.
> 
> Anyway, this is how I think about this, I'd love to hear different views.

I like your idea, but in case that system can't handle for example 12k addresses test failed and we don't have other results, so from my POV we should test how many addresses become active with plans we provided, next test should deploy some median of addresses and start operations like connections and messaging, next test should deploy addresses and do some kind of chaos monkey etc...

kornys:
@lulf or as Ondrej mentioned, we can setup scale shared env for scale tests... deploy baseline 12k addresses and then start 1st test with max count addresses, then remove difference to 12k and start 2nd test which start messaging testing with 12k addresses etc...

lulf:
So to summarize discussion in chat, the following tests will be created:

1) Address and connection scale tests - create 4 anycast + 1 queue, check create and ready time. Then create 1 sender/receiver pair connecting to the queue. Check that system is stable. Rinse & Repeat
2) Messaging performance tests - create addresses first, then create messaging clients and increase load from messaging clients
3) HA tests - create addresses first, then create messaging clients and api clients creating/deleting addresses and measure downtime.

PR #4056, opened 2020-03-06
Use user.openshift.io to make the determination about openshift rather than route.openshift.io

k-wall:
### Type of change

- Bugfix

### Description

For 0.31, we changed the way openshift was detected to use make an API refactoring the `detectOpenshift` function in the process to use `Route`.  Unfortunately, this brought with it an implict requirement that the service account running the code had access to that API.  For the console-server process, this wasn't the case.

* Switched `detectOpenshift` to use `user.openshift.io`.
* also propagared ENMASSE_OPENSHIFT from the enmasse-operator to the console-server to allow the check to the short circuited in environments where OpenShift is a certainty.

### Checklist

- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/pkg/controller/consoleservice/consoleservice_controller.go
+++ b/pkg/controller/consoleservice/consoleservice_controller.go
@@ -12,6 +12,7 @@ import (
 	"fmt"
 	"net"
 	"net/url"
+	"os"
 	"reflect"
 	"strconv"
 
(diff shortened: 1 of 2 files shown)


k-wall:
@enmasse-ci run tests testcase=*ConsoleTest test_only

PR #4066, opened 2020-03-09
Fix - console does not show custom address plan properly

drcoolsanjeev:
### Type of change
- Bugfix
- Enhancement / new feature
- Refactoring
- Documentation

### Description
- fix issue - console does not show custom address plan properly.
- show metadata.name as label for plan if spec.displayName is not present in response of addressPlan

### Checklist

- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/console/console-init/ui/src/pages/AddressDetail/AddressDetailPage.tsx
+++ b/console/console-init/ui/src/pages/AddressDetail/AddressDetailPage.tsx
@@ -211,7 +211,10 @@ export default function AddressDetailPage() {
         <AddressDetailHeader
           type={addressDetail.spec.plan.spec.addressType}
           name={addressDetail.spec.address}
-          plan={addressDetail.spec.plan.spec.displayName}
+          plan={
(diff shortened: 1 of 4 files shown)


k-wall:
Surely this same issue applies to address space plans too? They too have a displayName which is also could be empty.

kornys:
@drcoolsanjeev test passed, plan is shown properly

drcoolsanjeev:
@lulf , for short description (optional) , what should we replace with? should I replace shortDescripiton with shortDescription || longDescription || "<empty string>"?

lulf:
@drcoolsanjeev I'm a bit divided if it should be `shortDescription || longDescription || ""` or just `shortDescription || ""`. We don't really have any guidance on what longDescription should be, but I think your suggestion is probably fine for now.

lulf:
@drcoolsanjeev Will you make the change to use displayName for address space plans in a separate PR?

anukritijha:
@lulf Sanjeev has created a separate issue for the same and raised a PR.

PR #4070, opened 2020-03-09
Ensure finalizers are persisted immediately

lulf:
### Type of change

- Bugfix

### Description

This change ensures that finalizers are persisted immediately and that
we don't orphan resources created if address space was deleted before
finalizers are persisted.

### Checklist

- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/address-space-controller/src/main/java/io/enmasse/controller/ControllerChain.java
+++ b/address-space-controller/src/main/java/io/enmasse/controller/ControllerChain.java
@@ -97,12 +97,19 @@ public class ControllerChain implements Watcher<AddressSpace> {
                 log.debug("Controller chain input: {}", original);
 
                 log.info("Checking address space {}", addressSpaceName);
-                addressSpace.getStatus().setReady(true);
-                addressSpace.getStatus().clearMessages();
                 for (Controller controller : chain) {
                     log.info("Controller {}", controller);
                     log.debug("Address space input: {}", addressSpace);
-                    addressSpace = controller.reconcileAnyState(addressSpace);
+                    Controller.ReconcileResult result = controller.reconcileAnyState(addressSpace);
+                    addressSpace = result.getAddressSpace();
+
+                    // If instructed to requeue, break loop and let the comparison of original vs current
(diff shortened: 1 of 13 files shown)


lulf:

address-space-controller/src/main/java/io/enmasse/controller/ControllerChain.java
```diff
@@ -102,7 +102,12 @@ public void onUpdate(List<AddressSpace> resources) throws Exception {
                 for (Controller controller : chain) {
                     log.info("Controller {}", controller);
                     log.debug("Address space input: {}", addressSpace);
-                    addressSpace = controller.reconcileAnyState(addressSpace);
+                    Controller.ReconcileResult result = controller.reconcileAnyState(addressSpace);
+                    addressSpace = result.getAddressSpace();
+                    // If instructed to requeue, break loop and replace if modified
+                    if (result.isPersistAndRequeue()) {
+                        break;
```
ctron: Shouldn't there be some code which actually re-triggers the processing? I guess it would still work if you did make a change to the address space. But if you did not, then it would also not re-run. Which might be ok in this specific case.

lulf: It is assumed, that if this happens, and you actually made a change to the object, then you call replace and will get a notification of the change, so I'm not sure if we need to explicitly re-trigger.

ctron: I would just add the requirement to some javadoc :)

lulf:

lulf:

lulf:

lulf:

lulf:
@enmasse-ci run tests testcase=ConnectorsTest

lulf:
@enmasse-ci run tests testcase=ConnectorsTest

lulf:
@enmasse-ci run tests testcase=ConnectorsTest

PR #4089, opened 2020-03-11
Shared infra proposal improvements

lulf:
### Type of change

- Documentation

### Description

Modifications of proposal based on meeting in Brno.

### Checklist

- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/documentation/design/proposals/shared-infrastructure.adoc
+++ b/documentation/design/proposals/shared-infrastructure.adoc
@@ -131,7 +131,9 @@ When deployed, the above would have the minimum impact of creating 2 routers and
 
 ==== MessagingPlan
 
-The `MessagingPlan` defines limits that map to the routers per vhost policies:
+The `MessagingPlan` defines limits that map to a particular namespaces' share of the infrastructure:
+
+For instance, routers per vhost policies:
 
 * connections
 * links per connection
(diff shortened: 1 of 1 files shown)


documentation/design/proposals/shared-infrastructure.adoc
```diff
@@ -382,9 +376,9 @@ The following would NOT be supported:
 
 The tasks of managing brokers and routers should be offloaded to standalone components as much as possible.
 
-For router deployments, the https://github.com/interconnectedcloud/qdr-operator[qdr-operator] will be used and it supports all features required by EnMasse.
+For router deployments, the builtin mesh-forming support of the router image will be used. All router configuration except basic static config is applied using AMQP management, as the router does not have a mechanism to distinguish between static and dynamic configuration.
 
-For broker deployments, the https://github.com/rh-messaging/activemq-artemis-operator[activemq-artemis-operator] will be used, (and modified to work with our requirements). The resulting changes should be submitted upstream, although short term there may exist a temporary fork in order to make progress.
+For broker deployments, a standard upstream broker image will be used (once provided, in the meantime use existing image). All broker configuration defined in MessagingInfra is applied statically to the broker.xml and makes use of the auto-reload feature of the broker config.
```
k-wall: I'm curious to understand why using static configuration is better than the approach we already take?    We will continue to need AMQP management for monitoring purposes, so we will still need this interface.

Is leaning on this feature for making changes to a live system a supported use-case from a Broker perspective?

k-wall: Question withdrawn

documentation/design/proposals/shared-infrastructure.adoc
```diff
@@ -430,50 +424,38 @@ Routers will be periodically queried (by independent goroutines) for:
...
 The controller should:
 
 * Watch MessagingInfra CR
-* Creates router CR to deploy routers based on infra config
-* Creates broker CRs to deploy brokers based on infra config and scale on-demand
-* Apply to router CR: router <-> broker connectors
+* Creates router statefulset to deploy routers based on infra config and using input from system metrics to adjust the number of replicas
+* Creates broker statefulset to deploy brokers based on infra config and using input from system metrics to adjust the number of replicas
 * Creates interal state representation for each router and broker in the CR status. This state should be shared with other controllers (details below)
 * Expose metrics about connections and links which is used by the console, either through HTTP or AMQP
```
k-wall: What component will take responsibility for providing management operations such as connection close, purge queue, move message?  The console (and eventually the CLI) will need access to these functions.    Should this component also be responsible for for producing this metrics?

documentation/design/proposals/shared-infrastructure.adoc
lulf: @k-wall I think the setting in StandardInfra/BrokeredInfra today is also static and requires a restart. We could make it dynamic, but its also not clear to me which entities support dynamic config over management, so I _think_ static is ok. We might want to make acceptors dynamic though.

documentation/design/proposals/shared-infrastructure.adoc
lulf: I think this point is a leftover from before our Brno discussion, where we discussed if it should be in the operator or not.

The idea of having a separate component has grown on me. I think conceptually, that responsibility belongs to a different component than the operator. Mainly because management operations and metric collection is not related to changing the configuration of the router or broker, which I consider the primary function of the operator.

I also think making this a standalone component would allow more flexibility in changing the memory needed for that component to maintain the connection and link information in cases where there are many of them, and provide isolation so that reconciliation is not affected should the management process have issues.

By metrics, I assume metrics for the console, which I think that component should produce.

documentation/design/proposals/shared-infrastructure.adoc
```diff
@@ -12,23 +12,33 @@ This document will consider the alternative with a new address space type + an a
...
+
+* `MessagingTenant`: a singleton per namespace, is introduced to contain the messaging configuration of a namespace. Auto-created when initial `Address` is created. May be used to override the default `MessagingPlan` and `AuthenticationService` of a tenant.
+
+* `MessagingConnector`: similar to existing connector spec in AddressSpace. 1 instance configures 1 connector.
+
+* `MessagingEndpoint`: allows configuring endpoints to access messaging infrastructure. Cluster-local endpoint is created by default on initial `Address` creation.
 
-MessagingUser and Address will no longer needed to be prefixed with address space name!
+* `Address`: similar to existing type. Name is no longer prefixed, `.spec.address` is optional. `.spec.plan` is optional if permitted by plan config. Addition per-type sub-objects such as `.spec.queue` or `.spec.topic` may be specified to override plan properties (if allowed by plan).
```
famartinrh: by "if permitted by plan config" it's referring to MessagingPlan right?

lulf: It is referring to the AddressPlan which covers per-address limits. The MessagingPlan covers per-namespace limits spanning all addresses in a namespace. I will make the sentence clearer, thanks!

famartinrh: but how can be the .spec.plan of an address be optional or not if permitted by AddressPlan config, that was my point

lulf: It is optional in the sense that the tenant does not have to specify it. However, if there is an addressplan with a selector for the namespace that address is created in, that plan will be the “default” if no plan is set. 

If there are multiple addressplans , we can select the default based on some annotation (similar to storage class for kubernetes). We can also define that if there is a plan with matching selector but without any annotation indicating it should be default, we require the user to specify the plan.

famartinrh: I got it now, thanks Ulf :)

PR #4102, opened 2020-03-13
Systemtests scale tests - messaging performance

famartinrh:
### Type of change

- Test

### Checklist

- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/systemtests/src/main/java/io/enmasse/systemtest/platform/Kubernetes.java
+++ b/systemtests/src/main/java/io/enmasse/systemtest/platform/Kubernetes.java
@@ -483,7 +483,11 @@ public abstract class Kubernetes {
     }
 
     public List<Deployment> listDeployments(Map<String, String> labels) {
-        return client.apps().deployments().inNamespace(infraNamespace).withLabels(labels).list().getItems();
+        return listDeployments(infraNamespace, labels);
+    }
+
+    public List<Deployment> listDeployments(String namespace, Map<String, String> labels) {
+        return client.apps().deployments().inNamespace(namespace).withLabels(labels).list().getItems();
     }
 
     public List<StatefulSet> listStatefulSets(Map<String, String> labels) {
@@ -564,10 +568,14 @@ public abstract class Kubernetes {
      * @throws Exception whe deployment failed
      */
     public void createDeploymentFromResource(String namespace, Deployment resources) throws Exception {
+        createDeploymentFromResource(namespace, resources, 2, TimeUnit.MINUTES);
+    }
+
+    public void createDeploymentFromResource(String namespace, Deployment resources, long time, TimeUnit unit) throws Exception {
         if (!deploymentExists(namespace, resources.getMetadata().getName())) {
             Deployment depRes = client.apps().deployments().inNamespace(namespace).create(resources);
             Deployment result = client.apps().deployments().inNamespace(namespace)
-                    .withName(depRes.getMetadata().getName()).waitUntilReady(2, TimeUnit.MINUTES);
+                    .withName(depRes.getMetadata().getName()).waitUntilReady(time, unit);
             log.info("Deployment {} created", result.getMetadata().getName());
         } else {
             log.info("Deployment {} already exists", resources.getMetadata().getName());
--- a/systemtests/src/main/java/io/enmasse/systemtest/platform/Kubernetes.java
+++ b/systemtests/src/main/java/io/enmasse/systemtest/platform/Kubernetes.java
@@ -568,17 +568,21 @@ public abstract class Kubernetes {
      * @throws Exception whe deployment failed
      */
     public void createDeploymentFromResource(String namespace, Deployment resources) throws Exception {
-        createDeploymentFromResource(namespace, resources, 2, TimeUnit.MINUTES);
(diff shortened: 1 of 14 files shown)


systemtests/src/main/java/io/enmasse/systemtest/platform/Kubernetes.java
```diff
@@ -564,13 +568,21 @@ public void deletePod(String namespace, String podName) {
...
+    }
+
+    public void createDeploymentFromResource(String namespace, Deployment resources, long time, TimeUnit unit, boolean doLog) throws Exception {
         if (!deploymentExists(namespace, resources.getMetadata().getName())) {
             Deployment depRes = client.apps().deployments().inNamespace(namespace).create(resources);
             Deployment result = client.apps().deployments().inNamespace(namespace)
-                    .withName(depRes.getMetadata().getName()).waitUntilReady(2, TimeUnit.MINUTES);
-            log.info("Deployment {} created", result.getMetadata().getName());
+                    .withName(depRes.getMetadata().getName()).waitUntilReady(time, unit);
+            if (doLog) {
```
lulf: Is it really necessary to enable/disable logging per method invocation? Does this belong somewhere at the class level or Environment?

systemtests/src/main/java/io/enmasse/systemtest/scale/MessagingPerformanceTestManager.java
```diff
@@ -0,0 +1,150 @@
...
+                .toArray(String[]::new);
+        var clientConfig = clientProvider.get();
+        clientConfig.setClientType(ScaleTestClientType.messaging);
+        clientConfig.setAddressesType(type);
+        clientConfig.setAddresses(addressesOfType);
+        clientConfig.setLinksPerConnection(linksPerConnection);
+
+        SystemtestsKubernetesApps.deployScaleTestClient(kubernetes, clientConfig);
+
+        int connectionsInThisClient = (addressesOfType.length/linksPerConnection) * 2; // *2 because client creates sender and receiver
```
lulf: We should make this configurable, i.e. for the IoT case where you might have a lot more senders than receivers.

systemtests/src/main/java/io/enmasse/systemtest/scale/ScaleTestClient.java
```diff
@@ -0,0 +1,37 @@
+/*
+ * Copyright 2019, EnMasse authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+package io.enmasse.systemtest.scale;
+
+import io.enmasse.systemtest.model.address.AddressType;
+import io.enmasse.systemtest.scale.metrics.ScaleTestClientMetricsClient;
+
+public class ScaleTestClient <X extends ScaleTestClientMetricsClient>{
```
lulf: I think the convention is to use T as the first generic parameter, we should stick to that.

systemtests/src/main/java/io/enmasse/systemtest/scale/metrics/MetricsAssertions.java
```diff
@@ -0,0 +1,57 @@
...
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+package io.enmasse.systemtest.scale.metrics;
+
+import java.util.Optional;
+import java.util.function.Predicate;
+
+import org.junit.jupiter.api.Assertions;
+
+public class MetricsAssertions<X> implements Predicate<X> {
```
lulf: X -> T

systemtests/src/main/java/io/enmasse/systemtest/time/TimeMeasuringSystem.java
```diff
@@ -37,6 +37,8 @@
     private String testClass;
     private String testName;
 
+    private boolean logResults = true;
```
lulf: Is it better enabled/disabled via Environment.java ?

systemtests/src/main/java/io/enmasse/systemtest/platform/Kubernetes.java
famartinrh: with this changes I only disabled the logging in a few places where it was too verbose for the scale test, I'm not sure I want to disable logging for this full class, altough I agree this is not the most elegant solution

systemtests/src/main/java/io/enmasse/systemtest/scale/MessagingPerformanceTestManager.java
famartinrh: that will require to implement some login in the client, but it's ok and I like it, so I'll add it to my TODO list

systemtests/src/main/java/io/enmasse/systemtest/time/TimeMeasuringSystem.java
famartinrh: agree, but first need support to set that env var in our pipelines

famartinrh:
@lulf yes I would prefer that

PR #4116, opened 2020-03-14
Dont run scale in isolated profile

kornys:
### Type of change

- Bugfix


kornys:

PR #4120, opened 2020-03-16
Systemtests - failure recovery test

famartinrh:
### Type of change

- Test

### Checklist

- [ ] Update documentation
- [ ] Update CHANGELOG.md


systemtests/src/main/java/io/enmasse/systemtest/scale/ScalePerformanceTestManager.java
```diff
@@ -161,10 +197,56 @@ public void monitorMetrics() {
...
+            }
+
+        }
+
+        double sum = 0;
+        for (var a : averages) {
+            sum += a;
+        }
+        double average = sum / averages.size();
+        return Duration.ofSeconds((long) average);
```
lulf: I would recommend that we use a percentile such as 99%, as that is easiest to give guarantees about. Averages can lie and hide real issues. You should be able to get that from the histogram bucket for that percentile.

famartinrh: I have just addressed this comment, wdyt?

PR #4124, opened 2020-03-16
Allow configuring console to use impersonation

### LINKED ISSUE #4117: Support user impersonation in console
**Overview**

It is currently not possible to use the console when installing EnMasse on a managed Kubernetes service, or if one wishes to use an OIDC provider without having to configure the Kubernetes API server with that provider.

To handle that, the console server could support impersonation of users, so that instead of using the user token to authorize, impersonation headers are used instead.

**Design**

To enable this feature, the ConsoleService CRD would be extended:

```
.spec.impersonation # Enables impersonation mode - default header settings are used
.spec.impersonation.userHeader # Customize which header from oauth-proxy to use as 'user' header. Default is X-Forwarded-User
```

When this is enabled, the oauth-proxy will get an extra flag '-pass-basic-auth' which will set the headers 'X-Forwarded-User', 'X-Forwarded-Email' and 'X-Forwarded-Preferred-Username'. the impersonation.userHeader setting can be used to select which of these headers should be used as the user being impersonated.

If impersonation is enabled, the console-server will configure the clients it uses in 'server/handler.go' so that they use impersonation rather than the user token.

lulf:
### Type of change

- Enhancement / new feature

### Description

This extends the consoleservice configuration with settings to control
the use of impersonation of the console (disabled by default).

Fixes #4117 

### Checklist

- [ ] Update documentation
- [x] Update CHANGELOG.md


lulf:

templates/console-server/020-ClusterRole-console-server.yaml
```diff
@@ -11,3 +11,6 @@ rules:
   - apiGroups: [ "" ]
     resources: [ "namespaces" ]
     verbs: [ "get", "list", "watch" ]
+  - apiGroups: [ "authorization.k8s.io" ]
+    resources: [ "subjectaccessreviews" ]
```
k-wall: what part does this play?  I don't see a change making a subjectaccessreviews.

templates/crds/consoleservices.crd.yaml
```diff
@@ -60,6 +60,17 @@ spec:
               type: string
             host:
               type: string
+            impersonation:
+              type: object
+              description: "Enable impersonation as the mechanism for authorizing user actions. For impersonation to work, it is important to also grant the console-server permissions to impersonate users."
```
k-wall: How exactly is the ability to impersonate enabled?  My own curiosity and should we include a pointer to the relevant part of the kubernetes documentation in the EnMasse kubernetes docs?

templates/console-server/020-ClusterRole-console-server.yaml
lulf: It was something I added earlier as I was changing it to use subjectaccessreviews when using impersonation. However, that seems to no longer be necessary, so I've removed that now. THanks!

templates/crds/consoleservices.crd.yaml
lulf: What I have not yet done is to add some docs on how to use this. To be able to impersonate, the console-server service-account must be granted 'impersonate' verb on a resource named 'users'. However, this permission is considered very broad, and not something I think we should add to the default permissions. 

I was thinking that having that role + rolebinding in a separate 'example-console' folder or something, and then refer to that in the documentation. Wdyt? Agreed on referring the kubernetes docs.

pkg/consolegraphql/accesscontroller/access_controller.go
```diff
@@ -91,6 +91,7 @@ func (k *kubernetesRbac) CanRead(obj interface{}) (bool, error) {
 			Resource:  plural.Resource, // Resource is Kind
 			Version:   gvk.Version,
 			Namespace: o.Name,
+			Name:      o.Name,
```
k-wall: when I wrote access_controller.go, i was guided by what `oc auth can-i get namespaces --loglevel=9` did.  It passes only the `namespace` field and not the `name`.

pkg/consolegraphql/server/handler.go
```diff
@@ -32,19 +33,32 @@ const loggedOnUserSessionAttribute = "loggedOnUserSessionAttribute"
...
 	return http.HandlerFunc(func(rw http.ResponseWriter, req *http.Request) {
 		var state *RequestState
 
-		accessToken := getAccessToken(req)
 		useSession := true
 		if healthProbe, _ := strconv.ParseBool(req.Header.Get("X-Health")); healthProbe {
 			useSession = false
 		}
 
+		accessToken := getAccessToken(req)
```
k-wall: In the impersonating case, whose accessToken is it?

pkg/consolegraphql/accesscontroller/access_controller.go
lulf: `oc auth can-i get namespaces --loglevel=9` checks if you can 'get' namespaces in general, but not a specific namespace, which is the permission I was using.

If you create a role using resources: ["namespaces"] and resourceNames: ["foo"] to grant permissions to a single namespace, then omitting the 'Name' in the ssar will cause it to fail.

pkg/consolegraphql/server/handler.go
lulf: The console-server service-account

lulf: It is still the token forwarded by the oauth proxy, but its not passed on to the kubernetes client if impersonation is configured (kube client uses the service-account token by default I think). If my understanding of the code is correct, the accessToken in impersonation case would only be used to keep track of the session.

pkg/consolegraphql/accesscontroller/access_controller.go
k-wall: so that would sound like an existing defect against 0.31?  however, I know we test case where the user has access to a only a single namespace and the namespace filtering works as expected.

k-wall:
Keep in mind that the accessToken is propagated down to the agent over a AMQP SASL X-OAUTH exchange to allow the user to perform operations such as purge.

pkg/consolegraphql/accesscontroller/access_controller.go
lulf: I'll check if this is only a problem for Kubernetes or not. I think that if the namespace is created by the user, then you automatically get some permissions applied. In my case, the namespace is created by someone else, and the user has no permissions at all to begin with.

lulf:
@k-wall So I wonder how best to handle agent authentication in this case. I was thinking that ideally the impersonation headers would be forwarded to agent as well. However, I feel that granting agent impersonation privileges is a bit excessive and I'm not sure if there is a way to pass such properties in the SASL handshake.

Another alternative is to have console-server pass its own token to agent, and the agent would essentially always allow any operation that console-server could make. This would effectively disable finer-grained permissions for users I guess, but maybe thats acceptable for now.

templates/crds/consoleservices.crd.yaml
lulf: @k-wall I've confirmed that this issue happens on OpenShift as well. If I create a user 'test1', and then apply the following as a cluster-admin user: 

```
---
apiVersion: v1
kind: Namespace
metadata:
  name: test1
---
kind: ClusterRole
metadata:
  labels:
    app: enmasse
  name: test1
rules:
  - apiGroups: [ "" ]
    resources: [ "namespaces" ]
    resourceNames: [ "test1" ]
    verbs: [ "get" ]
---
kind: ClusterRoleBinding
metadata:
  name: test1
  labels:
    app: enmasse
roleRef:
  kind: ClusterRole
  name: test1
subjects:
- kind: User
  name: user1
```

The 'test1' user is unable to see the namespace. However, if I modify the console-server to pass the Name in addition to Namespace, it works. I've raised a PR with a fix https://github.com/EnMasseProject/enmasse/pull/4133

lulf:
@k-wall I've gone ahead and made use of the username field in the sasl handshake. If non-empty, it signals that impersonation should be used.

Moreover, I changed agent to use TokenReviews for authenticating users, as the user api does not exist on Kubernetes (and consequently users cannot use purge functionality on Kubernetes). 

I considered either having an environment variable specifying if openshift or kubernetes is used, or doing a fallback to tokenreviews. However, since the clusterrole tokenreview would have to be applied to the agent service account anyway, I figured it would probably be ok to always use tokenreviews.

lulf:

pkg/consolegraphql/server/handler.go
k-wall: The think the session tracking would need to be keyed on the impersonation info.  The session tracking is required because when a user logs out of oauth-proxy,  there console-server session will live on.  The the session tracking was absent, if the same browser (same cookies) was used to authenticate a different user, the new user would pickup the old user's  console server session.

pkg/consolegraphql/resolvers/resolver_user.go
```diff
@@ -19,12 +19,24 @@ func (r *queryResolver) Whoami(ctx context.Context) (*v1.User, error) {
...
+		if requestState.ImpersonateUser {
+			return &v1.User{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: requestState.User,
+				},
+				FullName:   requestState.User,
+				Identities: []string{requestState.User},
+			}, nil
+		} else {
+			// TODO: Use TokenReview's?
```
k-wall: Ah, I see you'd spotted my mistake here #4172

pkg/consolegraphql/server/handler.go
lulf: @k-wall I've updated the PR and added impersonated user to the sha digest.

lulf:

lulf:
@enmasse-ci run tests profile=shared testcase=web.FirefoxConsoleTest

lulf:

lulf:

documentation/modules/proc-configure-external-openid-connect-for-kubernetes.adoc
```diff
@@ -0,0 +1,63 @@
+// Module included in the following assemblies:
+//
+// assembly-configuring.adoc
+
+[id='config-external-openid-connect-for-kubernetes']
+= Configuring {ProductName} Console to use cluster OpenID Connect
```
k-wall:
```suggestion
= Configuring {ProductName} Console to use external OpenID Connect
```

templates/example-console/020-ClusterRole-console-server-impersonate.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  labels:
+    app: enmasse
+  name: enmasse.io:console-server-impersonation
+rules:
+  - apiGroups: [ "" ]
+    resources: [ "users" ]
+    verbs: [ "impersonate "]
```
k-wall:
```suggestion
    verbs: [ "impersonate"]
```

templates/example-console/020-ClusterRole-console-server-impersonate.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  labels:
+    app: enmasse
+  name: enmasse.io:console-server-impersonation
```
k-wall:
```suggestion
  name: enmasse.io:console-server-impersonate
```

lulf:
@k-wall thanks for the typo fixes. I honestly don’t know how those got in as I tested this with the sandbox. Hope you didn’t spent a lot of time figuring that one out.

pkg/controller/consoleservice/consoleservice_controller.go
```diff
@@ -866,6 +866,13 @@ func applyDeployment(consoleservice *v1beta1.ConsoleService, deployment *appsv1.
 			install.RemoveEnv(container, "HTTP_SESSION_IDLE_TIMEOUT")
 		}
 
+		if consoleservice.Spec.Impersonation != nil {
+			install.ApplyEnvSimple(container, "IMPERSONATION_ENABLE", "true")
+			if consoleservice.Spec.Impersonation.UserHeader != "" {
+				install.ApplyEnvSimple(container, "IMPERSONATION_USER_HEADER", consoleservice.Spec.Impersonation.UserHeader)
+			}
+		}
```
k-wall:
```suggestion
		} else {
			install.RemoveEnv(container, "IMPERSONATION_ENABLE")
		}
```

PR #4130, opened 2020-03-17
fix route, create single typeahead component

drcoolsanjeev:
### Type of change

- Refactoring

### Description

- create single typeAhead component to use in all places
- use some interfaces to avoid any
- structuring code in folder format
- fix child route
- use TablePagination component instead of renderPagination


console/console-init/ui/src/components/TypeAheadSelect/TypeAheadSelect.tsx
```diff
@@ -0,0 +1,109 @@
...
+  SelectOption,
+  SelectOptionObject
+} from "@patternfly/react-core";
+import {
+  TypeAheadMessage,
+  TYPEAHEAD_REQUIRED_LENGTH
+} from "constants/constants";
+import { ISelectOption } from "utils";
+
+interface ITypeAheadSelectProps {
```
ajaypratap003: an you try to use SelectProps to get all the properties, We should avoid to create unnecessary properties those already are available in importing component interface. here you can extend SelectProps and avoid the properties declare those are already threre.
interface ITypeAheadSelectProps extends SelectProps {
//declare properties which are not available in SelectProps
}

we will follow this practice for all the components.

console/console-init/ui/src/components/TypeAheadSelect/TypeAheadSelect.tsx
```diff
@@ -0,0 +1,109 @@
...
+          disabled={true}
+        />
+      ) : (
+        <SelectOption
+          key={"no-results-found"}
+          value={TypeAheadMessage.NO_RESULT_FOUND}
+          disabled={true}
+        />
+      )}
+      {/* {} */}
```
ajaypratap003: is there any significance of this commented line?

console/console-init/ui/src/modules/address-detail/containers/AddressLinks/AddressLinksPage.tsx
```diff
@@ -118,7 +78,12 @@ const AddressLinksPage: React.FunctionComponent<IAddressLinksListPageProps> = ({
             />
           </GridItem>
           <GridItem span={5}>
-            {addresLinksTotal > 0 && renderPagination(page, perPage)}
+            <TablePagination
```
anukritijha: The same component is being used with different values, better way would be to create a method like renderPagination like we had earlier and pass different arguments to it in both places.

console/console-init/ui/src/modules/address/AddressPage.tsx
```diff
@@ -342,7 +302,12 @@ export default function AddressPage() {
           />
         </GridItem>
         <GridItem span={5}>
-          {totalAddresses > 0 && renderPagination(page, perPage)}
+          <TablePagination
```
anukritijha: Similar to my previous comment, please encapsulate this in a method.

console/console-init/ui/src/components/TypeAheadSelect/TypeAheadSelect.tsx
drcoolsanjeev: Yes, i remove this then there will be a TS lint error. I don't know why. But some unstrange things happening

PR #4133, opened 2020-03-17
Pass name of namespace to check access to

lulf:
### Type of change

- Bugfix

### Description

This prevents an issue where the user is granted specific access to a
namespace (using resourceNames in the role), but it will not be able to
see it.

### Checklist

- [ ] Update documentation
- [ ] Update CHANGELOG.md


k-wall:
@lulf we added this case to catch this use case  `io.enmasse.systemtest.isolated.web.FirefoxConsoleTest#testCreateAddrSpaceNonClusterAdmin `already.  It creates a addresspace (from the UI) as a user who has access to one namespace.  In what way is your scenario different?

lulf:
@k-wall I can spot 2 differences:

1. User creates namespace - in my case the namespace is created by a different/admin user
2. User has admin rolebinding - which I think automatically grants some resources. In my case there are no permissions beside get on a specific namespace + ability to create address spaces and addresses in that namespace.

k-wall:
This isn't working for me in two ways:

I created a user test1 and apply the yaml from https://github.com/EnMasseProject/enmasse/pull/4124#discussion_r393935265

First problem is the I can't get past the login page.   After a successful login, I get redirected straight back to the oauth-proxy.   It is as if oauth-proxy itself somehow doesn't like the user configured in this way.  I haven't yet worked out why.   I can see that it is not getting as far as the console-server.

Second problem, if I then go to the console's url (pasting the console url back into the same browser) I get into the console, but when I try to create an address space, the namespace dropdown is empty.

Just for confirmation I created a second user test2, created a namespace on the common whilst logged on as him and then use the console.  Neither problem manifested.

k-wall:
> This isn't working for me in two ways:
> 
> I created a user test1 and apply the yaml from [#4124 (comment)](https://github.com/EnMasseProject/enmasse/pull/4124#discussion_r393935265)
> 
> First problem is the I can't get past the login page. After a successful login, I get redirected straight back to the oauth-proxy. It is as if oauth-proxy itself somehow doesn't like the user configured in this way. I haven't yet worked out why. I can see that it is not getting as far as the console-server.
> 
> Second problem, if I then go to the console's url (pasting the console url back into the same browser) I get into the console, but when I try to create an address space, the namespace dropdown is empty.
> 
> Just for confirmation I created a second user test2, created a namespace on the common whilst logged on as him and then use the console. Neither problem manifested.

Testing with the permission user `user1`, I am seeing correct behaviour/

lulf:
@enmasse-ci run tests testcase=web.FirefoxConsoleTest#testCreateAddrSpaceNonClusterAdmin

lulf:
@enmasse-ci run tests testcase=web.FirefoxConsoleTest

k-wall:
@enmasse-ci run tests testcase=web.FirefoxConsoleTest

lulf:
@enmasse-ci run tests testcase=web.FirefoxConsoleTest

PR #4139, opened 2020-03-18
refactor in address and addres-detail module

drcoolsanjeev:
### Type of change

- Refactoring

### Description

- Removed some inline styles
- created a enum for addressTypes to avoid using string
- Break Create address configuration into component and container


console/console-init/ui/src/modules/address/dialogs/CreateAddress/Configuration.tsx
```diff
@@ -0,0 +1,191 @@
...
+  const { loading, error, data } = useQuery<IAddressTypes>(
+    RETURN_ADDRESS_TYPES,
+    {
+      variables: {
+        a: addressSpaceType
+      }
+    }
+  );
+  if (loading) return <Loading />;
+  if (error) return <Loading />;
```
anukritijha: I don't think we need a condition for errors here anymore

console/console-init/ui/src/modules/address/dialogs/CreateAddress/Configuration.tsx
```diff
@@ -0,0 +1,191 @@
+/*
```
anukritijha: Can we qualify this as a container, as it's mainly calling graphQL for different scenarios?

console/console-init/ui/src/modules/address/dialogs/CreateAddress/Configuration.tsx
drcoolsanjeev: I will remove this in next PR.

console/console-init/ui/src/modules/address/dialogs/CreateAddress/Configuration.tsx
drcoolsanjeev: Yeah , but we discussed to put all modals like create and edit in dialogs folder

PR #4140, opened 2020-03-18
Add additional test only

lulf:


lulf:
@enmasse-ci run tests testcase=web.FirefoxConsoleTest#testCreateAddrSpaceNonClusterAdminMinimal

lulf:
Fails because it wasn't able to delete address space (because it wasn't able to create, probably test check could be clearer @kornys ), so expected.

PR #4148, opened 2020-03-19
Wait for prometheus to be running before restarting operator

lulf:


lulf:
@enmasse-ci run tests testcase=MonitoringTest

PR #4150, opened 2020-03-19
Run systemtests as tekton pipeline

kornys:
### Type of change

- Enhancement / new feature

### Description

Ability to run enmasse systemtests as tekton pipeline on ocp4, this is only draft, it contains only fer params and it can test only master right now. But I will work more on that.

### Checklist

- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -127,8 +127,8 @@ $(DOCKER_DIRS):
 systemtests:
 	make -C systemtests
 
-tekton-test:
-	make -C systemtests tekton-test
+systemtests-tekton:
+	make -C systemtests systemtests-tekton
 
 docu_html:
 	make -C documentation build
--- a/Makefile
+++ b/Makefile
@@ -127,6 +127,9 @@ $(DOCKER_DIRS):
 systemtests:
 	make -C systemtests
 
+tekton-test:
+	make -C systemtests tekton-test
+
 docu_html:
 	make -C documentation build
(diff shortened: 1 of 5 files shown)


lulf:
@kornys I think we should consider moving this out of systemtests either into a toplevel module (tekton or .tekton) where we can create other pipelines for building images as well. Also an additional option would be to put this in a separate repository.

kornys:
@lulf yes I agree, we can move it i to .tekton and add there many others tasks and pipelines. I was thinking about ability to build you images, push into ocp registries and run systemtests. All steps with tekton

lulf:
Anything missing here before we can merge it?

kornys:
@lulf yes, I want to update that to use your current code from local repo and add ability to build images

kornys:
@lulf I think it can be merged if you want, in current state it is able to run master code of systemtests, as next PR I will add more parameters, and off course pipeline which build images from branch

Makefile
```diff
@@ -127,6 +127,9 @@ $(DOCKER_DIRS):
 systemtests:
 	make -C systemtests
 
+tekton-test:
```
lulf, line 130: Maybe name it systemtests-tekton to make the names a bit more similar and logical?

kornys: Renamed

PR #4157, opened 2020-03-20
Test verifying Agent's behaviour when connection to the Kubernetes API server is lost

k-wall:
### Type of change

- Enhancement / new feature

### Description

System test verifying behaviour when connection to the Kubernetes API server is lost

### Checklist

- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/systemtests/src/main/java/io/enmasse/systemtest/platform/KubeCMDClient.java
+++ b/systemtests/src/main/java/io/enmasse/systemtest/platform/KubeCMDClient.java
@@ -253,6 +253,24 @@ public class KubeCMDClient {
         return Exec.execute(deleteCmd, DEFAULT_SYNC_TIMEOUT, true);
     }
 
+    public static ExecutionResultData setResourceEnvVarByLabel(String resourceType, Optional<String> containers, String resourceName, String envVarName, String envVarValue) {
+        ArrayList<String> setEnvCmd = new ArrayList<>();
+        setEnvCmd.add(CMD);
+        setEnvCmd.add("set");
+        setEnvCmd.add("env");
+        setEnvCmd.add(resourceType);
+        setEnvCmd.add(resourceName);
+        containers.ifPresent(s -> setEnvCmd.add("--containers=" + s));
+
+        if (envVarValue == null) {
+            setEnvCmd.add(envVarName + "-");
+        } else {
+            setEnvCmd.add(envVarName + "=" + envVarValue);
+        }
+
+        return Exec.execute(setEnvCmd, DEFAULT_SYNC_TIMEOUT, true);
+    }
+
     public static ExecutionResultData runOnPod(String namespace, String podName, Optional<String> container, String... args) {
         List<String> runCmd = new ArrayList<>();
         String[] base = container.map(s -> new String[]{CMD, "exec", podName, "-n", namespace, "--container", s, "--"})
--- a/systemtests/src/main/java/io/enmasse/systemtest/platform/KubeCMDClient.java
+++ b/systemtests/src/main/java/io/enmasse/systemtest/platform/KubeCMDClient.java
@@ -253,7 +253,7 @@ public class KubeCMDClient {
         return Exec.execute(deleteCmd, DEFAULT_SYNC_TIMEOUT, true);
     }
 
-    public static ExecutionResultData setResourceEnvVarByLabel(String resourceType, Optional<String> containers, String resourceName, String envVarName, String envVarValue) {
+    public static ExecutionResultData setResourceEnvVarByLabel(Optional<String> namespace, String resourceType, Optional<String> containers, String envVarName, String resourceName, String envVarValue) {
         ArrayList<String> setEnvCmd = new ArrayList<>();
         setEnvCmd.add(CMD);
         setEnvCmd.add("set");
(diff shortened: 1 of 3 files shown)


k-wall:
Originally I thought we could apply the same technique to the standard-controller and the address-space-controler but as these components, properly validate trust, this fails.

k-wall:
@enmasse-ci run tests testcase=CommonTest  test_only

k-wall:
@enmasse-ci run tests testcase=CommonTest test_only

k-wall:
> Would an alternative to use a separate proxy process be to create an openshift route to the api server and delete it to simulate failure?

Interesting idea. Less stuff is always appealing. You mean create a openshift route to kubernetes.default.svc?   Is it possible for a route to reference a service in another namespace?

lulf:
> Interesting idea. Less stuff is always appealing. You mean create a openshift route to kubernetes.default.svc? Is it possible for a route to reference a service in another namespace?

Yes, but you'd have to create the route in the 'default' namespace, which I think the tests have permission to do anyway.

k-wall:
I tested your approach manually (a route in the default namespace to `kubernetes.default.svc` with TLS `passthrough`), it works but a drawback and some different behaviour which I like to discuss.

* Deleting the route does not break the *established* connections.  (I think this is expected behaviour.  The OpenShift route must implement a socket listener somewhere. Once the socket connection is established, I think its role is at an end).  The upshot for this test is the Agent's existing watch doesn't fail.  However when it times out (we could tune RESYNC_INTERVAL for the nest), the next attempted connection does.   The proxy approach doesn't suffer from this, as the proxy dies, the established connection goes with it.

* When the route is absent, the Agent sees OpenShift's HTML Application in not available (503 service temp unavailable).  The Agent handles this error fine and the error path is the same from the Agent's point of view.    With the proxy approach, the Agent sees a resolution error when poxy deployment is scaled down.

btw Neither approach work of the box for standard-controller.  Proxy approach fails the TLS owing to trust (unexpected cert), Route approach passes TLS trust but fails the hostname validation. If we wanted to test standard-controller (and address-space-controller and the operator), we'd need a switch to disable trust for the system test, which displeases.

k-wall:
@enmasse-ci run tests testcase=CommonTest test_only

k-wall:
@enmasse-ci run tests testcase=CommonTest test_only

lulf:
> I tested your approach manually (a route in the default namespace to `kubernetes.default.svc` with TLS `passthrough`), it works but there is a drawback and some different behaviour which I like to discuss.
> 
>     * Deleting the route does not break the _established_ connections.  (I think this is expected behaviour.  The OpenShift route must implement a socket listener somewhere. Once the socket connection is established, I think its role is at an end).  The upshot for this test is the Agent's existing watch doesn't fail.  However when it times out (we could tune RESYNC_INTERVAL for the test), the next attempted connection does.   The proxy approach doesn't suffer from this, as the proxy dies, the established connection goes with it.
> 
>     * When the route is absent, the Agent sees OpenShift's HTML Application in not available (503 service temp unavailable).  The Agent handles this error fine and the error path is the same from the Agent's point of view.    With the proxy approach, the Agent sees a resolution error when proxy deployment is scaled down.
> 
> 
> btw Neither approach work of the box for standard-controller. Proxy approach fails the TLS owing to trust (unexpected cert), Route approach passes TLS trust but fails the hostname validation. If we wanted to test standard-controller (and address-space-controller and the operator), we'd need a switch to disable trust for the system test, which displeases.
> 
> WDYT?

I'm fine with either approach but maybe the api-proxy approach is the most realistic, so I'd say go for that. For the standard-controller we could pass controller parameters to instruct it to trust everything for instance. However, since I think we haven't seen the standard-controller having this issue I'm relaxed about adding that to the mix.

k-wall:
> > I tested your approach manually (a route in the default namespace to `kubernetes.default.svc` with TLS `passthrough`), it works but there is a drawback and some different behaviour which I like to discuss.
> > ```
> > * Deleting the route does not break the _established_ connections.  (I think this is expected behaviour.  The OpenShift route must implement a socket listener somewhere. Once the socket connection is established, I think its role is at an end).  The upshot for this test is the Agent's existing watch doesn't fail.  However when it times out (we could tune RESYNC_INTERVAL for the test), the next attempted connection does.   The proxy approach doesn't suffer from this, as the proxy dies, the established connection goes with it.
> > 
> > * When the route is absent, the Agent sees OpenShift's HTML Application in not available (503 service temp unavailable).  The Agent handles this error fine and the error path is the same from the Agent's point of view.    With the proxy approach, the Agent sees a resolution error when proxy deployment is scaled down.
> > ```
> > 
> > 
> > btw Neither approach work of the box for standard-controller. Proxy approach fails the TLS owing to trust (unexpected cert), Route approach passes TLS trust but fails the hostname validation. If we wanted to test standard-controller (and address-space-controller and the operator), we'd need a switch to disable trust for the system test, which displeases.
> > WDYT?
> 
> I'm fine with either approach but maybe the api-proxy approach is the most realistic, so I'd say go for that. For the standard-controller we could pass controller parameters to instruct it to trust everything for instance. However, since I think we haven't seen the standard-controller having this issue I'm relaxed about adding that to the mix.

Thanks - I'll keep the api-proxy approach.  I'm not advocating adding coverage for standard controller and other components at the moment.

k-wall:
@kornys would you mind reviewing?

kornys:
@k-wall LGTM in general, good job, just question about test -> do you want to backport this test into 0.30.x or you will merge this test into current master as well?

k-wall:
> @k-wall LGTM in general, good job, just question about test -> do you want to backport this test into 0.30.x or you will merge this test into current master as well?

I think given we have so much non library code in Agent, handling our kubernetes interaction having it in master is valuable as it helps guards against regression.    So I'm thinking master + release-0.30.

PR #4173, opened 2020-03-23
Shared infrastructure - initial CRD and controller implementation

lulf:
* Add the MessagingInfra CRD
* Add the messaginginfra-controller reconciler. This reconciler creates the routers and brokers belonging to an instance of MessagingInfra as well as managing a per-infra CA and per-component certificates signed by that CA. The reconciling loop also takes care of certificate renewal.
* Add support to broker-plugin for deploying the broker in 'shared infra' mode.
* Add a new template component for deploying the enmasse-operator in shared infra mode to avoid impacting existing functionality.
* Add support for generating java types from Go types, only applied to the new MessagingInfra type in the first instance.
* Add initial systemtest for shared infra that validates routers and brokers being deployed and can be scaled.


lulf:

lulf:

lulf:

lulf:

lulf:

lulf:

lulf:

lulf:

lulf:

lulf:

famartinrh:

lulf:

pkg/controller/messaginginfra/broker/controller.go
```diff
@@ -0,0 +1,248 @@
...
+	return &BrokerController{
+		client:         client,
+		scheme:         scheme,
+		certController: certController,
+	}
+}
+
+func getBrokerLabels(infra *v1beta2.MessagingInfra) map[string]string {
+	labels := make(map[string]string, 0)
+	labels["component"] = fmt.Sprintf("broker-%s", infra.Name)
```
ctron, line 46: Wouldn't it make sense to have the infra name also in an dedicated label?

pkg/controller/messaginginfra/broker/controller.go
```diff
@@ -0,0 +1,248 @@
...
+ */
+func (b *BrokerController) ReconcileBrokers(ctx context.Context, logger logr.Logger, infra *v1beta2.MessagingInfra) error {
+	setDefaultBrokerScalingStrategy(&infra.Spec.Broker)
+	logger.Info("Reconciling brokers", "broker", infra.Spec.Broker)
+
+	labels := getBrokerLabels(infra)
+
+	// Update broker condition
+	brokersCreated := infra.Status.GetMessagingInfraCondition(v1beta2.MessagingInfraBrokersCreated)
+	brokersCreated.SetStatus(corev1.ConditionTrue, "", "")
```
ctron, line 65: Wouldn't this always reset the timestamp if the original status was "false"?

pkg/controller/messaginginfra/broker/controller.go
```diff
@@ -0,0 +1,248 @@
...
+				return err
+			}
+		}
+
+		toCreate := numBrokersToCreate(infra.Spec.Broker.ScalingStrategy, brokers.Items)
+		if toCreate > 0 {
+			logger.Info("Creating brokers", "toCreate", toCreate)
+			for i := 0; i < toCreate; i++ {
+				statefulset := &appsv1.StatefulSet{
+					ObjectMeta: metav1.ObjectMeta{Namespace: infra.Namespace, Name: fmt.Sprintf("broker-%s-%s", infra.Name, util.RandomBrokerName())},
```
ctron, line 86: If creating a single stateful set would fail, you would re-queued and re-try the stateful set creation. Having a random component in the name, would creating X additional, rather than the remaining stateful sets.

pkg/controller/messaginginfra/broker/controller.go
```diff
@@ -0,0 +1,248 @@
...
+
+	_, err = b.certController.ReconcileCert(ctx, logger, infra, statefulset)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func int32ptr(v int32) *int32 {
```
ctron, line 215: I would love to have this in a more global package!

pkg/controller/messaginginfra/messaginginfra_controller.go
```diff
@@ -0,0 +1,216 @@
...
+ * Automatically handle status update of the infra after running processor.
+ */
+type infraContext struct {
+	ctx    context.Context
+	client client.Client
+	infra  *v1beta2.MessagingInfra
+	status *v1beta2.MessagingInfraStatus
+}
+
+func (i *infraContext) Process(processor func(infra *v1beta2.MessagingInfra) (reconcile.Result, error)) (reconcile.Result, error) {
```
ctron, line 197: Calling this for every step would cause a huge amount of changes. I am also not quite sure if you can call update twice on the same object. I think the optimistic lock would fail, at least that is what happens when you update the object itself.

pkg/controller/messaginginfra/messaginginfra_controller.go
```diff
@@ -0,0 +1,216 @@
...
+	// Reconcile Brokers
+	err = ic.ProcessSimple(func(infra *v1beta2.MessagingInfra) error {
+		return r.brokerController.ReconcileBrokers(ctx, reqLogger, infra)
+	})
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	// Update main condition
+	err = ic.ProcessSimple(func(infra *v1beta2.MessagingInfra) error {
```
ctron, line 164: It looks like the error is not propagated.

pkg/controller/messaginginfra/router/controller.go
```diff
@@ -0,0 +1,188 @@
...
+		// Set static scaler by default
+		router.ScalingStrategy = &v1beta2.MessagingInfraSpecRouterScalingStrategy{
+			Static: &v1beta2.MessagingInfraSpecRouterScalingStrategyStatic{
+				Replicas: 1,
+			},
+		}
+	}
+}
+
+func int32ptr(val int32) *int32 {
```
ctron, line 180: I think it would make sense to re-use the other implementation.

pkg/controller/messaginginfra/router/controller.go
```diff
@@ -0,0 +1,188 @@
...
+
+		routerConfig := generateConfig(&infra.Spec.Router)
+		routerConfigBytes, err := serializeConfig(&routerConfig)
+		if err != nil {
+			return err
+		}
+		config := &corev1.ConfigMap{
+			ObjectMeta: metav1.ObjectMeta{Namespace: infra.Namespace, Name: routerInfraName},
+			Data: map[string]string{
+				"qdrouterd.json": string(routerConfigBytes),
```
ctron, line 143: I think that when you change the config here, the pods of the stateful set will not pick up the changed configuration. For IoT, I've been using the functionality from the `cchange` package to generate a hash, putting that into the deployments.

Makefile
```diff
@@ -133,5 +137,26 @@ docu_check:
 docu_clean:
 	make -C documentation clean
 
+# Targets related to kubebuilder
```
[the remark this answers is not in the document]
lulf: On my machine, this target will use any controller-gen you have on the path if it exists. If you don't have controller-gen on the path, it will download it and use it, which I think is the behavior you want?

pkg/controller/messaginginfra/broker/controller.go
lulf: IMO this is such a small utility and something that will never change, so keeping it close to the code where it is used increases readability, and there is no question about maintainability in this case.

pkg/controller/messaginginfra/messaginginfra_controller.go
lulf: Hmm, good point. The intention was that status gets updated so that a user can see if an error has occured and/or what changes have actually passed, but I see the problem you're describing.

So I propose that I change this to the following:

If an error is returned, update status with latest changes
If error is not returned, accumulate status changes until all reconcile loops have changed.

I think this is similar to the work you recently did.

pkg/controller/messaginginfra/router/controller.go
lulf: Good point, will fix that.

pkg/controller/messaginginfra/messaginginfra_controller.go
lulf: Thanks, will fix that.

pkg/controller/messaginginfra/broker/controller.go
lulf: I'm not sure I understand the concern. I think the following three scenarios are working right now:

Scenario 1: Statefulset does not get created, reconcile returns error and reconcile is re-run with the initial listBrokers
Scenario 2: Statefulset gets created, but error in creating other resources lead to error. Reconcile returns error, but statefulset is re-read during listBrokers and reconcile continues
Scenario 3: Statefulset gets created, other resources get created. Things proceed as normal.

pkg/controller/messaginginfra/broker/controller.go
lulf: So the timestamp should always be that of the 'initial' error?

pkg/controller/messaginginfra/broker/controller.go
lulf: Agreed, will add that.

api-model-annotator/src/main/java/io/enmasse/api/annotator/KubernetesTypeAnnotator.java
```diff
@@ -0,0 +1,174 @@
...
+ * Copyright (C) 2015 Red Hat, Inc.
+ */
+
+package io.enmasse.api.annotator;
+
+/**
+ * This file is copied from
+ * https://github.com/fabric8io/kubernetes-client/blob/master/kubernetes-model/kubernetes-model-annotator/src/main/java/io/fabric8/kubernetes/annotator/KubernetesTypeAnnotator.java
+ *
+ * A small modification to this class is needed to work with EnMasse types and a newer version of the generator.
```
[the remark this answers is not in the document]
lulf: Agreed, will add that.

pkg/controller/messaginginfra/broker/controller.go
ctron: No, however you already have the same function twice, and I know at least two more locations with similar code in the IoT parts … wouldn't it make sense to have a single function?

pkg/controller/messaginginfra/broker/controller.go
ctron: Ah ok, I see now. Maybe adding a few comments would have helped understanding :)

Makefile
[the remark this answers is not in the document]
ctron: Yes, but if it doesn't exists. Wouldn't it execute `go get` every time again? My idea was to leverage the functionality of make, to create a local version of the binary (using `go get`) and then use it.

pkg/controller/messaginginfra/cert/internal.go
```diff
@@ -0,0 +1,192 @@
...
+		return nil, errors.New("Failed to parse PEM private key")
+	}
+
+	key, err := x509.ParsePKCS1PrivateKey(block.Bytes)
+	return key, err
+}
+
+func encodePemKey(key *rsa.PrivateKey) []byte {
+	certPrivateKeyPem := new(bytes.Buffer)
+	pem.Encode(certPrivateKeyPem, &pem.Block{
```
k-wall, line 177: Uncaught error

k-wall:
@lulf looks good

lulf:

pkg/controller/messaginginfra/cert/controller.go
```diff
@@ -0,0 +1,230 @@
...
+			}
+			return c.applyCaSecret(secret, logger)
+		})
+		return err
+	})
+}
+
+func (c *CertController) applyCaSecret(secret *corev1.Secret, logger logr.Logger) error {
+	install.ApplyDefaultLabels(&secret.ObjectMeta, "", secret.Name)
+	if secret.ObjectMeta.Annotations == nil {
```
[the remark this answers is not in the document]
lulf: Addressed in next PR.

pkg/controller/messaginginfra/cert/controller.go
```diff
@@ -0,0 +1,230 @@
...
+		secret.Data["tls.key"] = key
+		secret.Data["tls.crt"] = cert
+		secret.Data["ca.crt"] = caSecret.Data["tls.crt"]
+		secret.Data["keystore.p12"] = keystore
+		secret.Annotations[ANNOTATION_CA_DIGEST] = actualCaDigest
+
+	} else if shouldRenewCert(now, expiryDate) || expectedCaDigest != actualCaDigest {
+		// If we should renew or if CA has changed, renew this certificate
+
+		logger.Info("Renewing component certificate", "secret", secret.Name)
```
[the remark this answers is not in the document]
lulf: Addressed in next PR.

pkg/controller/messaginginfra/cert/internal.go
lulf: Addressed in next PR.

PR #4187, opened 2020-03-26
Pf update

drcoolsanjeev:
### Type of change

- Bugfix
- Enhancement / new feature
- Refactoring

### Description
- update PF versions
- fix the compilation issue after the upgrade
- added a dummy toolbar with updated PF code in ASToolbar.tsx


console/console-init/ui/src/modules/address-space/AddressSpacePage.tsx
```diff
@@ -22,6 +22,7 @@ import { compareObject } from "utils";
 import { useStoreContext, types, MODAL_TYPES } from "context-state-reducer";
 import { getHeaderForDeleteDialog, getDetailForDeleteDialog } from "./utils";
 import { TablePagination } from "components";
+import { ASToolbar } from "./containers/AddressSpaceFilterContainer/ASToolbar";
```
anukritijha: As per the convention, I think that we agreed on not using relative paths

console/console-init/ui/src/modules/address-space/containers/AddressSpaceFilterContainer/ASToolbar.tsx
```diff
@@ -0,0 +1,251 @@
...
+  );
+  const [nameSelected, setNameSelected] = useState<string>();
+  const [nameInput, setNameInput] = useState<string>();
+  const [namespaceSelectIsExpanded, setNamespaceSelectIsExpanded] = useState<
+    boolean
+  >(false);
+  const [namespaceSelected, setNamespaceSelected] = useState<string>();
+  const [namespaceInput, setNamespaceInput] = useState<string>();
+  const [nameOptions, setNameOptions] = useState<any[]>([]);
+  const [namespaceOptions, setNamespaceOptions] = useState<any[]>([]);
```
anukritijha: Please avoid using any.

console/console-init/ui/src/modules/address-space/AddressSpacePage.tsx
drcoolsanjeev: @anukritijha , this is just a test file for donald meeting. I have to remove this file in future.

ajaypratap003: we can use this path. this is current path not relative path and we need to avoid relative path i.e. ../../ etc.

console/console-init/ui/src/modules/address-space/containers/AddressSpaceFilterContainer/ASToolbar.tsx
drcoolsanjeev: It is just a temporary file

anukritijha: I didn't realize that. my bad.

PR #4233, opened 2020-04-01
Fix YAML parsing

ctron:
### Type of change

- Bugfix

### Checklist

- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/iot/iot-jdbc-base/src/main/java/io/enmasse/iot/jdbc/store/StatementConfiguration.java
+++ b/iot/iot-jdbc-base/src/main/java/io/enmasse/iot/jdbc/store/StatementConfiguration.java
@@ -60,8 +60,11 @@ public class StatementConfiguration {
         }
 
         final Yaml yaml = createYamlParser();
-        @SuppressWarnings("unchecked")
-        final Map<String, Object> properties = yaml.loadAs(input, Map.class);
+        /*
+         * we must load using "load(input)" not "loadAs(input, Map.class)", because the
+         * latter would require to construct a class (Map) by name, which we do not support.
+         */
+        final Map<String, Object> properties = yaml.load(input);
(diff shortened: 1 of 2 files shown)


ctron:

iot/iot-jdbc-base/src/main/java/io/enmasse/iot/jdbc/store/StatementConfiguration.java
```diff
@@ -159,12 +159,7 @@ public void dump(final Logger logger) {
      * @return A new YAML parser, never returns {@code null}.
      */
     static Yaml createYamlParser() {
-        return new Yaml(new Constructor() {
-            @Override
-            protected Class<?> getClassForName(String name) throws ClassNotFoundException {
-                throw new IllegalArgumentException("Class instantiation is disabled");
-            }
-        });
+        return new Yaml(new Constructor(Map.class));
```
k-wall, line 162: This doesn't prevent the !!class being constructed.

ctron:

iot/iot-jdbc-base/src/main/java/io/enmasse/iot/jdbc/store/StatementConfiguration.java
ctron: It did, but as you mentioned, only on the root level. I switched to loading the YAML via `load` (vs `loadAs(Map.class)`.

With `loadAs(Map.class)` the problem is, that the constructor still tries to construct an new object, by class name, for Map.

`load` on the other hand doesn't do that, as it returns basic types only, not mapping anything to a bean. So it will only return a `Map<String,Object>`, and that allows to disable all class loading.

I also added you suggested test, which triggered the issue.

PR #4247, opened 2020-04-03
Delegate authentication work to thread pool, improve exception handling

k-wall:
### Type of change

- Bugfix

### Description

Use Vertx worker threads to process authication work in parallel.
Ensure that all error paths through SaslAuthenticator cause the completion handler to called `true`.

### Checklist

- [ ] Update documentation
- [ ] Update CHANGELOG.md


k-wall:
> Looks good, but is worker = true implying that vertx proton will use worker threads instead of event threads?

Yes, I went with essentially the same setWorker(true).setInstances() fix as we applied to the api-server.  I did look at delegating the work of the SASLAuthicator to Vertx.vertx().executeBlocking(). I think I see the way, but I'd need to spend longer reasoning about the change, and I'd be less confident to see back ported to 0.31.1.

I need to apply the same change to the none authservice.  I think this belongs in the same change.

k-wall:

keycloak-plugin/src/main/configuration/standalone-h2.xml
```diff
@@ -588,7 +590,7 @@
             <inet-address value="${jboss.bind.address.management:127.0.0.1}"/>
         </interface>
         <interface name="public">
-            <inet-address value="${jboss.bind.address:127.0.0.1}"/>
+            <inet-address value="${jboss.bind.address:127.f.0.1}"/>
```
lulf, line 593: Typo?

k-wall, line 593:
```suggestion
            <inet-address value="${jboss.bind.address:127.0.0.1}"/>
```

k-wall:

PR #4249, opened 2020-04-03
WIP - systemtests - cli clients wait for receivers to attach

famartinrh:
### Type of change

- Bugfix

### Description

This PR tries to make messaging client tests more stable, currently there a sleep in some tests where receivers are started before senders, the changes in this PR add an active wait sending test messages to ensure the receivers are attached before running senders and start with the actual process of the test.
This is still WIP because proton.python tests are having some issues where the new functionalities implemented doesn't work...

### Checklist

- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/systemtests/src/main/java/io/enmasse/systemtest/executor/Exec.java
+++ b/systemtests/src/main/java/io/enmasse/systemtest/executor/Exec.java
@@ -4,7 +4,6 @@
  */
 package io.enmasse.systemtest.executor;
 
-import io.enmasse.systemtest.Environment;
 import io.enmasse.systemtest.logs.CustomLogger;
 import org.slf4j.Logger;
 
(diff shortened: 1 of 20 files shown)


k-wall:
I'm not sure I fully understand your approach.   Outwardly it makes the tests even more complicated by adding a lot of code. I'm not sure this is really what we want.   Keeping the tests as easy to reason about as possible should be our code.  Maybe you could elaborate more on your approach.

My approach to the problem would be to change the message clients.  Give them a new flag which allows them to signal when they have a link attached.  Just write a flag to stdout.  Change the ExternalMessagingClient to expose a Future that the tests can await.  The stream reader just needs to detect the flag and set the Future.    That way pub/sub tests can start the subscriber, wait on the Future until it is attached, then start the publisher.

famartinrh:
I understand your point @k-wall and agree with the approach, thanks for your feedback, a better solution for this is provided in this PR #4268

PR #4251, opened 2020-04-03
Disable framework test that requires Kubernetes to run

lulf:
This test fails if the machine building the code is not running Kubernetes, which I think framework tests should require.

@ctron I'm just disabling it for now.


kornys:
@lulf  you should use `@Disable` or add comment TODO so we will not forget to handle that in future

PR #4256, opened 2020-04-04
Add state management of routers and brokers

lulf:
Add new module for doing state management of router and broker. The
different reconiliation loops notify the management API of the desired
state.

The state manager synchronizes its state to the routers and brokers,
at present in the context of the messaging infra reconcile loop.

In this first phase, the state manager creates a management connection
to all routers using a non-TLS port, until the AMQP client supports
SASL EXTERNAL. The state manager creates router - broker connections
between all known pods.

A PKCS12 truststore is created for the broker in order for it to be
able to trust the router client certs.

Add another condition to the messaging infra indicating whether all
router - broker connections have been made. This can be used to verify
that the messaginginfra is deployed and running as expected.
Store and query router status

Add a finalizer to the messaginginfra which is used to shut down the
internal state management when a messaginginfra is deleted.


lulf:

kornys:
@lulf could you please rebase on current master and rewrite tests to current new model? Or I can do that

lulf:

lulf:

lulf:

lulf:

pkg/controller/messaginginfra/broker/controller.go
```diff
@@ -102,15 +109,27 @@ func (b *BrokerController) ReconcileBrokers(ctx context.Context, logger logr.Log
...
 				}
+				delete(hosts, toHost(&brokers.Items[i]))
 				toDelete--
 			}
 		} else {
 			logger.Info("No changes to broker pool")
 		}
+
+		// Update discoverable brokers
+		newHosts := make([]string, 0)
```
k-wall, line 120: you've called the variable newHosts, but it also contains the existing hosts.  Should it be called allHosts?

PR #4263, opened 2020-04-06
Add proposal for plan functionality to shared infra proposal

lulf:


documentation/design/proposals/shared-infrastructure.adoc
```diff
@@ -469,6 +469,126 @@ The following controllers and components must be implemented:
...
+```
+
+```
+apiVersion: enmasse.io/v1
+kind: MessagingTenant
+metadata:
+  name: default
+  namespace: myspace
+spec:
+  plan:
```
k-wall, line 526: how about calling plan planRef, borrowing from the way kuberentes references secrets.

documentation/design/proposals/shared-infrastructure.adoc
```diff
@@ -469,6 +469,126 @@ The following controllers and components must be implemented:
...
 * (Phase 2) Messaging-connector controller - Managing external connectors of namespaces
 
+=== Plan behavior
+
+Both for MessagingTenantPlan and AddressPlans - Only MessagingTenantPlan used in examples here.
+
+1. No plan - No plans defined by service admin - no plan specified by messaging tenant => no plan applied => no limits
+2. Restricted - Plans defined by service admin - plan set by messaging tenant
+3. Restricted by default - Plans defined and default defined by service admin - plan not set by messaging tenant
+
```
k-wall, line 479: I'm not sure I follow the idea behind "restricted" and "restricted by default".  Are these two modes of operation?  Can you elaborate?

documentation/design/proposals/shared-infrastructure.adoc
```diff
@@ -469,6 +469,126 @@ The following controllers and components must be implemented:
...
+    name: plan2
+    namespace: enmasse-plans
+```
+
+Alternative 2: After initial reconcile, MessagingTenant is not updated.
+
+NOTE:
+* Alternative 1: If default changes from `plan2` to `plan1`, nothing changes
+* Alternative 2: If default changes from `plan2` to `plan1`, change all plans without anything set
+* Both alternatives: If `plan2` is modified, changes are applied to all messaging tenants
```
k-wall, line 590: +1 to alternative 2.  

I think alternative 1 brings with it an inconsistency.  If you were to create a MessageTenant before the any MessagingTenantPlans existed, it would implicitly reference the notional 'unlimited' plan. The unlimited plan would be nameless so there would be nothing to materialise on the MessageTenant resource .    If you then subsequently added a MessagingTenantPlan and made it default this would leave the system in an inconsistent state.  Presumably it would have to update MessagingTenant and apply the new default plan.  This behaviour would break the "If default changes ..." rule you stated above.

documentation/design/proposals/shared-infrastructure.adoc
```diff
@@ -496,41 +495,11 @@ metadata:
...
-  namespace: myspace
-spec:
-  plan:
-    name: plan1
-    namespace: enmasse-plans
-```
+NOTE:
 
-If plan1 is modified, changes are applied to all messaging tenants.
+* If default changes from implicit default to plan 1 to `plan1`, change all tenants without anything set
```
k-wall, line 500:
```suggestion
* If default changes from implicit default to `plan1`, change all tenants without anything set
```

PR #4266, opened 2020-04-06
Add broker.globalMaxSize to the infraconfig CRDs

k-wall:
### Type of change

- Improvement

### Description

Field broker.globalMaxSize is missing from the standardinfraconfig and brokeredinfraconfig CRDs.
Applies regex that has same effect of that of ARTEMIS-873.

### Checklist

- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 
 ## 0.31.1
 * #4241: Fix a race in upgrader logic that prevented upgrade from continuing
+* #4266: Add broker.globalMaxSize to the infraconfig CRDs 
 * #4269: bump Netty dependency from netty-4.1.45.Final to netty-4.1.48.Final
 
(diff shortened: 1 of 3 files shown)


k-wall:
It would have been nicer if the CRD was in terms of the SI unit, then internally converted to the bytes value.  Having the mixture of the informal KB, Mb, b in the same CRD as the SI values on the resource limits is distasteful.

kornys:
do we need update CSV for OLM?

k-wall:
> do we need update CSV for OLM?

Yes, but the OLM bundle is derived `templates/crds` too so there is nothing extra to do at this point.  Images supporting OLM will need to be released.

k-wall:

k-wall:

k-wall:

PR #4268, opened 2020-04-06
systemtests - cli clients tests - wait for receivers to attach

famartinrh:
### Type of change

- Bugfix

### Description

This PR tries to make messaging client tests more stable, currently there are tests where receivers are started before senders and this is a source of failures in some environments...
To address that systemtests now detects when the receivers are attached reading the trace log from the clients, note some clients we use do not support any kind of trace log that we can use for this purpose.

a sleep in some tests where receivers are started before senders, 

### Checklist

- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/systemtests/src/main/java/io/enmasse/systemtest/messagingclients/AbstractClient.java
+++ b/systemtests/src/main/java/io/enmasse/systemtest/messagingclients/AbstractClient.java
@@ -46,7 +46,7 @@ public abstract class AbstractClient {
     private List<String> executable;
     private String podName;
     private String podNamespace;
-    private CompletableFuture<Void> clientAttached;
+    private CompletableFuture<Void> linkAttached;
     /**
      * Important: this is not any container_id nor nothing related with amqp, this is just an identifier for logging in our tests
      */
@@ -75,8 +75,8 @@ public abstract class AbstractClient {
         this.podName = SystemtestsKubernetesApps.getMessagingAppPodName(this.podNamespace);
         this.fillAllowedArgs();
         this.executable = transformExecutableCommand(ClientType.getCommand(clientType));
-        if (clientAttachedProbeFactory() != null) {
-            this.clientAttached = new CompletableFuture<Void>();
+        if (linkAttachedProbeFactory() != null) {
+            this.linkAttached = new CompletableFuture<Void>();
         }
     }
 
@@ -163,8 +163,8 @@ public abstract class AbstractClient {
         return executor.getStdErr();
     }
 
-    public Future<Void> getClientAttached() {
-        return clientAttached;
+    public Future<Void> getLinkAttached() {
+        return linkAttached;
     }
 
     /**
@@ -209,8 +209,8 @@ public abstract class AbstractClient {
         messages.clear();
         try {
             executor = new Exec(logPath);
-            if (clientAttachedProbeFactory() != null) {
-                setClientAttachedProbe();
+            if (linkAttachedProbeFactory() != null) {
+                setLinkAttachedProbe();
             }
             int ret = executor.exec(prepareCommand(), timeout);
             synchronized (lock) {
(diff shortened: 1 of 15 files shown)


famartinrh:

systemtests/src/main/java/io/enmasse/systemtest/messagingclients/AbstractClient.java
```diff
@@ -42,20 +46,21 @@
     private List<String> executable;
     private String podName;
     private String podNamespace;
+    private CompletableFuture<Void> clientAttached;
```
k-wall, line 49: Naming - it is not the client that is attached, but link, so I think linkAttached would be better.

systemtests/src/main/java/io/enmasse/systemtest/platform/apps/SystemtestsKubernetesApps.java
```diff
@@ -919,6 +918,7 @@ private static Deployment getMessagingAppDeploymentResource() {
                 .withImage("quay.io/enmasse/systemtests-clients:latest")
                 .withCommand("sleep")
                 .withArgs("infinity")
+                .withEnv(new EnvVarBuilder().withName("PN_TRACE_FRM").withValue("true").build())
```
k-wall, line 921: It makes me nervous the build on the protocol trace for this feature.  My main concerns:

1. It is not the protocol trace's intended purpose
1. AMQP uses links for other purposes (such as transaction coordination http://docs.oasis-open.org/amqp/core/v1.0/os/amqp-core-transactions-v1.0-os.html), you run the risk of set the Future prematurely owing to the wrong attach being seen.
1. The protocol trace format may change.

That said, may be this is pragmatic step towards a solution we want.  We could always refactor the messaging clients later, one by one, adding 'proper' support then switch it out in the client implementations.







 AMQP uses links for other features - for instance, transaction in AMQP are coordinated via the transaction coordinator

systemtests/src/test/java/io/enmasse/systemtest/bases/clients/ClientTestBase.java
```diff
@@ -170,7 +171,19 @@ protected void doRoundRobinReceiverTest(AbstractClient sender, AbstractClient re
         Future<Boolean> recResult = receiverClient1.runAsync();
         Future<Boolean> rec2Result = receiverClient2.runAsync();
 
-        Thread.sleep(15000);
+        Thread.sleep(2);
```
k-wall, line 174: Does the 2 millisecond sleep actually contribute anything :)

systemtests/src/test/java/io/enmasse/systemtest/bases/clients/ClientTestBase.java
```diff
@@ -170,7 +171,19 @@ protected void doRoundRobinReceiverTest(AbstractClient sender, AbstractClient re
...
+
+        if (receiverClient1.getClientAttachedProbe() != null) {
```
k-wall, line 176: Couldn't you just make `getClientAttachedProbe() ` return a default implementation whose get() call simply sleeps for the duration of the timeout argument.  That'd get rid of the conditions.

k-wall:
Overall comment - much less scary! thanks @famartinrh

systemtests/src/main/java/io/enmasse/systemtest/messagingclients/AbstractClient.java
famartinrh: changed

systemtests/src/main/java/io/enmasse/systemtest/platform/apps/SystemtestsKubernetesApps.java
famartinrh: Agree on that, atm this is the best approach we can have given with the functionalities the clients currently have

systemtests/src/test/java/io/enmasse/systemtest/bases/clients/ClientTestBase.java
famartinrh: I like the idea, done

famartinrh:

famartinrh:

PR #4272, opened 2020-04-07
Add retry for createAddress kube request

kornys:


kornys:
@lulf do you think it is good?

kornys:
@lulf it is createOrReplace so I think it is fine

PR #4273, opened 2020-04-07
Change docker installation

obabec:
### Type of change

- Enhancement / new feature

### Description

I want try to change our docker installation script to official docker action.

### Checklist

- [ ] Update documentation
- [ ] Update CHANGELOG.md


obabec:
@lulf I have found that, you should add update to each action where you are installing something so we should try to add it to docker installation and see how it's going because this is not official docker action either and I think, they will be reaching same issue as we did with wrong installation path :)

PR #4276, opened 2020-04-07
Broker support access

k-wall:
### Type of change

- Enhancement / new feature

### Description

For the support use case, a support user should be able to connect to the broker(s) of an address space and run standard artemis support commands ` address show` and query JMX via Jolokia.
### Checklist

- [x] Update documentation
- [x] Update CHANGELOG.md

### DIFF
--- a/address-space-controller/src/main/java/io/enmasse/controller/TemplateInfraResourceFactory.java
+++ b/address-space-controller/src/main/java/io/enmasse/controller/TemplateInfraResourceFactory.java
@@ -14,12 +14,7 @@ import java.io.File;
 import java.io.IOException;
 import java.io.UncheckedIOException;
 import java.nio.file.Files;
-import java.util.ArrayList;
-import java.util.Base64;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Optional;
+import java.util.*;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -44,6 +39,7 @@ import io.fabric8.kubernetes.api.model.HasMetadata;
 import io.fabric8.kubernetes.api.model.PersistentVolumeClaim;
 import io.fabric8.kubernetes.api.model.PodTemplateSpec;
 import io.fabric8.kubernetes.api.model.SecretReference;
+import io.fabric8.kubernetes.api.model.Secret;
 import io.fabric8.kubernetes.api.model.apps.Deployment;
 import io.fabric8.kubernetes.api.model.apps.StatefulSet;
 
@@ -100,6 +96,26 @@ public class TemplateInfraResourceFactory implements InfraResourceFactory {
             }
         }
         parameters.put(TemplateParameter.MESSAGING_SECRET, serviceCertMapping.get("messaging").getSecretName());
+
+        Optional<Secret> secret = kubernetes.getSecret("broker-support-" + infraUuid);
+        if (secret.isPresent()) {
+            Map<String, String> secretData  = secret.get().getData();
+            if (secretData.containsKey("username") && secretData.containsKey("password")) {
+                parameters.put(TemplateParameter.BROKER_SUPPORT_USER, secret.get().getData().get("username"));
+                parameters.put(TemplateParameter.BROKER_SUPPORT_PWD, secret.get().getData().get("password"));
+            } else {
+                createSupportCredentials(parameters);
+            }
+        } else {
+            createSupportCredentials(parameters);
+        }
+    }
+
+    private void createSupportCredentials(Map<String, String> parameters) {
+        String brokerSupportUser =  String.format("broker-support-%s", UUID.randomUUID());
+        String brokerSupportPwd = UUID.randomUUID().toString();
+        parameters.put(TemplateParameter.BROKER_SUPPORT_USER, Base64.getEncoder().encodeToString(brokerSupportUser.getBytes()));
+        parameters.put(TemplateParameter.BROKER_SUPPORT_PWD, Base64.getEncoder().encodeToString(brokerSupportPwd.getBytes()));
     }
 
     private void prepareMqttParameters(AddressSpace addressSpace, Map<String, String> parameters) {
@@ -314,6 +330,5 @@ public class TemplateInfraResourceFactory implements InfraResourceFactory {
             throw new IllegalArgumentException("Unknown address space type " + addressSpace.getSpec().getType());
         }
     }
-
 
 }
--- a/address-space-controller/src/main/resources/templates/brokered-space-infra.yaml
+++ b/address-space-controller/src/main/resources/templates/brokered-space-infra.yaml
@@ -16,6 +16,21 @@ objects:
       infraUuid: ${INFRA_UUID}
       infraType: brokered
 - apiVersion: v1
+  kind: Secret
+  metadata:
+    name: broker-support-${INFRA_UUID}
+    annotations:
+      addressSpace: ${ADDRESS_SPACE}
+    labels:
+      app: enmasse
+      infraType: standard
+      infraUuid: ${INFRA_UUID}
+      role: support-credentials
+  type: Opaque
+  data:
+    username: ${BROKER_SUPPORT_USER}
+    password: ${BROKER_SUPPORT_PWD}
+- apiVersion: v1
   kind: PersistentVolumeClaim
   metadata:
     annotations:
@@ -136,6 +151,9 @@ objects:
           - mountPath: /opt/apache-artemis/custom
             name: broker-custom
             readOnly: false
+          - mountPath: /opt/apache-artemis/support
+            name: broker-support
+            readOnly: true
         initContainers:
         - env:
           - name: INFRA_UUID
@@ -193,6 +211,9 @@ objects:
         - name: external-cert
           secret:
             secretName: ${MESSAGING_SECRET}
+        - name: broker-support
+          secret:
+            secretName: broker-support-${INFRA_UUID}
 - apiVersion: v1
   kind: Service
   metadata:
@@ -463,3 +484,7 @@ parameters:
 - description: Interval (in seconds) to use between controller resync
   name: CONTROLLER_RESYNC_INTERVAL
   value: '600'
+- description: Username for broker support access
+  name: BROKER_SUPPORT_USER
+- description: Password for broker support access
+  name: BROKER_SUPPORT_PWD
--- a/address-space-controller/src/main/resources/templates/standard-space-infra.yaml
+++ b/address-space-controller/src/main/resources/templates/standard-space-infra.yaml
@@ -197,6 +197,21 @@ objects:
       app: enmasse
       infraType: standard
       infraUuid: ${INFRA_UUID}
+- apiVersion: v1
+  kind: Secret
+  metadata:
+    name: broker-support-${INFRA_UUID}
+    annotations:
+      addressSpace: ${ADDRESS_SPACE}
+    labels:
+      app: enmasse
+      infraType: standard
+      infraUuid: ${INFRA_UUID}
+      role: support-credentials
+  type: Opaque
+  data:
+    username: ${BROKER_SUPPORT_USER}
+    password: ${BROKER_SUPPORT_PWD}
 - apiVersion: apps/v1
   kind: Deployment
   metadata:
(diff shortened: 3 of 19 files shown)


k-wall:
This is just a POC with hardcoded support account name and no password check.

Here's invocation of `artemis`:

```
oc exec broker-queuespace-ly4y-0 -- /opt/apache-artemis/bin/artemis address show --user support
SLF4J: Class path contains multiple SLF4J bindings.
SLF4J: Found binding in [jar:file:/opt/apache-artemis/lib/broker-cli.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: Found binding in [jar:file:/opt/apache-artemis/lib/slf4j-jboss-logmanager-1.0.4.GA.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: See http://www.slf4j.org/codes.html#multiple_bindings for an explanation.
SLF4J: Actual binding is of type [ch.qos.logback.classic.util.ContextSelectorStaticBinder]
!!GLOBAL_DLQ
activemq.management
activemq.management.79227cac-a5ac-425c-8cff-6baed4a0da65
activemq.management.tmpreply.6a0f5889-4291-214a-beb3-2b4baf509047
activemq.management.tmpreply.c3efa509-06ee-534a-966c-1442f0924316
activemq.notifications
myqueue1

```

And an invocation of JMX over Jolokia:

`oc exec broker-queuespace-ly4y-0 -- curl --silent --insecure --user support:bar https://localhost:8161/console/jolokia/read/org.apache.activemq.artemis:broker=%22broker-queuespace-ly4y-0%22/AddressMemoryUsage`

k-wall:

k-wall:

k-wall:

k-wall:
@lulf @rgodfrey  for this work, I was considering making 8161 bind to 127.0.0.1 in order to prevent access to 8161 from the cluster. My motivation is to limit the places the broker support credentials are valid.  However, I notice we already create services for the Jolokia ports.   I am contemplating removing the services.  

```
broker-queuespace-1sb1           ClusterIP   None             <none>        8161/TCP                                                  98s
```

I think these are redundant (or maybe were always superfluous).

*The broker probes are made by with container command invocation `probe.sh` so there don't rely on the existence of the service.  In the service we name them `health-tls` making me think they were introduced for the probe use-case.
*We have no tests /documented use-cases that required these services to exist.

That said, I suppose it would be possible for the users to have built their own use-cases around the existence of these services.  This would be a breaking change for those users.

I am veering towards not leaving the services in place and keeping 8161 bound to all interfaces.   I wanted to give you chance to express an opinion.

lulf:
@k-wall As you describe, the services are probably not needed. I think they might have been at some point when we scraped metrics off the jolokia endpoint a long time ago.

address-space-controller/src/main/java/io/enmasse/controller/TemplateInfraResourceFactory.java
```diff
@@ -100,6 +100,7 @@ private void prepareParameters(AddressSpace addressSpace,
             }
         }
         parameters.put(TemplateParameter.MESSAGING_SECRET, serviceCertMapping.get("messaging").getSecretName());
+        parameters.put(TemplateParameter.BROKER_SUPPORT_PWD, Base64.getEncoder().encodeToString(infraUuid.getBytes()));
```
k-wall, line 103: We need to generate a random password.

address-space-controller/src/main/resources/templates/brokered-space-infra.yaml
```diff
@@ -15,6 +15,21 @@ objects:
...
+    annotations:
+      addressSpace: ${ADDRESS_SPACE}
+    labels:
+      app: enmasse
+      infraType: standard
+      infraUuid: ${INFRA_UUID}
+      addressspace:
+  type: Opaque
+  data:
+    username: YnJva2VyLXN1cHBvcnQ=
```
k-wall, line 31: I think we said we wanted the support user id to be:

* unpredictable, but obviously a 'support' user id so its use is obvious from the logs.
* unlikely to collide with user id from the user domain

How about: `broker-support-<random>`

address-space-controller/src/main/resources/templates/standard-space-infra.yaml
```diff
@@ -197,6 +197,21 @@ objects:
...
+  kind: Secret
+  metadata:
+    name: broker-support-${INFRA_UUID}
+    annotations:
+      addressSpace: ${ADDRESS_SPACE}
+    labels:
+      app: enmasse
+      infraType: standard
+      infraUuid: ${INFRA_UUID}
+      addressspace:
```
k-wall, line 210: addressspace field important?

address-space-controller/src/main/resources/templates/brokered-space-infra.yaml
```diff
@@ -15,6 +15,21 @@ objects:
...
+  kind: Secret
+  metadata:
+    name: broker-support-${INFRA_UUID}
+    annotations:
+      addressSpace: ${ADDRESS_SPACE}
+    labels:
+      app: enmasse
+      infraType: standard
+      infraUuid: ${INFRA_UUID}
+      addressspace:
```
k-wall, line 28: addressspace label important?

k-wall:

address-space-controller/src/main/resources/templates/standard-space-infra.yaml
```diff
@@ -207,10 +207,10 @@ objects:
       app: enmasse
       infraType: standard
       infraUuid: ${INFRA_UUID}
-      addressspace:
+      role: broker
```
k-wall, line 210: not sure if this is the right use of role.  Maybe role ought to be 'support-credentials'    @lulf do you have best practice advice to share?

address-space-controller/src/main/resources/templates/brokered-space-infra.yaml
```diff
@@ -193,6 +211,9 @@ objects:
         - name: external-cert
           secret:
             secretName: ${MESSAGING_SECRET}
+        - name: broker-support
```
k-wall, line 214: Wondering if we should mark this as optional?   We always expect the secret to be present, but breaking in the case where it is absent seems excessive.  WDYT?

broker-plugin/plugin/src/main/resources/bootstrap.xml
```diff
@@ -25,7 +25,7 @@
    <server configuration="${artemis.URI.instance}/etc/broker.xml"/>
 
    <!-- The web server is only bound to localhost by default -->
-   <web bind="https://0.0.0.0:8161"
+   <web bind="https://127.0.0.1:8161"
```
k-wall, line 28:
```suggestion
   <web bind="https://0.0.0.0:8161"
```
reverting back to bind all interfaces (see comments above).

broker-plugin/plugin/src/main/sh/launch-broker.sh
```diff
@@ -1,5 +1,9 @@
 #!/bin/sh
 
+// KWTODO tmp - refer to secret
+echo "support" > /tmp/support_user
+echo "support" > /tmp/support_password
+
```
k-wall, line 6:
```suggestion
```

systemtests/src/test/java/io/enmasse/systemtest/isolated/support/SupportToolingTest.java
```diff
@@ -0,0 +1,144 @@
...
+                    "show",
+                    "--user", supportUser,
+                    "--password", supportPassword
+            );
+            assertThat(artemisCmdResponse.getRetCode(), is(true));
+            List<String> addresses = Arrays.asList(artemisCmdResponse.getStdOut().split("\n"));
+            assertThat(addresses, hasItem(addr.getSpec().getAddress()));
+
+            if (AddressSpaceType.STANDARD.toString().equals(type)) {
+                // FIXME: can protect the brokered address space owing to #4295
```
k-wall, line 128:
```suggestion
                // FIXME: can't protect the brokered address space owing to #4295
```

address-space-controller/src/main/resources/templates/standard-space-infra.yaml
robshelly: I agree that it looks a bit add having `role: broker` here. However, applying the same labels as those applied to corresponding broker pods allows the correct secret to be easily found by the system test. 
https://github.com/EnMasseProject/enmasse/pull/4276/files#diff-b037b734d02edac5aba40f8fd25d5d1eR88

address-space-controller/src/main/java/io/enmasse/controller/TemplateInfraResourceFactory.java
```diff
@@ -100,6 +95,11 @@ private void prepareParameters(AddressSpace addressSpace,
             }
         }
         parameters.put(TemplateParameter.MESSAGING_SECRET, serviceCertMapping.get("messaging").getSecretName());
+
+        String brokerSupportUser =  String.format("broker-support-%s", UUID.randomUUID());
+        String brokerSupportPwd = UUID.randomUUID().toString();
```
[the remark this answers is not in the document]
k-wall: @robshelly I find prepareParameters could try to read the secret.  If it exists and username and password are populated pass those as args to the template.  otherwise, create the new defaults.

address-space-controller/src/main/resources/templates/standard-space-infra.yaml
k-wall: The content of CRs should be as meaningful and least surprise is possible. Changing the test to use a different set of labels is the right thing to do I think.

address-space-controller/src/main/resources/templates/brokered-space-infra.yaml
lulf: We could also take it a step further, to have the admin configure the secret with the credentials and point to it in the infra config for instance. I think it depends on how we think this feature would be used. For now, I'm thinking having it required is fine since its always set by the address-space-controller.

address-space-controller/src/main/resources/templates/standard-space-infra.yaml
lulf: Role is just a label we have defined to use for brokers, and not really a convention, just something that has always been that way. We don't even use the same label key for routers, so I'm fine just using 'role' here as well.

address-space-controller/src/main/java/io/enmasse/controller/TemplateInfraResourceFactory.java
```diff
@@ -96,10 +97,16 @@ private void prepareParameters(AddressSpace addressSpace,
...
-        parameters.put(TemplateParameter.BROKER_SUPPORT_USER, Base64.getEncoder().encodeToString(brokerSupportUser.getBytes()));
-        parameters.put(TemplateParameter.BROKER_SUPPORT_PWD, Base64.getEncoder().encodeToString(brokerSupportPwd.getBytes()));
+        Optional<Secret> secret = kubernetes.getSecret("broker-support-" + infraUuid);
+        if (!secret.isPresent()) {
+            String brokerSupportUser =  String.format("broker-support-%s", UUID.randomUUID());
+            String brokerSupportPwd = UUID.randomUUID().toString();
+            parameters.put(TemplateParameter.BROKER_SUPPORT_USER, Base64.getEncoder().encodeToString(brokerSupportUser.getBytes()));
+            parameters.put(TemplateParameter.BROKER_SUPPORT_PWD, Base64.getEncoder().encodeToString(brokerSupportPwd.getBytes()));
+        } else {
+            parameters.put(TemplateParameter.BROKER_SUPPORT_USER, secret.get().getData().get("username"));
```
k-wall, line 107: @robshelly what if the secret is present but username and/or password is absent?  I think the missing ones should be recreated.

robshelly:
@enmasse-ci run tests testcase=*SupportToolingTest test_only

address-space-controller/src/main/java/io/enmasse/controller/TemplateInfraResourceFactory.java
k-wall: lgtm

k-wall:
@enmasse-ci run tests testcase=*SupportToolingTest

k-wall:

PR #4277, opened 2020-04-08
Fix - Creating an address fails with message "address space plan '' not found"

drcoolsanjeev:
### Type of change

- Bugfix

### Description

- Handled null check for fetching queries to get address types and plans in create Address configuration view.
- Fix - #4279 
### Checklist

- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/console/console-init/ui/src/pages/AddressSpaceDetail/AddressList/AddressListFilterPage.tsx
+++ b/console/console-init/ui/src/pages/AddressSpaceDetail/AddressList/AddressListFilterPage.tsx
@@ -23,6 +23,7 @@ import useWindowDimensions from "components/common/WindowDimension";
 import { SortForMobileView } from "components/common/SortForMobileView";
 import { ISortBy } from "@patternfly/react-table";
 import { FetchPolicy } from "constants/constants";
+import { Loading } from "use-patternfly";
 
 interface AddressListFilterProps {
   filterValue: string | null;
@@ -65,7 +66,9 @@ export const AddressListFilterPage: React.FunctionComponent<AddressListFilterPro
   isPurgeAllDisabled
 }) => {
   const { name, namespace, type } = useParams();
-  const [addressSpacePlan, setAddressSpacePlan] = React.useState();
+  const [addressSpacePlan, setAddressSpacePlan] = React.useState<string | null>(
+    null
+  );
   const client = useApolloClient();
   const { width } = useWindowDimensions();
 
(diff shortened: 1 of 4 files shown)


drcoolsanjeev:
@enmasse-ci run tests testcase=*ConsoleTest

console/console-init/ui/src/pages/CreateAddress/CreateAddressDefinition.tsx
```diff
@@ -164,28 +178,26 @@ export const AddressDefinition: React.FunctionComponent<IAddressDefinition> = ({
...
-    return {
-      value: type.spec.displayName,
-      label: type.spec.displayName,
-      description: type.spec.shortDescription
-    };
-  });
-  if (typeOptions.length === 0) setTypeOptions(types);
-
+  if (addressSpaceType) {
+    const addressTypes = client.query<IAddressTypes>({
```
ajaypratap003, line 182: is it possible to wrap this  api code in function like getAddressType and call this function.

console/console-init/ui/src/pages/AddressSpaceDetail/AddressList/AddressListFilterPage.tsx
```diff
@@ -65,7 +66,9 @@ export const AddressListFilterPage: React.FunctionComponent<AddressListFilterPro
   isPurgeAllDisabled
 }) => {
   const { name, namespace, type } = useParams();
-  const [addressSpacePlan, setAddressSpacePlan] = React.useState();
+  const [addressSpacePlan, setAddressSpacePlan] = React.useState<string | null>(
```
ajaypratap003, line 69: why do we need null type with string. some times null and undefined cause the issue . avoid to use null and undefined as much as possible. in case you want to set state empty just use blank string i.e. "".

console/console-init/ui/src/pages/CreateAddress/CreateAddressDefinition.tsx
```diff
@@ -37,14 +37,14 @@ export interface IAddressDefinition {
   addressspaceName: string;
   namespace: string;
   addressName: string;
-  addressSpacePlan: string;
+  addressSpacePlan: string | null;
```
ajaypratap003, line 40: remove null type if possible.

console/console-init/ui/src/pages/AddressSpaceDetail/AddressList/AddressListFilterPage.tsx
drcoolsanjeev: server doesn't support empty string, we have pass the value as null. This is main fix for this PR.

PR #4284, opened 2020-04-08
Prevents react warnings being promoted to errors failing the build on CI

k-wall:
Investigating GIT ci build issue


k-wall:
@obabec running the build with env var CI=true I saw the same issue as GitHub CI.  This change resolves the issue locally by flipping the env var for the react build only.  Hoping this does the trick on CI too.

PR #4285, opened 2020-04-08
Update broker image used in tests

lulf:
ConnectorsTest have been failing several times failing to connect to the remote broker. Hoping a new broker will improve that, and maybe we should also consider to use an on-cluster client and turn of TLS.


famartinrh:
why should we consider turning of TLS?

lulf:
@famartinrh mainly because I was seeing ssl connect errors for the test client and it was a bit hard to know if it was a client, broker or openshift issue.

PR #4291, opened 2020-04-09
Only build the controller-gen when necessary

ctron:
Also: this removes the error output, when the "which" command
cannot find the controller-gen binary, which is not an error.

### Type of change

- Bugfix
- Enhancement / new feature
- Refactoring

### Checklist

- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,6 +26,7 @@ velocity.log.*
 
 # Local go binary directory
 /go-bin/
+/local/
 
 # Openshift local cluster files
 **/openshift.local.clusterup/
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-TOPDIR          := $(shell dirname $(realpath $(lastword $(MAKEFILE_LIST))))
+TOPDIR          := $(dir $(lastword $(MAKEFILE_LIST)))
 include $(TOPDIR)/Makefile.env.mk
 
 GO_DIRS = \
@@ -136,26 +136,38 @@ docu_check:
 docu_clean:
 	make -C documentation clean
 
-# Targets related to kubebuilder
-controller-gen:
-ifeq (, $(shell which controller-gen))
+#region Targets related to kubebuilder
+
+ifeq (, $(shell which controller-gen 2>/dev/null))
+
+LOCALBIN:=$(TOPDIR)/local/go/bin
+CONTROLLER_GEN:=$(abspath $(LOCALBIN)/controller-gen )
+controller-gen: $(CONTROLLER_GEN)
+
+$(CONTROLLER_GEN):
+	@mkdir -p "$(LOCALBIN)"
 	@{ \
 	set -e ;\
 	CONTROLLER_GEN_TMP_DIR=$$(mktemp -d) ;\
 	cd $$CONTROLLER_GEN_TMP_DIR ;\
 	go mod init tmp ;\
-	GOPATH=$(GOPATH) go get sigs.k8s.io/controller-tools/cmd/controller-gen@v0.2.5 ;\
+	GOPATH=$(abspath $(LOCALBIN)/../) go get sigs.k8s.io/controller-tools/cmd/controller-gen@v0.2.5 ;\
 	rm -rf $$CONTROLLER_GEN_TMP_DIR ;\
 	}
-CONTROLLER_GEN=$(GOPATH)/bin/controller-gen
+
 else
-CONTROLLER_GEN=$(shell which controller-gen)
+
+controller-gen:
+.PHONY: controller-gen
+CONTROLLER_GEN:=$(shell which controller-gen)
+
 endif
 
 manifests: controller-gen
 	$(CONTROLLER_GEN) crd paths=./pkg/apis/enmasse/v1beta2 output:dir=./templates/shared-infra
 
+#endregion
 
 .PHONY: test_go_vet test_go_plain build_go imageenv
 .PHONY: all $(GO_DIRS) $(DOCKER_TARGETS) $(DOCKER_DIRS) build_java test_go systemtests clean_java docu_html docu_check docu_clean templates
-.PHONY: controller-gen manifests
+.PHONY: manifests


k-wall:
This fails for me like so.  abspath appears to prepend the CWD.   CI is seeing a failure that looks the same:

```
go: creating new go.mod: module tmp
KWDEBUG result of abspath: /Users/kwall/go/src/github.com/enmasseproject/enmasse//Users/kwall/go/src/github.com/enmasseproject/enmasse/local/go/bin/..
go: finding sigs.k8s.io/controller-tools/cmd/controller-gen v0.2.5
go: finding sigs.k8s.io/controller-tools/cmd v0.2.5
/Users/kwall/go/src/github.com/enmasseproject/enmasse/local/go/bin/controller-gen crd paths=./pkg/apis/enmasse/v1beta2 output:dir=./templates/shared-infra
make: /Users/kwall/go/src/github.com/enmasseproject/enmasse/local/go/bin/controller-gen: No such file or directory
```

ctron:
@k-wall Should be fixed now. CI looks green as well.

.gitignore
```diff
@@ -26,6 +26,7 @@ velocity.log.*
 
 # Local go binary directory
 /go-bin/
+/local/
```
lulf, line 29: Why do we need 2 different folders for go binaries, shouldn't they be the same?

ctron: I am not sure if we still use "go-bin" … but we can always remove it in a new PR right away if we don't.

ctron:
Originally I did put the binary under the `target` folder. However, this gets cleaned quite often, and so the lengthy part of re-building the controller-gen gets triggered over and over again.

I don't really care where it is located, I think it should be located in the repository root, and put into gitignore. Also it should not be cleaned by a standard `make clean`.

I am not sure if hiding it is a good approach though.

lulf:
> Originally I did put the binary under the `target` folder. However, this gets cleaned quite often, and so the lengthy part of re-building the controller-gen gets triggered over and over again.
>

Downloading and installing controller-gen to PATH should fix that, is that not sufficient?

> 
> 
> I don't really care where it is located, I think it should be located in the repository root, and put into gitignore. Also it should not be cleaned by a standard `make clean`.
> 
> 
> I am not sure if hiding it is a good approach though.

Why is that? It seems to be done by a lot of IDEs etc.

Thinking a bit more about this, maybe we should go in the other direction and not download anything by default. 

We could have a script that you would need to run before building enmasse, that would install all build time dependencies rather than automatically downloading unverified third party code from public internet and execute it, which is not an ideal security practice. 

This makes it easier to audit that download script for the paranoid :)

ctron:
> > Originally I did put the binary under the `target` folder. However, this gets cleaned quite often, and so the lengthy part of re-building the controller-gen gets triggered over and over again.
> 
> Downloading and installing controller-gen to PATH should fix that, is that not sufficient?

Sure, but I don't want project specific dependencies in my PATH.

> 
> > I don't really care where it is located, I think it should be located in the repository root, and put into gitignore. Also it should not be cleaned by a standard `make clean`.
> > I am not sure if hiding it is a good approach though.
> 
> Why is that? It seems to be done by a lot of IDEs etc.

I think IDEs would either integrate with what you provide, or embed the functionality as part of the IDE. But not download, build and store a binary in some hidden place.

> 
> Thinking a bit more about this, maybe we should go in the other direction and not download anything by default.
> 
> We could have a script that you would need to run before building enmasse, that would install all build time dependencies rather than automatically downloading unverified third party code from public internet and execute it, which is not an ideal security practice.

Especially with go :)

> 
> Right now we have a mix of tools you need tool install before building and tools that are fetched automatically.
> 
> This makes it easier to audit that download script for the paranoid :)

So what do you think about a sub-folder "tools", which contains a script (Makefile) that populates the required tools? That would also solve the issue that a project specific dependency pollutes your system/user wide path.

lulf:
> > > Originally I did put the binary under the `target` folder. However, this gets cleaned quite often, and so the lengthy part of re-building the controller-gen gets triggered over and over again.
> 
> > 
> 
> > Downloading and installing controller-gen to PATH should fix that, is that not sufficient?
> 
> 
> 
> Sure, but I don't want project specific dependencies in my PATH.
> 
> 
> 
> > 
> 
> > > I don't really care where it is located, I think it should be located in the repository root, and put into gitignore. Also it should not be cleaned by a standard `make clean`.
> 
> > > I am not sure if hiding it is a good approach though.
> 
> > 
> 
> > Why is that? It seems to be done by a lot of IDEs etc.
> 
> 
> 
> I think IDEs would either integrate with what you provide, or embed the functionality as part of the IDE. But not download, build and store a binary in some hidden place.
> 
> 
> 
> > 
> 
> > Thinking a bit more about this, maybe we should go in the other direction and not download anything by default.
> 
> > 
> 
> > We could have a script that you would need to run before building enmasse, that would install all build time dependencies rather than automatically downloading unverified third party code from public internet and execute it, which is not an ideal security practice.
> 
> 
> 
> Especially with go :)
> 
> 
> 
> > 
> 
> > Right now we have a mix of tools you need tool install before building and tools that are fetched automatically.
> 
> > 
> 
> > This makes it easier to audit that download script for the paranoid :)
> 
> 
> 
> So what do you think about a sub-folder "tools", which contains a script (Makefile) that populates the required tools? That would also solve the issue that a project specific dependency pollutes your system/user wide path.
> 
> 

I think that is a decent compromise and easy to document. If we can make the install path somewhat configurable that would be a bonus. Should we move the existing tools in scripts and hack there as well?

ctron:
> I think that is a decent compromise and easy to document. If we can make the install path somewhat configurable that would be a bonus. Should we move the existing tools in scripts and hack there as well?

I would love the get rid of `hack` altogether. I have big hopes in your work around kubebuilder :grinning:  … we could use the common way of `PREFIX` … so `make install` would install this in the local EnMasse path, and `make PREFIX=/usr/local` would install it into `/usr/local/bin`.

lulf:
Created https://github.com/EnMasseProject/enmasse/issues/4326 to track the cleanup work

PR #4305, opened 2020-04-13
Inject OpenShift custom ca bundle into console pod

k-wall:
### Type of change

- Bugfix

### Description

Inject custom CA bundle into Console pod so that Console authentication functions even when the OpenShift cluster uses a custom CA.

### Checklist

- [ ] Update documentation
- [x] Check RBAC rights for Kubernetes / OpenShift roles
- [x] Update CHANGELOG.md

### DIFF
--- a/systemtests/src/main/java/io/enmasse/systemtest/platform/Kubernetes.java
+++ b/systemtests/src/main/java/io/enmasse/systemtest/platform/Kubernetes.java
@@ -25,6 +25,7 @@ import java.util.concurrent.TimeoutException;
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
 
+import io.fabric8.openshift.api.model.Route;
 import org.apache.commons.io.output.CloseShieldOutputStream;
 import org.slf4j.Logger;
 
@@ -972,7 +973,7 @@ public abstract class Kubernetes {
         } while (!unready.isEmpty() && budget.timeLeft() > 0);
 
         if (!unready.isEmpty()) {
-            fail(String.format(" %d pod(s) still unready", unready.size()));
+            fail(String.format(" %d pod(s) still unready in namespace : %s", unready.size(), namespace));
         }
     }
 
@@ -1072,4 +1073,9 @@ public abstract class Kubernetes {
         log.info("Output: {}", stdoutString);
         return stdoutString;
     }
+
+    public List<Route> listRoutes(String namespace, Map<String, String> labels) {
+        throw new UnsupportedOperationException();
+    }
+
 }
--- a/systemtests/src/main/java/io/enmasse/systemtest/platform/Minikube.java
+++ b/systemtests/src/main/java/io/enmasse/systemtest/platform/Minikube.java
@@ -15,16 +15,12 @@ import io.fabric8.kubernetes.client.Config;
 import io.fabric8.kubernetes.client.ConfigBuilder;
 import io.fabric8.kubernetes.client.DefaultKubernetesClient;
 import io.fabric8.kubernetes.client.utils.HttpClientUtils;
-import io.fabric8.openshift.api.model.Route;
-import io.fabric8.openshift.client.OpenShiftClient;
 import okhttp3.OkHttpClient;
 import okhttp3.Protocol;
 import org.slf4j.Logger;
 
 import java.util.Arrays;
 import java.util.Collections;
-import java.util.List;
-import java.util.Map;
 
 public class Minikube extends Kubernetes {
     private static Logger log = CustomLogger.getLogger();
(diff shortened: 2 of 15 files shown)


lulf:
Should we support a way of explicitly configuring the console CA in the CRD as well? It could be useful both on Kubernetes and as an alternative fallback.

k-wall:
> Should we support a way of explicitly configuring the console CA in the CRD as well? It could be useful both on Kubernetes and as an alternative fallback.

Yes. I think so.  I'm planning to add providerTrustBundle that references a configmap.  OpenShift will default its name and add the annotation (enmasse.io/ca-bundle-hash).  On Kubernetes, the user can provide the configmap if he wishes. In the configmap, key(s that match pattern *.crt will be assumed to contain a PEM(s).

The configmap will be optionally mounted by console-init.  It will aggregate all the trust into a single file and write to a location known to oauth-proxy.  oauth-proxy will read use this file for its provider trust.   There's a slight wrinkle with oauth-proxy - the OpenShift version requires a file, whereas the other one requires a cert directory.

systemtests/src/main/java/io/enmasse/systemtest/platform/Kubernetes.java
```diff
@@ -532,6 +534,10 @@ public Pod getPod(String name) {
         return client.persistentVolumeClaims().inNamespace(infraNamespace).withLabels(labels).list().getItems();
     }
 
+    public List<Route> listRoutes(String namespace, Map<String, String> labels) {
```
lulf, line 537: This should be in OpenShift.java I think and the default implementation should be throwing an exception maybe? (CC @kornys )

kornys: @lulf yup I agree

k-wall: Fixed.

k-wall:
@enmasse-ci run tests testcase=FirefoxConsoleTest crc

systemtests/src/test/java/io/enmasse/systemtest/bases/web/ConsoleTest.java
```diff
@@ -1419,6 +1424,134 @@ protected void doTestErrorDialog(AddressSpace addressSpace) throws Exception {
...
+        assertThat(console.isPresent(), is(true));
+
+        CertPair originalOauthCert = OpenSSLUtil.downloadCert(oauthRoute.get().getSpec().getHost(), 443);
+
+        String customCaCnName = "custom-ca";
+        String customIngressSecretName = "custom-ingress";
+        try (CertPair ca = OpenSSLUtil.createSelfSignedCert("/O=io.enmasse/CN=MyCA");
+             CertPair unsignedCluster = OpenSSLUtil.createSelfSignedCert("/O=io.enmasse//CN=MyCluster");
+             CertSigningRequest csr = OpenSSLUtil.createCsr(unsignedCluster);
+             CertPair cluster = OpenSSLUtil.signCrs(csr, Collections.singletonList("*.apps-crc.testing"), ca)
```
kornys, line 1453: this is really CRC specific, it cause issues when we will run it agains real cluster, better is use `Environment.getKubernetesDomain()`

systemtests/src/main/java/io/enmasse/systemtest/platform/Minikube.java
```diff
@@ -15,12 +15,16 @@
 import io.fabric8.kubernetes.client.ConfigBuilder;
 import io.fabric8.kubernetes.client.DefaultKubernetesClient;
 import io.fabric8.kubernetes.client.utils.HttpClientUtils;
+import io.fabric8.openshift.api.model.Route;
```
kornys, line 18: why we need imports in this class? I don't see any use in code

systemtests/src/test/java/io/enmasse/systemtest/bases/web/ConsoleTest.java
k-wall: Fixed. I've now run the test on a proper OpenShift 4 cluster.  It passes.

systemtests/src/main/java/io/enmasse/systemtest/platform/Minikube.java
k-wall: fixed

k-wall:
I've verified the changes on CRC, OpenShift 4 and checked for regressions on Minikube.

k-wall:
@famartinrh @kornys re. CertificateUtils.  I didn't know about this class when I wrote the new test.  I will refactor the test classes to use a common openssl utility class, but don't intend to do this until after 0.31.1.   Okay?

k-wall:
> > Should we support a way of explicitly configuring the console CA in the CRD as well? It could be useful both on Kubernetes and as an alternative fallback.
> 
> Yes. I think so. I'm planning to add providerTrustBundle that references a configmap. On OpenShift the controller will create the config, default its name and add the annotation (enmasse.io/ca-bundle-hash). On Kubernetes, the user can provide the configmap if he wishes, but it won't be mandatory. In the configmap, key(s that match pattern *.crt will be assumed to contain a PEM(s).
> 
> The configmap will be optionally mounted by console-init. It will aggregate all the trust into a single file and write to a location known to oauth-proxy. oauth-proxy will read use this file for its provider trust. There's a slight wrinkle with oauth-proxy - the OpenShift version requires a file, whereas the other one requires a cert directory.

@lulf  captured as a separate feature request. #4322

k-wall:
@enmasse-ci run tests  testcase=FirefoxConsoleTest  ocp4

PR #4309, opened 2020-04-15
fix-column1 for name/namespace column in mobile view

drcoolsanjeev:
### Type of change

- Bugfix

### Description
- FIx - The column Name\Namespace is displayed as column-1 in mobile view of Address Space List.
- Replace Configuring status with Pending status when the address and address space is not picked up by controller in the server side. As configuring status misleads the user.

### DIFF
--- a/console/console-init/ui/src/modules/address-space/utils/TableFormatter.tsx
+++ b/console/console-init/ui/src/modules/address-space/utils/TableFormatter.tsx
@@ -18,7 +18,7 @@ const getTableCells = (row: IAddressSpace) => {
       {
         header: "name",
         title: (
-          <>
+          <span>
             <Link
               to={`address-spaces/${row.nameSpace}/${row.name}/${row.type}/addresses`}
             >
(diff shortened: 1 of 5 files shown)


console/console-init/ui/src/modules/address-space/utils/TableFormatter.tsx
```diff
@@ -18,24 +18,24 @@ const getTableCells = (row: IAddressSpace) => {
       {
         header: "name",
         title: (
-          <>
+          <span>
```
ajaypratap003, line 21: is there any specific reason to add span tag.

console/console-init/ui/src/modules/address-space/utils/TableFormatter.tsx
```diff
@@ -18,24 +18,24 @@ const getTableCells = (row: IAddressSpace) => {
...
             {row.nameSpace}
-          </>
+          </span>
         )
       },
       {
         header: "type",
         title: (
-          <>
+          <span>
```
ajaypratap003, line 35: is there any specific reason to add span tag.

console/console-init/ui/src/utils/Formatter.tsx
```diff
@@ -24,9 +24,9 @@ const ConnectionProtocolFormat: React.FunctionComponent<ProtocolIcon> = ({
   encrypted
 }) => {
   return (
-    <>
+    <span>
```
ajaypratap003, line 27: is there any specific reason to add span tag.

drcoolsanjeev: Yes, we have to replace Fragment with some html tag to wrap the content in a single tag, so that nothing get breaks in mobile version

PR #4312, opened 2020-04-15
Rerun reconcile if requeued

lulf:
### Type of change

- Enhancement / new feature

### Description

This ensures that creation of address spaces doesnt have to wait for
the next recheck interval before infrastructure gets provisioned.

### Checklist

- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/address-space-controller/src/main/java/io/enmasse/controller/ControllerChain.java
+++ b/address-space-controller/src/main/java/io/enmasse/controller/ControllerChain.java
@@ -88,10 +88,8 @@ public class ControllerChain implements Watcher<AddressSpace> {
         List<AddressSpace> updatedResources = new ArrayList<>();
 
 
-        boolean requeue = true;
-        while (requeue) {
-
-            requeue = false;
+        boolean requeue = false;
+        do {
             for (final AddressSpace original : resources) {
 
                 AddressSpace addressSpace = new AddressSpaceBuilder(original).build();
@@ -145,7 +143,10 @@ public class ControllerChain implements Watcher<AddressSpace> {
                     log.warn("Exception in {} reconcileAll", controller, e);
                 }
             }
-        }
+            if (requeue) {
+                resources = new ArrayList<>(addressSpaceApi.listAllAddressSpaces());
+            }
+        } while(requeue);
     }
 
     static boolean hasAddressSpaceChanged(AddressSpace original, AddressSpace addressSpace) {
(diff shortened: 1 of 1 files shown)


address-space-controller/src/main/java/io/enmasse/controller/ControllerChain.java
```diff
@@ -87,59 +87,65 @@ public void onUpdate(List<AddressSpace> resources) throws Exception {
 
         List<AddressSpace> updatedResources = new ArrayList<>();
 
-        for (final AddressSpace original : resources) {
 
-            AddressSpace addressSpace = new AddressSpaceBuilder(original).build();
-            final String addressSpaceName = addressSpace.getMetadata().getNamespace() + ":" + addressSpace.getMetadata().getName();
+        boolean requeue = true;
+        while (requeue) {
```
ctron, line 92: I think a `do {} while()` loop would work better here.

address-space-controller/src/main/java/io/enmasse/controller/ControllerChain.java
```diff
@@ -87,59 +87,65 @@ public void onUpdate(List<AddressSpace> resources) throws Exception {
...
-                log.debug("Controller chain output: {}", addressSpace);
+                    log.debug("Controller chain output: {}", addressSpace);
 
-                if (hasAddressSpaceChanged(original, addressSpace)) {
-                    log.debug("Change detected. Executing update.");
-                    if (!this.addressSpaceApi.replaceAddressSpace(addressSpace)) {
-                        log.info("Unable to persist address space state: {}", addressSpaceName);
+                    if (hasAddressSpaceChanged(original, addressSpace)) {
+                        log.debug("Change detected. Executing update.");
+                        if (!this.addressSpaceApi.replaceAddressSpace(addressSpace)) {
```
ctron, line 125: I would assume the replace call will change the resource version. As the replace call uses `lockResourceVersion`, this will work the first time, but not a second time, without reloading the resource.

lulf: I didn't test if this change has an effect, but I think you're right. I think this change will anyway not work then, back to the drawing table...

lulf: I think I have a solution that would work now. If requeue is requested, re-list address spaces, which would get the latest resourceVersion of the address spaces. Given that requeues are infrequent, I think the extra 'hit' on the api server is ok.

ctron: :+1:

I only think that the logic should be described in some source code comment. Also I think the current comment block (about the requeue) is no longer 100% correct.

lulf:

PR #4333, opened 2020-04-16
Docs:update_example

hmaclean73:
### Type of change

- Documentation

Docs: Update Example

### Checklist

- [ ] Update documentation


hmaclean73:
There are two different versions of the example; one for upstream and one for downstream. The downstream has the additional line: standard_authentication_service_postgresql=False. So, I could remove the service_admin/example.inventory link and add the example inline, but would I then use the ifdef directive for the one additional line in the downstream docs?

PR #4336, opened 2020-04-16
Test operable enmasse after kube node drain

kornys:
### Type of change

- Enhancement / new feature

### Checklist


kornys:
@lulf do you think that this test should be in standard suite or it should run as part soak job?

kornys:
@lulf moved to soak suite

PR #4357, opened 2020-04-21
Shared infra defaults

kornys:
### Type of change

- Enhancement / new feature

### Description

Alow using annotations with default messaging infra/tenant for easier writing of tests

### Checklist

- [ ] Update documentation
- [ ] Update CHANGELOG.md


kornys:
@enmasse-ci run tests profile=isolated-shared-infra test_only crc

systemtests/src/test/java/io/enmasse/systemtest/sharedinfra/SharedInfraDefaultsTest.java
```diff
@@ -0,0 +1,56 @@
...
+import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.api.Test;
+
+import java.util.Map;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+
+@Tag(TestTag.ISOLATED_SHARED_INFRA)
+public class SharedInfraDefaultsTest extends TestBase implements ITestIsolatedSharedInfra {
```
lulf, line 27: We should really start to use a better name than SharedInfra. The fact that it is shared is kind of a coincidence, and in 1.0 it will be the only infra.

systemtests/src/test/java/io/enmasse/systemtest/sharedinfra/SharedInfraDefaultsTest.java
```diff
@@ -0,0 +1,56 @@
...
+
+    @Test
+    @DefaultMessagingInfra
+    public void testDefaultInfra() {
+        MessagingInfra infra = infraResourceManager.getDefaultInfra();
```
lulf, line 32: For this to be more useful, would it be an idea to be able to specify a signature like this:

```
public void testDefaultInfra(MessagingInfra default) {
}
```

And to have multiple defaults:

```
public void testFoo(MessagingInfra infra1, MessagingTenant tenant1) {}
```
?

kornys: @lulf junit test methods are not like a normal methods -> it does not have method signatures like standard methods, you can only pass what Extension context allows you. And if you need something more you want to use parametrized test, but it is not easy for use for all, My idea was provide the most basic stuff. I like the idea that I can write easy test without taking care about default resources.

systemtests/src/test/java/io/enmasse/systemtest/sharedinfra/SharedInfraDefaultsTest.java
kornys: Ok

systemtests/src/test/java/io/enmasse/systemtest/sharedinfra/SharedInfraDefaultsTest.java
lulf: Ok, so maybe not in the method signature... I was wondering if we could maybe use reflection + annotations to support the following:

```
public class MyTest {
   @ Default
   public MessagingInfra infra

   @ Test  
    public void mytest() {
      // infra is automatically created.
    }
}
```

kornys: so what about this solution? `@DefaultMessagingInfra` if you want just base and `DefaultMessagingInfra(infra = instance)` if you want your own specified.

kornys: but Im not sure it is posible in annotation :) let me investigate it

lulf: Sure :) Do you want to merge this now, or should we take the time to explore some options before we go all in on a specific approach?

PR #4359, opened 2020-04-21
broker init.sh - stops execution of a script if a command has an error

k-wall:
### Type of change

- Bugfix

### Description

Broker's init.sh should stop execution and return a non-zero return code to the execution environment if an initialisation command ends in error.

### Checklist

- [ ] Update documentation
- [x] Update CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 
 ## 0.31.2
 * #4305: Inject OpenShift generated custom CA trust bundle into console pod so that console authentication works when a custom CA is in use.
+* #4358: broker init.sh stop script execution of a script if a command has an error (#4359)
 
 ## 0.31.1
 * #4046: add network-only as fetch policy for all queries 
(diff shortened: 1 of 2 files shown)


k-wall:

kornys:
does it cause pod restart?

lulf:
@kornys yes it should

k-wall:
> does it cause pod restart?

indirectly, yes.  The init container will fail if any of the commands within the init script fail.  The pod's `restartPolicy` then applies.

kornys:
So I think it is fine because pod restart always resolve this issue

k-wall:
> So I think it is fine because pod restart always resolve this issue

Yes, that is my expectation and was my experience yesterday on your cluster.  That suggests in this case the the clock skew is small.  If the clock skew was large, with the new script option, the init container will fail (early) and the root cause (cert not valid) will be obvious from the logs.

kornys:
nice, so tests and system will be more stable

k-wall:

k-wall:

k-wall:
The ConnectorsTest failure seem unrelated to the change.

PR #4361, opened 2020-04-21
User stories 1.0

lulf:
Convenient readable link: https://github.com/lulf/enmasse/blob/user-stories-1.0/documentation/design/user-stories/shared-infra.adoc

### DIFF
--- a/documentation/design/user-stories/shared-infra.adoc
+++ b/documentation/design/user-stories/shared-infra.adoc
@@ -568,6 +568,11 @@ spec:
 
 Create the following `MessagingAddressPlan` to enforce limits for multiple address types:
 
+* Multiple address types can be defined in the same plan
+* If resource limit is not set - limit is not specified
+* If resource request is not set - it equals the limit
+* If resource is not set - no limits are specified
+
 ```
 apiVersion: enmasse.io/v1beta2
 kind: MessagingAddressPlan
@@ -585,6 +590,9 @@ spec:
         storage: 30Mi
   topic:
     resources:
+      limits:
+        memory: 5Mi
+        storage: 20Mi
 
 ```
 
--- a/documentation/design/user-stories/shared-infra.adoc
+++ b/documentation/design/user-stories/shared-infra.adoc
@@ -790,6 +790,7 @@ the number of router replicas based on the parameters `connections` and `message
 
 * If the observed value of these parameters go above the max, the number of router replicas is increased.
 * If the observed value of these parameters go below min, the number of router replicas is decreased.
+* Client connections will need to be terminated at scaling events in order for them to be rebalanced.
 
 ```
 apiVersion: enmasse.io/v1beta2
@@ -875,6 +876,8 @@ Once the second cluster is configured with tenants and addresses, the next steps
 ** Send messages to each corresponding address and tenant on the 1.x cluster
 * Sanity check migration application output
 * Switch traffic to 1.x cluster
+
+NOTE: This approach will currently only work for queues. For subscriptions, we may need a way to enable producing to a subscription for an admin user for instance.
 
 === CRUDLW an Identity provider that allows tenants to utilize of a namespace scoped users for the purpose of messaging authentication
 
--- a/documentation/design/user-stories/shared-infra.adoc
+++ b/documentation/design/user-stories/shared-infra.adoc
@@ -8,13 +8,13 @@ toc::[]
 == Tenant Admin
 
 === Enable messaging for a namespace by creating a messaging tenant
-=== Reference infra, plans and authentication services in different namespaces
+=== Reference infra, plans and identity services in different namespaces
 
 To enable messaging in a Kubernetes namespace, a `MessagingTenant` resource must be created by the tenant admin.
 
-* The `MessagingTenant` may reference a specific `MessagingInfra`, `MessagingPlan` and `AccessControlService`
+* The `MessagingTenant` may reference a specific `MessagingInfra`, `MessagingPlan` and `IdentityProviders`
 that should be used for the namespace.
-* If a plan, infra or accessControlService is not specified, a default will be selected based on the service admin configuration.
+* If a plan, infra or identity providers is not specified, a default will be selected based on the service admin configuration.
 * If the `namespace` field of the reference is omitted, the namespace of the operator is assumed.
 * The `MessagingTenant` must have a name set to `default`
 
@@ -34,9 +34,9 @@ spec:
     name: small
     namespace: plan-space
 
-  accessControlServiceRef:
-    name: corp-auth
-    namespace: auth-space
+  identityProviderSelector:
+    matchLabels:
+      type: uat
 ```
 
 Then apply this resource using the command line tools or the console. Once applied, the status
@@ -151,6 +151,8 @@ Create a `MessagingEndpoint` resource with one of the supported types.
 * Default protocols are defined in a `MessagingPlan` for a namespace. If the plan does not specify defaults, they will be:
 ** For `route` and `ingress`: AMQPS
 ** For `cluster`, `nodePort` and `loadBalancer`: AMQP,AMQPS
+* By default the endpoint allows authentications to `IdentityProviders` realised at the `MessageTenant`.  It
+  is possible to restrict the endpoint to a subset of these identity providers using a selector.
 
 ==== Resource definition
 
@@ -182,6 +184,13 @@ spec:
     external:
       cert: # Reuse structure from address space connector spec
       key: # Reuse structure from address space connector spec
+
+  # Optional: restricts the identity providers available at this endpoint.
+  identityProviderSelector:
+    matchLabels:
+      type: messaging
+  # Optional: restrict the secureOnlyMechanism
+  secureOnlyMechanism: []
 
   # Cluster type creates a on-cluster load balancer
   cluster: {}
@@ -216,13 +225,18 @@ status:
   - name: endpoint1-amqp
     port: 5672
     protocol: AMQP
+  tls:
+    selfsigned:
+      ca.pem: <PEM VALUE>
+  identityProviders:
+  - name:  myldap
+    namespace: infrastuff
+    secureOnlyMechanisms: PLAIN, OAUTHBEARER
+    subjectPrefix: enamsse.io:myldap:infrastuff:
   conditions:
   - type: FoundTenant
     status: "True"
     lastTransitionTime: 2020-......
-  tls:
-    selfsigned:
-      ca.pem: <PEM VALUE>
 ```
 
 ==== Configure an in-cluster endpoint with TLS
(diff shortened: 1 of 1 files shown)


documentation/design/user-stories/shared-infra.adoc
```diff
@@ -0,0 +1,277 @@
...
+```
+apiVersion: enmasse.io/v1beta2
+kind: MessagingTenant
+metadata:
+  name: default
+  namespace: myapp
+spec: {}
+---
+apiVersion: enmasse.io/v1beta2
+kind: MessagingAddress
```
kornys, line 14: @lulf so what should happen when I create MessagingAddress CR in namespace where MessagingTenant is not deployed? Does it create messaging tenant default in namespace, or it should fail?

lulf: @kornys I was thinking it would report in the address status that it cannot find a messaging tenant, and address will be in phase Pending (or Failed?).

kornys: @lulf maybe keep it in Pending phase with error message, and I expect that when I create MessagingTenant in namespace, address will be configured and CR switch to Active phase? WDYT?

kornys: or maybe would be nice if operator creates messaging tenant if does not exist

lulf: > @lulf maybe keep it in Pending phase with error message, and I expect that when I create MessagingTenant in namespace, address will be configured and CR switch to Active phase? WDYT?

Yes, that is what I had envisioned.

> or maybe would be nice if operator creates messaging tenant if does not exist

We discussed this when going through these user stories initially, and found that automatically creating the messaging tenant would make it a bit strange wrt. what happens when you remove all addresses. I.e. it would either have to remain or we would have to delete it when there are no address. The behavior gets a bit messy, so I think the cleanest way is to require it to be created, which shouldn't be much of a problem for the 'getting started' experience.

kornys: Hope it won't be confusing for users :)

documentation/design/user-stories/shared-infra.adoc
```diff
@@ -0,0 +1,277 @@
...
+    external:
+      # Can be input directly or using a secret
+      certificate:
+        value: PEM-encoded cert
+      key:
+        valueFromSecret:
+          name: "mysecret" # Name of secret containing certificate
+          key: "tls.key"
+    # Self-signed
+    selfsigned: {}
```
k-wall, line 244: in the case where many endpoints take the self signed option, would there be.a single CA shared by all the endpoints (like we have now)?  How would a user get the CA certificate?

documentation/design/user-stories/shared-infra.adoc
lulf: Yeah, I'm open to change behavior here based on what we see, but I think explicit creation is a better initial default than auto-create to avoid accidental confusion.

documentation/design/user-stories/shared-infra.adoc
lulf: Thats a good point. If we want to keep self-signed (which I'm open to discuss), we could create the initial CA when tenant is created and then use that CA to sign any endpoints that would need creation. It would mean we generate a potentially unused CA, but I feel like it would be cleaner than to auto-create CA when endpoint is created etc (running into the same lifecycle issues as for auto-creating MessagingTenant).

documentation/design/user-stories/shared-infra.adoc
```diff
@@ -0,0 +1,277 @@
...
+==== Configure an endpoint exposed as a route
+
+```
+apiVersion: enmasse.io/v1beta2
+kind: MessagingEndpoint
+metadata:
+  name: endpoint1
+  namespace: myapp
+spec:
+  route:
```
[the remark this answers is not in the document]
lulf: I think we should have sane defaults for the protocols based on the endpoint type, but in this case I agree it should not allow specifying protocols other than AMQPS.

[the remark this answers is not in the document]
k-wall: AMQP+TLS or AMQP-WS+TLS.

documentation/design/user-stories/shared-infra.adoc
k-wall: I'd lean towards keeping self signed.   I think it makes the getting started experience easier.  I like the of creating the initialCA when the tenant is created. The overhead of it is sufficiently small to not be of a concern.  I suppose the tls section of each endpoint that refers to the self-signed would expose a ca field.

documentation/design/user-stories/shared-infra.adoc
```diff
@@ -0,0 +1,277 @@
...
+apiVersion: enmasse.io/v1beta2
+kind: MessagingEndpoint
+metadata:
+  name: endpoint1
+  namespace: myapp
+spec:
+  # TLS configuration for endpoint
+  tls: {}
+
+  protocols:
```
k-wall, line 183: I think in the status section, some part of the object needs to map from protocols to port numbers.

lulf: Updated proposal with a sub-header on endpoint status section.

documentation/design/user-stories/shared-infra.adoc
lulf: Added to the status section example.

documentation/design/user-stories/shared-infra.adoc
```diff
@@ -0,0 +1,307 @@
...
+kind: MessagingEndpoint
+metadata:
+  name: endpoint1
+  namespace: myapp
+spec:
+  ...
+
+  # Required for endpoints of type 'route' and 'ingress', as these only support TLS
+  # For other types, specifying `tls` will ensure they expose traffic on TLS
+  tls:
```
k-wall, line 262: A tis endpoint should be configurable to want or need a client certificate.  We need configuration for this use case together with configuration for the trusted CA anchor(s)/intermediates.

documentation/design/user-stories/shared-infra.adoc
```diff
@@ -115,32 +80,28 @@ metadata:
...
+  phase: Pe
 ```
 
+---
+
 === Ability to configure DLQ, expiry, forwarders etc. per address
 
-Define a special system queue that gets created on all brokers:
+* Queues that should be configured with a DLQ or ExpiryQueue should reference an address with type `deadLetter`.
+* Forwarders maybe configured for the queue (`in` and `out`), subscription (`out`) or deadLetter(`out`)) types.
```
k-wall, line 102:
```suggestion
* Forwarders may be configured for the queue (`in` and `out`), subscription (`out`) or deadLetter(`out`)) types.
```

documentation/design/user-stories/shared-infra.adoc
```diff
@@ -115,32 +80,28 @@ metadata:
...
+* If the referenced DLQ does not exist, the status section of the address will provide an error message.
```
k-wall, line 103: Just an observation, but if you forward an expired messages, won't it spontaneously expire again at its new destination.

documentation/design/user-stories/shared-infra.adoc
```diff
@@ -182,17 +163,71 @@ metadata:
...
-
   protocols:
-    - AMQPS
-    - AMQP
+  - AMQP
+  - AMQPS
+  - AMQP-WS
+  - AMQP-WSS
 
+  # Optional: Override generated hostname for endpoint. Clients must have the AMQP OPEN hostname set to this value.
```
k-wall, line 172: Will this be ignored for endpoints except route and ingress?

documentation/design/user-stories/shared-infra.adoc
```diff
@@ -182,17 +163,71 @@ metadata:
...
+
+  # Route type maps to OpenShift Route resource
+  route:
+    # Allows controlling the TLS termination of this route. Only supported for AMQP-WSS protocol.
+    tlsTermination: passthrough | reencrypt
+
+status:
+  phase: Pending | Configuring | Active | Failed | Terminating
+  message: "Useful error message"
+
```
k-wall, line 211: should we add a type: like you have done for MessagingAddress?

documentation/design/user-stories/shared-infra.adoc
```diff
@@ -182,17 +163,71 @@ metadata:
...
+    # External allows providing certificate externally. CN/SANs are validated against host
+    external:
+      cert: # Reuse structure from address space connector spec
+      key: # Reuse structure from address space connector spec
+
+  # Cluster type creates a on-cluster load balancer
   cluster: {}
+
+  # NodePort type creates a nodePort service, exposing the application on an ephemeral port on all nodes.
+  nodePort: {}
```
k-wall, line 190: it ephemeral the right word? I think the port number will be random (from the point of view of the enduser), but stable for the lifetime of the MessagingEndpoint.   I think if the cluster gets restarted, the port number may change.  Is my understanding right?

documentation/design/user-stories/shared-infra.adoc
```diff
@@ -229,75 +280,297 @@ metadata:
...
-          key: "tls.key"
-    # Self-signed
-    selfsigned: {}
-    # OpenShift-generated
-    openshift: {}
+    caCert:
+      value: <Base64-encoded PEM>
+    clientCert:
+      valueFromSecret:
+        name: client-cert
```
k-wall, line 317: missing key

documentation/design/user-stories/shared-infra.adoc
```diff
@@ -229,75 +280,297 @@ metadata:
...
+Create the following `MessagingPlan` to enforce limits:
+
+```
+apiVersion: enmasse.io/v1beta2
+kind: MessagingPlan
+metadata:
+  name: plan1
+  namespace: planspace
+spec:
+  broker:
```
k-wall, line 535: should there be a restrict object here?

documentation/design/user-stories/shared-infra.adoc
```diff
@@ -229,75 +280,297 @@ metadata:
...
+```
+apiVersion: enmasse.io/v1beta2
+kind: MessagingPlan
+metadata:
+  name: myplan
+  namespace: planspace
+spec:
+  restrict:
+    endpoints:
+     - route: {}
```
k-wall, line 509: the semantics don't seem clear to me.

documentation/design/user-stories/shared-infra.adoc
lulf: It is useful for both loadbalancer and nodeport as well, because the router will expect that host for clients connecting. So if you want to setup messaging on messaging.example.com, you create an endpoint with that hostname, and then point your DNS record to the created loadbalancer.

For nodeport and local development, the host would be set to something like $(minikube ip) for clients to be able to connect outside the cluster.

For cluster service, it is less useful, but you could end up setting the AMQP host in your client and connect to the ip instead of the host, and it would work.

documentation/design/user-stories/shared-infra.adoc
lulf: Good point, we should note that in the docs then.

documentation/design/user-stories/shared-infra.adoc
lulf: I'm used to calling it the 'ephemeral port range' (https://en.wikipedia.org/wiki/Ephemeral_port), but I agree its perhaps not the best term to use. Maybe dynamic or private is a better term.

documentation/design/user-stories/shared-infra.adoc
lulf: Agreed, good idea.

documentation/design/user-stories/shared-infra.adoc
lulf: Thanks, will add it!

documentation/design/user-stories/shared-infra.adoc
rgodfrey: well, can queues be configured so that they effectively ignore expiry?  You'd want that generally for DLQs / expiry queues

documentation/design/user-stories/shared-infra.adoc
lulf: I kind of think of it like templates of endpoint specs that must be matched in order to be created. If any sub-path of the object does not match, then the endpoint will not be created.

documentation/design/user-stories/shared-infra.adoc
rgodfrey: I presume Ulf is using it in the sense of a "port in the ephemeral range" (although actually the default for K8s seems to be a subset of the default ephemeral range)

documentation/design/user-stories/shared-infra.adoc
lulf: I'm not sure, there are kind of conflicts of interest here between the desire to provide kubernetes-type resource specifications, which have both requests and limits, and restrictions which are kind of just restrictions.

Need to think some more about that.

documentation/design/user-stories/shared-infra.adoc
```diff
@@ -229,75 +280,297 @@ metadata:
...
+
+```
+apiVersion: enmasse.io/v1beta2
+kind: MessagingPlan
+metadata:
+  name: myplan
+  namespace: planspace
+spec:
+  restrict:
+    endpoints:
```
k-wall, line 508: I am wondering if there is a more expressive way to let the user encode restrictions.  It reminds me of schematron's niche  back in the XML world (Pangea :) ).

documentation/design/user-stories/shared-infra.adoc
k-wall: I'm not getting the analogy - do you podTemplates?

documentation/design/user-stories/shared-infra.adoc
```diff
@@ -0,0 +1,599 @@
...
+
+  ports:
+  - name: endpoint1-amqp
+    port: 5672
+    protocol: AMQP
+  conditions:
+  - type: FoundTenant
+    status: "True"
+    lastTransitionTime: 2020-......
+  tls:
```
k-wall, line 223: the TLS section should include the certificate's start and end date.  This will make it easy to display expiry info in the Console and gives a convenient location for a mechanism to monitor for certificate expiry.

documentation/design/user-stories/shared-infra.adoc
k-wall: Answering my own original question, no.  When artemis expires a message., internally it actually creates a copy of the message for the expiry queue, and set the copy's expiry to infinity.  The message's original expiry is set in a custom application property.  Not an AMQP compliant approach, but understandable one for an multi protocol broker.

org.apache.activemq.artemis.core.server.impl.QueueImpl#makeCopy(org.apache.activemq.artemis.core.server.MessageReference, boolean, boolean)

documentation/design/user-stories/shared-infra.adoc
lulf: Updated with OPA rules as examples

documentation/design/user-stories/shared-infra.adoc
```diff
@@ -744,13 +778,96 @@ spec:
...
+  router:
+    scalingStrategy:
+      dynamic:
+        # Specify the range of replicas this strategy can operate within
+        replicas:
+          min: 2
+          max: 5
+        # Strategy parameters that is used for determining if it should scale up/down
+        parameters:
+          # Number of connections _per router replica_
```
k-wall, line 809: How will this work?  Won't we need some mechanism to rebalance connections?  For instance I have connections min/max set to 10/1000 and a single router with 1001 connections.  The system spins up a second router which will have 0 connections.  A native algorithm could flip flop state, shutting down router 2 as it has too few and then spinning it up again.

documentation/design/user-stories/shared-infra.adoc
```diff
@@ -744,13 +778,96 @@ spec:
...
+Data migration:
+
+Once the second cluster is configured with tenants and addresses, the next steps are:
+
+* Create user on 0.x cluster with access to all address spaces and addresses
+* Create user on 1.0 cluster with access to all tenants and their addresses
+* Stop traffic to 0.x cluster
+* Run migration application with configured users. The application will:
+** Lookup output data from the address space and address configuration conversion scripts
+** Consume messages from each address in each address space on the 0.x cluster
```
k-wall, line 872: Only works for queue types.  Contents of subscriptions would  be lost, unless we made special upgrade rules that allow subscriptions to be routed to as if they were queues.

documentation/design/user-stories/shared-infra.adoc
```diff
@@ -744,13 +778,96 @@ spec:
...
+
+Configuration migration:
+
+* StandardInfraConfig/BrokeredInfraConfig -> MessagingInfra: Manual
+* A separate cluster must be used for the 1.0 cluster
+* AddressPlan -> MessagingAddressPlan: Manual
+* AddressSpacePlan -> MessagingPlan: Manual
+* AddressSpace -> MessagingTenant, MessagingConnector and MessagingEndpoint: Script-assisted
+* Address -> MessagingAddress: Script-assisted
+
```
k-wall, line 862: MessagingUser -> MessageUser, MessageRole, MessageRoleBinding: Script assisted.

k-wall:
Looks good - a few comments left.

documentation/design/user-stories/shared-infra.adoc
lulf: I think we would need to rebalance, otherwise scaling will not help as you state. I think if you have > 2 routers, we can do an approximation of that by doing a controlled rolling restart if we are above the minimum available required replicas. 

However, ideally I think the operator would terminate the existing connections on routers during a scaleup/down event.

documentation/design/user-stories/shared-infra.adoc
lulf: Good point. Maybe a subscription type addresses could have a property specifying if an incoming linkroute/autoLink in the route should be created for it for this purpose? Would it be safer/better to make this a global setting for all subscriptions, settable on the 'MessagingInfra'?

k-wall: If we were to do this, I don't think it should be a standard user-level feature.  It should be just some switch we can throw for the upgrade and then turn it off after the upgrade is complete.  Maybe an annotation?   OTOH maybe we make subscriptions out of scope?

documentation/design/user-stories/shared-infra.adoc
k-wall: IdeallyI think  we'd want some kind of heuristic that could take into account the cost of the connection (sessions, link, message size/throughput) and feed that into a rebalancing algorithm.  That sounds tricky.  

I think ultimately we will want to do something with intelligence, but we don't necessarily need to include in 1.0.

documentation/design/user-stories/shared-infra.adoc
lulf: If we can avoid caring about subscriptions, that would be my preference.

I agree it would be an annotation and not something we put into the schema. Maybe just an env variable to the operator.

PR #4362, opened 2020-04-21
Simplify state management for shared infrastructure

lulf:
Remove unnecessary management of tenant states and move logic to the
messagingtenant-controller.

Simplify the InfraState interface and rename it to InfraClient, and
reduce the scope of the state package to concern only synchronization
and querying of configuration for a particular infrastructure.

Move usage of infra client only to places where it is needed to apply
configuration to routers and brokers.


lulf:

PR #4376, opened 2020-04-23
Look into unstable connectors and forwarders test

lulf:


lulf:

systemtests/src/main/java/io/enmasse/systemtest/amqp/AmqpClientFactory.java
```diff
@@ -106,10 +109,16 @@ public AmqpClient createClient(TerminusFactory terminusFactory, ProtonQoS qos, A
             clientOptions.setSniServerName(messagingEndpoint.getHost());
         }
         log.info("External endpoint: " + clientEndpoint + ", internal: " + messagingEndpoint);
+        probeEndpoint(messagingEndpoint);
 
         return createClient(terminusFactory, clientEndpoint, clientOptions, qos);
     }
 
+    private void probeEndpoint(Endpoint messagingEndpoint) throws IOException {
+        Socket socket = new Socket(messagingEndpoint.getHost(), messagingEndpoint.getPort());
```
kornys, line 118: try finally? to be sure connection is closed?

k-wall, line 118: I don't think the timeout behaviour of Socket(messagingEndpoint.getHost(), messagingEndpoint.getPort()) is too clearly specified.  I'd suggest using the no-arg form and passing an explicit timeout.

lulf:

lulf:

lulf:

lulf:

lulf:

lulf:

lulf:
@enmasse-ci run tests profile=isolated testcase=ConnectorsTest

lulf:
@enmasse-ci run tests profile=isolated testcase=ForwardersTest

lulf:
@enmasse-ci run tests profile=isolated testcase=ConnectorsTest

lulf:

lulf:

lulf:
These tests seem to be stable when bypassing the openshift router, @kornys. Should we change the clients to external clients running on cluster?

kornys:
@lulf I'm not sure why we should change it to cmd external clients pod?

lulf:
@kornys they would use the cluster service endpoint instead of the route to connect, which seems more reliable. Since it can only be connected to by clients running on the cluster or when using oc cluster up for other clients, I think external clients would be the only option.

kornys:
hmm it can be challenge in case client use certificate, I think not all cli implementation can work with that

lulf:
@k-wall Yeah, that have always puzzled me about this as well. I don't have a good answer, but one thing that is different in these tests, are that they have this external broker and a route to that as well, but I'm not sure how that could have an impact.

lulf:
@kornys I'd like to merge this PR first without the connector/forwarder changes, and raise a separate PR for that, just in case I'm able to spot something more when this is run on master.

PR #4385, opened 2020-04-24
Initial implementation of address and endpoint API + controller

lulf:
Initial implementation of MessagingAddress type and controller. Support
reconciling anycast and queue addresses. Add first systemtests for the
latter.

Initial implementation of MessagingEndpoint type and controller. Support
for creating NodePort, ClusterIP and LoadBalancer service without TLS.

Extend state management logic to cover broker queue creation/deletion,
router address, autoLink and listener create/deletion.

I will add some more unit tests and comments before I consider this ready for merge, but I'm happy with the current implementation design. There are a lot of TODOs around unimplemented functionality, but I think those can be addressed in additional PRs.


lulf:

lulf:

systemtests/src/main/java/io/enmasse/systemtest/messaginginfra/resources/MessagingEndpointResourceType.java
```diff
@@ -0,0 +1,76 @@
+/*
+ * Copyright 2020, EnMasse authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+package io.enmasse.systemtest.messaginginfra.resources;
+
+import io.enmasse.address.model.CoreCrd;
+import io.enmasse.api.model.*;
```
kornys, line 8: start import :)

systemtests/src/main/java/io/enmasse/systemtest/messaginginfra/resources/MessagingAddressResourceType.java
```diff
@@ -0,0 +1,84 @@
+/*
+ * Copyright 2020, EnMasse authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+package io.enmasse.systemtest.messaginginfra.resources;
+
+import io.enmasse.address.model.CoreCrd;
+import io.enmasse.api.model.*;
```
kornys, line 8: star import

systemtests/src/main/java/io/enmasse/systemtest/messaginginfra/ResourceManager.java
```diff
@@ -13,10 +13,7 @@
 import io.enmasse.systemtest.amqp.AmqpClientFactory;
 import io.enmasse.systemtest.bases.ThrowableRunner;
 import io.enmasse.systemtest.logs.CustomLogger;
-import io.enmasse.systemtest.messaginginfra.resources.MessagingInfraResourceType;
-import io.enmasse.systemtest.messaginginfra.resources.MessagingTenantResourceType;
-import io.enmasse.systemtest.messaginginfra.resources.NamespaceResourceType;
-import io.enmasse.systemtest.messaginginfra.resources.ResourceType;
+import io.enmasse.systemtest.messaginginfra.resources.*;
```
kornys, line 16: star import

systemtests/src/test/java/io/enmasse/systemtest/sharedinfra/MessagingAddressTest.java
```diff
@@ -0,0 +1,90 @@
+/*
+ * Copyright 2020, EnMasse authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+package io.enmasse.systemtest.sharedinfra;
+
+import io.enmasse.api.model.*;
```
kornys, line 7: star import

lulf:
@enmasse-ci run tests profile=isolated-shared-infra crc

lulf:
@enmasse-ci run tests profile=isolated-shared-infra crc

lulf:

lulf:
@enmasse-ci run tests profile=isolated-shared-infra crc

lulf:
@enmasse-ci run tests profile=isolated-shared-infra crc

broker-plugin/plugin/src/main/resources/shared/broker.xml
```diff
@@ -73,9 +73,27 @@ under the License.
 
       <acceptors>
          <acceptor name="artemis">tcp://127.0.0.1:61616?tcpSendBufferSize=1048576;tcpReceiveBufferSize=1048576;protocols=CORE</acceptor>
+         <!-- TODO: Remove this once go client supports SASL EXTERNAL -->
```
k-wall, line 76: What purpose does the insecure AMQP port serve?

broker-plugin/plugin/src/main/resources/shared/cert-users.properties
```diff
@@ -0,0 +1 @@
+admin=/.*router-${INFRA_NAME}.*/
```
k-wall, line 1: Maybe surround ${INFRA_NAME} with \Q...\E

broker-plugin/plugin/src/main/resources/shared/login.config
```diff
@@ -0,0 +1,26 @@
...
+};
+
+CertLogin {
+   org.apache.activemq.artemis.spi.core.security.jaas.TextFileCertificateLoginModule required
+       debug=true
+       org.apache.activemq.jaas.textfiledn.user="cert-users.properties"
+       org.apache.activemq.jaas.textfiledn.role="cert-roles.properties";
+};
+
+activemq {
```
k-wall, line 15: redundant?

```suggestion
activemq {
```

k-wall, line 15: redundant block?

broker-plugin/plugin/src/main/resources/shared/broker.xml
k-wall: Ok, I see we are using for broker management - temporary step assumed.

pkg/state/broker.go
```diff
@@ -5,12 +5,149 @@
...
+	"time"
+
+	"github.com/enmasseproject/enmasse/pkg/amqpcommand"
+	v1beta2 "github.com/enmasseproject/enmasse/pkg/apis/enmasse/v1beta2"
+
+	"pack.ag/amqp"
+)
+
+const brokerCommandAddress = "activemq.management"
+const brokerCommandResponseAddress = "broker_command_response"
```
k-wall, line 21: The reply address needs to be prefixed by activemq.management otherwise management responses will be dropped if the broker becomes full.

pkg/state/infra.go
```diff
@@ -116,24 +195,319 @@ func (i *infraClient) SyncConnectors(routers []string, brokers []string) ([]Conn
...
+	if err != nil {
+		return err
+	}
+
+	err = i.applyBrokers(ctx, func(broker *BrokerState) error {
+		if address.Spec.Queue != nil {
+			// Configure relevant brokers
+			for _, brokerInfo := range address.Status.Brokers {
+				if brokerInfo.Host == broker.Host {
+					err := broker.EnsureQueue(tenantId, address)
```
k-wall, line 497: We are serialising queue creation, which I think is okay for now.  Later I think we'll need to be prepare to dispatch n createQueues and then await them in order to efficiently handle the case where there are many thousands of queues.

pkg/state/infra.go
```diff
@@ -116,24 +195,319 @@ func (i *infraClient) SyncConnectors(routers []string, brokers []string) ([]Conn
...
+				Waypoint:     true,
+			})
+			if err != nil {
+				return err
+			}
+
+			for _, broker := range address.Status.Brokers {
+				brokerState := i.brokers[broker.Host]
+				if brokerState == nil {
+					return fmt.Errorf("Unable to configure address autoLink (tenant %s address %s) for unknown broker %s", tenantId, address.GetAddress(), broker.Host)
```
k-wall, line 455: The Go advice on capitalisation at the start of error messages.  https://github.com/golang/go/wiki/CodeReviewComments#error-strings

pkg/state/infra.go
```diff
@@ -143,5 +517,122 @@ func (i *infraClient) Shutdown() error {
...
+			})
+			if err != nil {
+				return err
+			}
+		} else {
+			// TODO
+			return fmt.Errorf("%s protocol not yet supported", internalPort.Protocol)
+		}
+		i.freePort(internalPort.Port)
+	}
```
k-wall, line 545: what is responsible for cleaning the entry from the e.endpoint map?

pkg/state/infra.go
```diff
@@ -116,24 +195,319 @@ func (i *infraClient) SyncConnectors(routers []string, brokers []string) ([]Conn
...
 
-			connectorStatuses = append(connectorStatuses, *status)
+			connectorStatuses <- *status
+			return nil
+		})
+	})
+	if err != nil {
+		return nil, err
+	}
+	close(connectorStatuses)
```
k-wall, line 221: perhaps organise closure with defer to avoid a leak in the event of error.

pkg/state/infra.go
```diff
@@ -86,28 +139,54 @@ func (i *infraClient) updateBrokers(hosts []string) {
...
 
-	connectorStatuses := make([]ConnectorStatus, 0)
+	// Sync all known addresses and endpoints
+	err = i.syncConfiguration()
+	if err != nil {
+		return nil, err
+	}
 
+	totalConnectors := len(i.routers) * len(i.brokers)
+	connectorStatuses := make(chan ConnectorStatus, totalConnectors)
```
k-wall, line 184: why is the use of the channel helpful here?  I think everything is on the same thread

broker-plugin/plugin/src/main/resources/shared/broker.xml
lulf: Yes, this is until https://github.com/Azure/go-amqp/pull/5 is merged (or moved to some other source for this module).

broker-plugin/plugin/src/main/resources/shared/login.config
lulf: This is not need I think, I copied it from https://github.com/k-wall/artemis-dual-login-module/blob/master/etc/login.config :)

pkg/state/infra.go
lulf: Thanks, I'll add a comment about that. I think we should do some perf tests to test this bottle neck before spending much time to fix it though.

In this case, the BrokerState maintains a map of the current queues, so the management command to create the queue will only happen if the internal brokerState.queues map does not have an entry for the queue. This means that the management request will only fire on initial address creation or perhaps when migrating queues rather, as opposed to the fixed interval management queries that occur in standard-controller/agent today.

pkg/state/infra.go
lulf: Thats an oversight on my part. It should be in DeleteEndpoint (and same for DeleteAddress which is also missing the delete).

lulf:
@enmasse-ci run tests profile=isolated-shared-infra crc

lulf:
@enmasse-ci run tests profile=isolated-shared-infra crc

lulf:
I need to do some additional testing and fixes before this can be merged.

lulf:
@enmasse-ci run tests profile=isolated-shared-infra crc

lulf:
@enmasse-ci run tests profile=isolated-shared-infra crc

lulf:
@enmasse-ci run tests profile=isolated-shared-infra crc

lulf:
@enmasse-ci run tests profile=isolated-shared-infra crc

lulf:
@k-wall The testInfraRestart test fails due to the following conditions:

1. router state requests to read entity from router, enqueued in the commandClient
2. router restarts
3. commandClient reconnects and performs command
4. router state gets response from commandClient, but with an unexpected/outdated result, because it assumed this entity to exist

I think it would be nice if the 'router state' could detect that a reconnect had occured, so that it cleared its internal state of the router. I'm thinking about having the command client fail existing requests in the channel when it gets disconnected (and return a NotConnected error for those requests), so that the router state could reset itself. Would the console server be able to cope with those semantics? Are there an alternative approaches you think would work better?

k-wall:
> @k-wall The testInfraRestart test fails due to the following conditions:
> 
> 1. router state requests to read entity from router, enqueued in the commandClient
> 2. router restarts
> 3. commandClient reconnects and performs command
> 4. router state gets response from commandClient, but with an unexpected/outdated result, because it assumed this entity to exist
> 
> I think it would be nice if the 'router state' could detect that a reconnect had occured, so that it cleared its internal state of the router. I'm thinking about having the command client fail existing requests in the channel when it gets disconnected (and return a NotConnected error for those requests), so that the router state could reset itself.

I think this sounds reasonable.
> Would the console server be able to cope with those semantics? Are there an alternative approaches you think would work better?

Yes I think the console server would cope.  The command mechanism is currently used for queue purge only.  If the agent were disconnected at the point there was a command in the queue, the end user would see the 'not connected' message.  They would retry which would work.

> 
> EDIT: Or, maybe the proper thing is to modify commandClient to exit when disconnected, so that the router state can start it when it has detected the condition and reset its state (behavior would be configurable though)

I'm neutral - I think either model is reasonable, so I think I would take your first idea.

lulf:
@enmasse-ci run tests profile=isolated-shared-infra crc

lulf:
@enmasse-ci run tests profile=isolated-shared-infra crc

pkg/amqpcommand/command_client.go
```diff
@@ -76,8 +82,9 @@ func (c *CommandClient) Start() {
 				return
 			default:
 				err := c.doProcess()
+				atomic.StoreInt32(&c.connected, 0)
 				if err != nil {
-					atomic.StoreInt32(&c.connected, 0)
+					c.lastError = err
```
k-wall, line 87: Doesn't this need to be a atomic.StorePointer()?

lulf: Its only used by the same thread AFAICT, whereas connected is used by both 'client request' thread, and the 'processor' thread.

k-wall: So it is.  Comment withdrawn.

PR #4391, opened 2020-04-26
Proposal: authentication and authorization API for 1.0

k-wall:
Convenient link:

https://github.com/k-wall/enmasse/blob/authn-authz-10/documentation/design/proposals/authn_authz_for_1.0.adoc


documentation/design/proposals/authn_authz_for_1.0.adoc
```diff
@@ -0,0 +1,273 @@
...
+1. SASL-MECHANISMS performative flows the Client.
+1. Client selects mechanism
+1. SASL-INIT flows to the Access Control Server
+1. SASL-CHALLENGE / SASL RESPONSE as necessary
+1. Access Control Server calls out to the external authentication service / uses `MessagingUser`/`ServiceAccount` (as appropiate) to determine if the authentication is successful.
+1. SASL-OUTCOME flows to the Client.
+1. Access Control Server uses the `MessagingRole`, and `MessagingRoleBinding` to calculate the `ADDRESS-AUTHZ`.
+1. AMQP OPEN flows to the Router carrying the authorisation policy.
+
+When the messaging clients authenticate, the SASL username passed by the application must be qualified with a domain suffix (`@DOMAIN`). This allows the Access Control Server to disambiguate identities between providers.
```
lulf, line 59: Couldn't we also use the 'realm'/'saslInitHost' that is provided by the router auth plugin?

documentation/design/proposals/authn_authz_for_1.0.adoc
```diff
@@ -0,0 +1,273 @@
...
+  name: myuser
+  namespace: mynamespace
+spec:
+  password: {SCRAMSHA256}_saltedpassword_
+```
+
+==== ServiceAccount
+
+The Tenant Admin is able to create an ServiceAccount in the namespace.  Annotating the service account with a special
+annotation enables the user of the service account for messaging authentication.
```
lulf, line 192: System-wide permissions to read serviceaccounts seems a bit excessive is my first impression. Could we instead accept any token that authenticates a serviceaccount on the cluster, and let the RoleBinding decide what access is granted?

documentation/design/proposals/authn_authz_for_1.0.adoc
k-wall: This was may first idea, unfortunately it doesn't work.  `saslInit` is too late.  The server needs to provide the SASL mechanisms response before the client sends saslInit. In order to do this, the server needs to know what mechanisms to present. To do this, is needs to be able to resolve the authentication configuration.  I think using SNI is the only option.

documentation/design/proposals/authn_authz_for_1.0.adoc
k-wall: In our current system we require the user to nominate `ServiceAccounts` that are eligible for use for messaging authentication using the MessagingUser resources with the special type.  My intent was to keep an explicit step, but simplify it by having the `ServiceAccount` marked with an annotation. That does mean that the AccessControlServer needs cluster wide read permission for ServiceAccounts - firstly to decide if the service account is enabled, and secondly (a nice to have), validate the ServiceAccount references in the MessagingRoleBinding.

If having cluster wide ServiceAccount read is problematic, we can drop the idea of testing the whether the service account is enabled, and drop the validation between ServiceAccount and MessagingRoleBinding (which was probably going to be nothing more than a status message).

documentation/design/proposals/authn_authz_for_1.0.adoc
lulf: Given that its possible to omit it for service accounts and messaging users , I think that is sufficient.

documentation/design/proposals/authn_authz_for_1.0.adoc
lulf: I think I would prefer to make this configurable at least. Given that Kubernetes does not do any such validation of users referenced in a RoleBinding, I'm not sure if it is the role of the AccessControlServer to do this validation. Maybe it could make sense to delegate this kinf of 'extra' validation to an admission controller instead?

k-wall: I'll drop the annotation from the service account and validation.

documentation/design/proposals/authn_authz_for_1.0.adoc
```diff
@@ -0,0 +1,329 @@
...
+
+This identity provider will offer SASL mechanisms PLAIN, XOAUTH2 (Obsolete), link:https://tools.ietf.org/html/rfc7628[OAUTHBEARER]
+
+====== Anonymous Identity Provider 
+
+For the Anonymous Provider, the SASL mechnism ANONYMOUS will be used.
+
+===== Behaviour Without AccessControlService
+
+To support the simple getting started use case, if there are no `AccessControlService` defined in the infra namespace,
```
k-wall, line 113: I am not too conformable with this idea.   My main issue is that it violates "secure by default" advocated by OWASP and others.    It also brings with it a complicated switch in behaviour which might confuse the user's mental model and complicate the coding and testing.

I'm wondering if for the getting started user, we should accept that the user has to create an additional three objects:
1. MessagingUser or ServiceAccount
1. MessagingRole
1. MessagingRoleBinding

The content of these objects will be short and self explanatory.  The concepts should already resonate with the user so hopefully the hurdle is not too high.

documentation/design/proposals/authn_authz_for_1.0.adoc
```diff
@@ -0,0 +1,329 @@
...
+metadata:
+  name: myapps1-permissions
+  namespace: mynamespace
+subjects:
+- kind: MessagingUser
+  name: myuser  # refers to a MessagingUser
+- kind: ServiceAccount
+  name: myapp  # refers to a ServiceAccount  
+- kind: Subject
+  name: "foouser@ldap.mycompany.com". # refers to an external subject which must be domain qualified.
```
k-wall, line 301:
```suggestion
  name: "foouser@ldap.mycompany.com". # refers to an external subject which must be domain qualified.   This can refer to either user or group subjects.
```

documentation/design/proposals/authn_authz_for_1.0.adoc
```diff
@@ -0,0 +1,329 @@
...
+
+For authorization, authorization rules sourced from custom kubernetes resources will allow individual users, or
+groups of users to be permissioned for sending to or receiving from address(es) within the namespace.
+
+The Service Admin will be able to selectively enable/disable identity providers by configuring the Access Control Service
+resource. This will allow EnMasse to conform to the identity management rules of a particular organisation.  For instance, it will be possible for the Access Control Service to be configured so that only the external LDAP is available, or that
+service account and the external LDAP are available.
+
+Each Messaging Tenant will be associated with exactly one Access Control Service. This can be done locally by a Tenant
+Admin, or alternatively the Service Admin will be able to configure one or more Access Control Services centrally. Selectors will automatically associate an Access Control Service with the Tenants.
```
lulf, line 25: By automatically associating, how would this work with overlapping selectors? Should it be possible to explicitly configure the default AccessControlService in a `MessagingPlan`?

documentation/design/proposals/authn_authz_for_1.0.adoc
```diff
@@ -0,0 +1,329 @@
...
+Allows the use of an AMQP SASL authentication endpoint that implements the `authServicePlugin` protocol.  When this identity
+provider is used, the Access Control Server acts a SASL proxy.  It will connect to the external provider and relay the
+the AMQP SASL performatives to/from the client.
+
+The response from the External AMQP SASL server must provides an `authenticated-identity`.  It may also provide
+`groups` and may provide `address-authz`.  If the external server provides an `address-authz`, it will be ANDed with the `address-authz` calculated from the `MessagingRoles` and `MessagingRoleBindings`.
+
+The SASL mechanisms offered when this identity provider are in use depend on the mechansims required by the External AMQP SASL server.
+
+====== Internal Identity Provider
```
lulf, line 82: Would a better name be Kubernetes Identity Provider, Local Identity Provider or Cluster Identity Provider?

documentation/design/proposals/authn_authz_for_1.0.adoc
```diff
@@ -0,0 +1,329 @@
...
+For client auth, the identity is established during the TLS negotiation phase between Client and Router.  The router will
+pass the established identity through the `authServicePlugin` to the Access Control Server via the SASL EXTERNAL mechanism
+via the link:https://tools.ietf.org/html/rfc4422#appendix-A[authz-id-string].  (TODO: check Router behaviour).
+
+This identity provider will offer SASL mechanism EXTERNAL.
+
+====== Service Account Identity Provider
+
+For the Service Account provider, the client passes the Service Account's bearer token as a password and a special user-id
+`@@serviceaccount@@`. The Access Control Server will perform a `TokenReview` to check the token's validity and retrieve the user username.  A check is made to ensure that the username refers to a Service Account which is annotated with `enmasse.io/messaging-authentication`.
```
lulf, line 103: Is this check still valid?

documentation/design/proposals/authn_authz_for_1.0.adoc
```diff
@@ -0,0 +1,329 @@
...
+==== Qpid Dispatch Router
+
+This proposal reuses the existing `authServicePlugin` of the Qpid Dispatch Router unchnaged.
+
+=== Resources
+
+The new API will consist of the following types at the infrastructure level:
+
+* `AccessControlService`: used to define an instance of an accesss control service. The `AccessControlService` will define
+the parameters to establish connection with external identity provider (if any).  Selectors will be to configure to which namespaces it should be applied.
```
lulf, line 149: Applied or available?

documentation/design/proposals/authn_authz_for_1.0.adoc
```diff
@@ -0,0 +1,329 @@
...
+
+* Configuration provided by `MessagingInfra` will define the Access Control Server deployment.
+
+At the tenant level:
+
+* `MessagingTennant`: can be used to specify a specific `AccessControlService` to be used for the tenant.
```
lulf, line 155: Again, I wonder if this should be in the `MessagingPlan` or `MessagingTenant`, maybe both?

documentation/design/proposals/authn_authz_for_1.0.adoc
```diff
@@ -0,0 +1,329 @@
...
+
+* `MessagingUser` (v2beta1): used with the internal identity source to provide a messaging username/password pairing.  The password will be stored in the resource a secure fashion.  `MessagingUser` is scoped to the messaging tenant. It cannot
```
lulf, line 157: Should we use v1beta2 for all the 1.0 resources, with the goal of moving them to v1 once we are happy with it and ready to release?

documentation/design/proposals/authn_authz_for_1.0.adoc
```diff
@@ -0,0 +1,329 @@
...
+
+==== AccessControlService
+
+The `AccessControlService` used to define an instance of an accesss control service.  It allows the configuration the external
+identity providers and gives the ability to disable the internal identity providers.
+
+When the messaging client authenticates the domain part must be provided as part of the SASL username.  An exception
+is made for the client authentication provider and the anonymous provider.
+
+The `AccessControlService` has a selector for namespaces that should use this AccessControlService by default.  This selector
```
lulf, line 176: As I noted earlier, should it be the default, or just available? We need to make sure we have a consistent approach across MessagingPlan, MessagingAddressPlan and AccessControllService for this IMHO.

documentation/design/proposals/authn_authz_for_1.0.adoc
```diff
@@ -0,0 +1,329 @@
+# Proposal - Authentication and Authorization for 1.0
+
+## Overview
+
+1.0 will introduce a new Access Control Service.  This will replace the existing authentication services.
+It will have responsibility for identity management and authorization for the messaging use-case.
+
+For identity, it will have provision for:
+
+* leverage an external identity provider (such as a LDAP provider),
```
k-wall, line 10: External Keycloak?

lulf: Yes, I think that would be useful for migrating from pre-1.0 versions as well.

documentation/design/proposals/authn_authz_for_1.0.adoc
k-wall: For a getting started experience using anonymous auth, the minimum you'd need is a `AccessControlService` defined as below and a `MessagingRoleBinding`/`MessagingRole` granting `*` address permission to the anonymous subject:

```
anonymous:
       enabled: true
```

As for an out of the box default (with no AccessControlService), I think the system should behave as if the following AccessControlService were in place.

```
internal:
       enabled: true
serviceAccount:
       enabled: true
```

That would mean for a getting started experience with a username/password, the user would just need to create a `MessagingUser` (or `ServiceAccount`) and the `MessagingRoleBinding`/`MessagingRole`.

documentation/design/proposals/authn_authz_for_1.0.adoc
```diff
@@ -0,0 +1,329 @@
...
+1. Service Account `<username>@serviceaccount.<namespace>`
+1. Client Auth - certificate distinguished name
+1. Anonynous - `anonymous@namespace`
+
+=== Metrics
+
+The AccessControlService will expose a Prometheus endpoint will the following metrics:
+
+1. `auth_requests_total{domain="", sasl_outcome=""}`
+1. `auth_requests_duration_milliseconds{domain="", sasl_outcome="", le=<buckets>}`
```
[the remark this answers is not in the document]
k-wall: That's true. I'll change it.

documentation/design/proposals/authn_authz_for_1.0.adoc
k-wall: Yeah, I agree Internal is not a great name.   *Kubernetes Identity Provider* might make user think that it somehow uses a user concept from Kubernetes.  Cluster Identity Provider might have the user believe it is somehow cluster scoped.  Local Identity Provider works I think, or maybe NamespaceIdentityProvider?

documentation/design/proposals/authn_authz_for_1.0.adoc
k-wall: No, let me kill it.

documentation/design/proposals/authn_authz_for_1.0.adoc
k-wall: I've added Keycloak as a support external provider - GoCloak appears to have all the API we need to implement username/password auth (SCRAM-SHA SASL) and OAUTHBEARER in the same way we did in Java.

k-wall:
@lulf @rgodfrey  I am considering the following change:

1. AccessControlService -> AccessControlPlan.
1. ServiceAdmins define one or AccessControlPlan (prod, uat, sit1) etc.  Messaging tenants will *not* be able to define these objects.
1. AccessControlPlan keeps its selector - this is used to assign a default AccessControlPlan to the MessagingTenant. However, a MessagingTenant may override that default using an `accessControlPlanRef`
1. For the use-case where a MessagingTenant needs to override settings (say a host), the MessagingTenant object would have an inline `accessControlOverride` object (structure follows that of the plan), that allows the user to override the required value(s).
1. The ability for the tenant to override the plan or fields within the plan would be controllable from the AccessControlPlan.

How does that sound?

lulf:
> @lulf @rgodfrey I am considering the following change:
> 
>     1. AccessControlService -> AccessControlPlan.
> 

I kind of like the former better, but either are ok with me.

>     2. ServiceAdmins define one or more AccessControlPlans (prod, uat, sit1) etc.  Messaging tenants will _not_ be able to define these objects.
> 

I think they should still be possible to create these outside of the enmasse-infra namespace, as the 'service-operator' persona would need that ability. Or are you referring to the default RBAC roles?

>     3. AccessControlPlan keeps its selector - this is used to assign a default AccessControlPlan to the MessagingTenant. However, a MessagingTenant may override that default using an `accessControlPlanRef`
> 

I'm also thinking this is probably the way we should go with plans as well. However, what should be the default if you have overlapping selectors for AccessControlPlans?

>     4. For the use-case where a MessagingTenant needs to override settings (say a host), the MessagingTenant object would have an inline `accessControlOverride` object (structure follows that of the plan), that allows the user to override the required value(s).
> 

If the tenant, if granted permission by the system admin (making the tenant a service operator as well), have the ability to create their own AccessControlPlan, would that not cover this use case? Or is this to get finer grained overrides?

>     5. The ability for the tenant to override the plan or fields within the plan would be controllable from the AccessControlPlan.
> 
> 
> How does that sound?

k-wall:
> > @lulf @rgodfrey I am considering the following change:
> > ```
> > 1. AccessControlService -> AccessControlPlan.
> > ```
> 
> I kind of like the former better, but either are ok with me. Naming it Plan makes me think, should this have been part of the MessagingPlan?

I changed the name partly because the Service had the implication of the existence of a running entity that might consume resources.  I used the word Plan just to follow the convention from MessagingPlan and MessagingAddressPlan.
   
I don't think merging AccessControlPlan and MessagingPlan.  You might want to use MessagingPlan small in dev and sit, but to the two would need different access control configuration (dev ldap, sit ldap)..  

> 
> > ```
> > 2. ServiceAdmins define one or more AccessControlPlans (prod, uat, sit1) etc.  Messaging tenants will _not_ be able to define these objects.
> > ```
> 
> I think they should still be possible to create these outside of the enmasse-infra namespace, as the 'service-operator' persona would need that ability. Or are you referring to the default RBAC roles?

I meant RBAC. I don't mean to restrict AccessControlPlans to enmasse-infra namespace.    The TenantAdmin won't be able to create/update/delete these objects.

> 
> > ```
> > 3. AccessControlPlan keeps its selector - this is used to assign a default AccessControlPlan to the MessagingTenant. However, a MessagingTenant may override that default using an `accessControlPlanRef`
> > ```
> 
> I'm also thinking this is probably the way we should go with plans as well. However, what should be the default if you have overlapping selectors for AccessControlPlans?

I think we just define a rule that says if overlapping, the alphabetically first resource will be used so that the system is deterministic.  I'd prefer to do this to the alternative which would be to make the referrers invalid with an "ambiguous plan ref".

> 
> > ```
> > 4. For the use-case where a MessagingTenant needs to override settings (say a host), the MessagingTenant object would have an inline `accessControlOverride` object (structure follows that of the plan), that allows the user to override the required value(s).
> > ```
> 
> If the tenant, if granted permission by the system admin (making the tenant a service operator as well), have the ability to create their own AccessControlPlan, would that not cover this use case? Or is this to get finer grained overrides?

The steer I understood from @rgodfrey was that we wanted to give the Tenant Admin the ability to selectively override values (such as hosts).    That said, there if the  Tenant Admin had the RBAC ability, there is no reason why it should not work.

> 
> > ```
> > 5. The ability for the tenant to override the plan or fields within the plan would be controllable from the AccessControlPlan.
> > ```
> > 
> > 
> > How does that sound?

rgodfrey:
> > > @lulf @rgodfrey I am considering the following change:
> > > ```
> > > 1. AccessControlService -> AccessControlPlan.
> > > ```
> > 
> > 
> > I kind of like the former better, but either are ok with me. Naming it Plan makes me think, should this have been part of the MessagingPlan?
> 
> I changed the name partly because the Service had the implication of the existence of a running entity that might consume resources. I used the word Plan just to follow the convention from MessagingPlan and MessagingAddressPlan.
> 

I admit I have been constantly fighting this confusion every time we have discussed this :) - to me "service" definitely triggers the impression of a distinct process (or service) rather than just a set of configuration.  I'm not totally happy with plan, because it is more of a config... but I prefer it to service :-)

> I don't think merging AccessControlPlan and MessagingPlan would be good. You might want to use MessagingPlan small in dev and sit, but to the two would need different access control configuration (dev ldap, sit ldap)..

+1 

> 
> > > ```
> > > 2. ServiceAdmins define one or more AccessControlPlans (prod, uat, sit1) etc.  Messaging tenants will _not_ be able to define these objects.
> > > ```
> > 
> > 
> > I think they should still be possible to create these outside of the enmasse-infra namespace, as the 'service-operator' persona would need that ability. Or are you referring to the default RBAC roles?
> 
> I meant RBAC. I don't mean to restrict AccessControlPlans to enmasse-infra namespace. The TenantAdmin won't be able to create/update/delete these objects.
> 

+1 

> > > ```
> > > 3. AccessControlPlan keeps its selector - this is used to assign a default AccessControlPlan to the MessagingTenant. However, a MessagingTenant may override that default using an `accessControlPlanRef`
> > > ```
> > 
> > 
> > I'm also thinking this is probably the way we should go with plans as well. However, what should be the default if you have overlapping selectors for AccessControlPlans?
> 
> I think we just define a rule that says if overlapping, the alphabetically first resource will be used so that the system is deterministic. I'd prefer to do this to the alternative which would be to make the referrers invalid with an "ambiguous plan ref" message.

+1

> 
> > > ```
> > > 4. For the use-case where a MessagingTenant needs to override settings (say a host), the MessagingTenant object would have an inline `accessControlOverride` object (structure follows that of the plan), that allows the user to override the required value(s).
> > > ```
> > 
> > 
> > If the tenant, if granted permission by the system admin (making the tenant a service operator as well), have the ability to create their own AccessControlPlan, would that not cover this use case? Or is this to get finer grained overrides?
> 
> The steer I understood from @rgodfrey was that we wanted to give the Tenant Admin the ability to selectively override values (such as hosts). That said, there if the Tenant Admin had the RBAC ability, there is no reason why it should not work.

Yeah - the RBAC mechanism is fine.  The per-field override is only required because of the current way the auth services are defined

> 
> > > ```
> > > 5. The ability for the tenant to override the plan or fields within the plan would be controllable from the AccessControlPlan.
> > > ```
> > > 
> > > 
> > > How does that sound?

documentation/design/proposals/authn_authz_for_1.0.adoc
k-wall: Getting Started Default is now rewritten.

documentation/design/proposals/authn_authz_for_1.0.adoc
k-wall: This comment was made against the old design, I don't think it applies to the new.

documentation/design/proposals/authn_authz_for_1.0.adoc
k-wall: I went with NamespaceIdentityProvider.

documentation/design/proposals/authn_authz_for_1.0.adoc
k-wall: Comment refers to the old version.

documentation/design/proposals/authn_authz_for_1.0.adoc
k-wall: As discussed below, I think identity provider config is orthogonal to messaging plans.

PR #4394, opened 2020-04-27
Feature: Create Alert for DeviceList

rkpattnaik780:
### Type of change

- new feature

### Description

The alert component for` DeviceListPage` has been created along with unit tests.

### DIFF
--- a/console/console-init/ui/src/modules/device/components/DeviceListAlert/DeviceListAlert.test.tsx
+++ b/console/console-init/ui/src/modules/device/components/DeviceListAlert/DeviceListAlert.test.tsx
@@ -1,0 +1,27 @@
+/*
+ * Copyright 2020, EnMasse authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+
+import React from "react";
+import { render } from "@testing-library/react";
+import { DeviceListAlert, IDeviceListAlertProps } from "./DeviceListAlert";
+
+describe("<DeviceListAlert />", () => {
+  const props: IDeviceListAlertProps = {
+    visible: true,
+    variant: "info",
+    title: "Run filter to view your devices",
+    description: "You have a total of 36,300 devices"
+  };
+
+  it("should render an alert on top of DeviceListPage", () => {
+    const { getByText } = render(<DeviceListAlert {...props} />);
+
+    const alertTitle = getByText("Run filter to view your devices");
+    const alertDescription = getByText("You have a total of 36,300 devices");
+
+    expect(alertTitle).toBeInTheDocument();
+    expect(alertDescription).toBeInTheDocument();
+  });
+});
--- a/console/console-init/ui/src/modules/device/components/DeviceListAlert/DeviceListAlert.tsx
+++ b/console/console-init/ui/src/modules/device/components/DeviceListAlert/DeviceListAlert.tsx
@@ -1,0 +1,30 @@
+/*
+ * Copyright 2020, EnMasse authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+
+import React from "react";
+import { Alert, AlertProps } from "@patternfly/react-core";
+
+export interface IDeviceListAlertProps extends Pick<AlertProps, "variant"> {
+  visible: boolean;
+  title: string;
+  description: string;
+}
+
+export const DeviceListAlert: React.FunctionComponent<IDeviceListAlertProps> = ({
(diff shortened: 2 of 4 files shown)


console/console-init/ui/src/modules/device/index.ts
```diff
@@ -0,0 +1,6 @@
+/*
```
drcoolsanjeev, line 1: no need to export the component from 'device'

rkpattnaik780: Exporting it here just to keep it consistent with structure of other modules.

console/console-init/ui/src/modules/device/components/DeviceListAlert/DeviceListAlert.test.tsx
```diff
@@ -0,0 +1,26 @@
...
+      variant: "info",
+      title: "Run filter to view your devices",
+      description: "You have a total of 36,300 devices"
+    };
+    const { getByText } = render(<DeviceListAlert {...props} />);
+
+    const alertTitle = getByText("Run filter to view your devices");
+    const alertDescription = getByText("You have a total of 36,300 devices");
+
+    expect(alertTitle).toBeDefined();
```
ajaypratap003, line 23: try to use toBeInTheDocument function instead of  toBeDefined. toBeDefined can cuase in case null.
expect(null).toBeDefined(); //it will pass but logically it should fail.
so try to use toBeInTheDocument that will more safe.
 expect(null).toBeInTheDocument(); //fail
 expect(alertTitle).toBeInTheDocument(); //pass

console/console-init/ui/src/modules/device/components/DeviceListAlert/DeviceListAlert.test.tsx
```diff
@@ -0,0 +1,26 @@
+/*
+ * Copyright 2020, EnMasse authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+
+import React from "react";
+import { render } from "@testing-library/react";
+import { DeviceListAlert, IDeviceListAlertProps } from "./DeviceListAlert";
+
+describe("Alert for DeviceListPage", () => {
```
ajaypratap003, line 10: follow some standard to write unit test.
1. Add component at top level describe 
describe("<DeviceListAlert/>", () => {}
2. use It instead test as highly used in react app.
3. Start your test information in It/test with should 
it("should render an alert on top of  DeviceListPage", () => {}

Nothing is logical wrong in your test. just lets try to follow some standard.

console/console-init/ui/src/modules/device/components/DeviceListAlert/DeviceListAlert.tsx
```diff
@@ -0,0 +1,31 @@
+/*
+ * Copyright 2020, EnMasse authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+
+import React from "react";
+import { Alert } from "@patternfly/react-core";
+
+export interface IDeviceListAlertProps {
```
ajaypratap003, line 9: always try to use interface of  PF component to avoid define same properties again in our components.
Here you can extend PF Alert interface AlertProps and removed properties form IDeviceListAlertProps interface whic are available in PF interface. title and variant properties are available in PF interface AlertProps. so you can remove.
export interface IDeviceListAlertProps extends AlertProps {
}

console/console-init/ui/src/stories/DeviceList.stories.tsx
```diff
@@ -0,0 +1,28 @@
...
+import { MemoryRouter } from "react-router";
+
+import { DeviceListAlert } from "modules/device";
+import { text } from "@storybook/addon-knobs";
+
+export default {
+  title: "Device"
+};
+
+export const deviceListAlert = () => (
```
ajaypratap003, line 16: component name should be in Pascal case i.e. DeviceListAlert. If you are worries about same name as you have already your component. You can have other name i.e. DeviceAlert

rkpattnaik780:
> added some review comments

Thanks for your review @ajaypratap003. Knowing about these standards will certainly help me in improving my PRs in future. Do have a look at the updated PR. :)

console/console-init/ui/src/modules/device/components/DeviceListAlert/DeviceListAlert.test.tsx
rkpattnaik780: Currently the ui is using `it` and `test` interchangeably. We should address that in near future I think and stick to only `it`.

Issue #2938, opened 2019-06-26
okhttp3.internal.http2.Http2Stream java.net.SocketTimeoutException: timeout sporadically seen during system test runs


k-wall:
During system test runs on master, we see the occasional test failure that looks like this:

```
2019-06-25T09:12:24.747Z &amp#27;[1;31mERROR&amp#27;[0;39m [ITestSeparator] Caught exception
io.fabric8.kubernetes.client.KubernetesClientException: Operation: [list]  for kind: [Address]  with name: [null]  in namespace: [enmasse-infra]  failed.
	at io.fabric8.kubernetes.client.KubernetesClientException.launderThrowable(KubernetesClientException.java:64)
	at io.fabric8.kubernetes.client.KubernetesClientException.launderThrowable(KubernetesClientException.java:72)
	at io.fabric8.kubernetes.client.dsl.base.BaseOperation.listRequestHelper(BaseOperation.java:169)
	at io.fabric8.kubernetes.client.dsl.base.BaseOperation.list(BaseOperation.java:607)
	at io.fabric8.kubernetes.client.dsl.base.BaseOperation.list(BaseOperation.java:67)
	at io.enmasse.systemtest.utils.AddressUtils.waitForAddressesMatched(AddressUtils.java:201)
	at io.enmasse.systemtest.utils.AddressUtils.waitForDestinationsReady(AddressUtils.java:179)
	at io.enmasse.systemtest.selenium.page.ConsoleWebPage.createAddressWebConsole(ConsoleWebPage.java:675)
	at io.enmasse.systemtest.selenium.page.ConsoleWebPage.createAddressesWebConsole(ConsoleWebPage.java:626)
	at io.enmasse.systemtest.bases.web.WebConsoleTest.doTestFilterConnectionsByHostname(WebConsoleTest.java:537)
	at io.enmasse.systemtest.brokered.web.FirefoxWebConsoleTest.testFilterConnectionsByHostname(FirefoxWebConsoleTest.java:109)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
[...]

PR #4399, opened 2020-04-28
Retry fabric8 on connection error

kornys:


kornys:
@enmasse-ci run tests profile=shared crc test_only

kornys:
@enmasse-ci run tests profile=shared crc test_only

kornys:
@enmasse-ci run tests profile=shared crc test_only

kornys:
@enmasse-ci run tests profile=shared testcase=FirefoxConsoleTest crc test_only

kornys:
@enmasse-ci run tests profile=shared test_only crc

systemtests/src/main/java/io/enmasse/systemtest/platform/Minikube.java
```diff
@@ -34,10 +34,11 @@ protected Minikube(Environment environment) {
             OkHttpClient httpClient = HttpClientUtils.createHttpClient(config);
             // Workaround https://github.com/square/okhttp/issues/3146
             httpClient = httpClient.newBuilder()
-                    .protocols(Collections.singletonList(Protocol.HTTP_1_1))
+                    .protocols(Collections.singletonList(Protocol.HTTP_2))
```
k-wall: It was this change #2938 that switched to HTTP_1_1 to avoid a OkHttp client bug (https://github.com/square/okhttp/issues/3146).  The defect has a fixed milestone of v4 (and maybe 3.12.8 judging by comments in the thread).  We still pickup 3.12.6 transitively from fabric8.    If we want to switch to HTTP_2, then we probably need to override okhttp too to pickup 3.12.8

kornys: @k-wall so do you think that we can remove this workaround?

systemtests/src/main/java/io/enmasse/systemtest/platform/Minikube.java
```diff
@@ -34,10 +34,11 @@ protected Minikube(Environment environment) {
...
                     .connectTimeout(environment.getKubernetesApiConnectTimeout())
                     .writeTimeout(environment.getKubernetesApiWriteTimeout())
                     .readTimeout(environment.getKubernetesApiReadTimeout())
+                    .retryOnConnectionFailure(true)
```
k-wall, line 41: I'd have been dead against this change when we had the api-server, as I'd have worried it would have masked our own defects.

As the api-server is now gone, I am more  relaxed (slightly). If we got a problem connecting to Kubernetes API server, we'd likely class it as platform or environment so retrying would be the right thing to do.  OTOH  if we see are needing to reconnect often - that's kind of worrying - and this change may mask it.  Mmmm.  Decisions, decision!  @ctron @lulf

lulf: I think it could mask an issue where we were overloading the OpenShift API due to a bug in our operator (see recent bug/fix I made to address-space-controller, which caused various issues in tests using the API), so I would prefer if we didn't retry.

PR #4407, opened 2020-04-29
Make uninstall part less verbose

kornys:
When I read transcript log of systemtests I'm totally lost in all output of command when uninstall is running. So this pr make uninstall part less verbose for better readability.


kornys:
> Is there some way we could get the logs to file instead of just removing it? I remember it was useful to see the return code for the commands when I was debugging finalizers.

@lulf good point I will implement that

PR #4418, opened 2020-04-30
Docs:Add {ApplicationMonitoringNamespace} attribute

hmaclean73:
### Type of change

- Documentation

### Checklist

- [ ] Update documentation


hmaclean73:
We can't remove the ProductMonitoringNamespace attribute because it is being used in other modules.

PR #4422, opened 2020-05-03
Add performance test of create and delete

lulf:
At present, the test needs to be run manually, but can be useful to
measure the impact of changes to the messaging address reconciling.

@kornys where would be the best place to put this kind of test that isn't too long-running but I'd like to run it nightly? Should we define a profile for messaginginfra tests? 

Also, what do I need beside creating the plot file? Does it need to have a specific name to get picked up by jenkins grapher?

### DIFF
--- a/systemtests/src/main/java/io/enmasse/systemtest/messaginginfra/ResourceManager.java
+++ b/systemtests/src/main/java/io/enmasse/systemtest/messaginginfra/ResourceManager.java
@@ -184,6 +184,9 @@ public class ResourceManager {
 
     @SafeVarargs
     public final <T extends HasMetadata> void createResource(boolean waitReady, T... resources) {
+
+        LOGGER.info("Creating {} resources", resources.length);
+
         for (T resource : resources) {
             ResourceType<T> type = findResourceType(resource);
             if (type == null) {
@@ -196,14 +199,18 @@ public class ResourceManager {
                 createResource(waitReady, new NamespaceBuilder().editOrNewMetadata().withName(resource.getMetadata().getNamespace()).endMetadata().build());
             }
 
-            LOGGER.info("Create/Update of {} {} in namespace {}",
+            LOGGER.debug("Create/Update of {} {} in namespace {}",
                     resource.getKind(), resource.getMetadata().getName(), resource.getMetadata().getNamespace() == null ? "(not set)" : resource.getMetadata().getNamespace());
 
             type.create(resource);
-            pointerResources.push(() -> {
-                deleteResource(resource);
-            });
         }
+
+        pointerResources.push(() -> {
+            deleteResource(resources);
+        });
+
+
+        LOGGER.info("Waiting for {} resources to be ready", resources.length);
 
         if (waitReady) {
             for (T resource : resources) {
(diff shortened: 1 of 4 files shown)


systemtests/src/main/java/io/enmasse/systemtest/messaginginfra/ResourceManager.java
```diff
@@ -229,13 +236,14 @@ public MqttClientFactory getMqttClientFactory() {
 
     @SafeVarargs
     public final <T extends HasMetadata> void deleteResource(T... resources) throws Exception {
+        LOGGER.info("Deleting {} resources", resources.length);
```
kornys, line 239: @lulf I don't see any useful reason why logging count of resources helps me in debugging or reading logs

systemtests/src/main/java/io/enmasse/systemtest/scale/ResultWriter.java
```diff
@@ -0,0 +1,45 @@
...
+
+import java.io.BufferedWriter;
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.StandardOpenOption;
+import java.util.Map;
+
+public class ResultWriter {
```
kornys, line 17: @lulf so this can be used in scale test as well, right? Or do we have any similiar class already (so you can reuse it)?

systemtests/src/main/java/io/enmasse/systemtest/messaginginfra/ResourceManager.java
```diff
@@ -184,6 +184,9 @@ public MqttClientFactory getMqttClientFactory() {
 
     @SafeVarargs
     public final <T extends HasMetadata> void createResource(boolean waitReady, T... resources) {
+
+        LOGGER.info("Creating {} resources", resources.length);
```
kornys, line 188: @lulf I really don't think this is useful log info

kornys:
@lulf if you store csv data into `environment.testLogDir().resolve("plot-data")` folder, automation should take care about adding it into jenkins plot processor.

kornys:
@lulf we should discuss on chat how run this new performance/scale tests :)

systemtests/src/main/java/io/enmasse/systemtest/scale/ResultWriter.java
lulf: In the scale test it was kind of built into the test class, so I can refactor it to use this as well.

systemtests/src/main/java/io/enmasse/systemtest/messaginginfra/ResourceManager.java
lulf: Ok, I can remove it. But I will still want to keep the per-address logging at the debug level, because it is way to spammy (and I prefer logging lot of info if there is an error, not if there is not an error).

kornys: per address? it is not only for address but for all resources, right? I don't know, we can discuss it, but I prefer to see creating resources on info level

lulf: I'm not sure its useful to know if it is being created when everything is working, but I do see its useful when it fails. I think maybe what we should've had, is a log buffer that we keep for each test case. All logging happens to this buffer, and it is printed if test fails. If test succeeds, it is just cleared.

Wouldn't this simplify reading logs in event of errors a lot? 

Out of the scope of this PR obviously, so I'll revert the logging changes for now.

systemtests/src/main/java/io/enmasse/systemtest/scale/ResultWriter.java
lulf: Actually, when reading the code, it seems it does not write the same data to result.json and plot data, so I'm not sure how useful this class will be there.

systemtests/src/main/java/io/enmasse/systemtest/messaginginfra/ResourceManager.java
kornys: This is makes sense, yeah it is out of scope for this PR, but I really love this idea

PR #4431, opened 2020-05-04
Save messaging infra perf csv

kornys:


systemtests/src/main/java/io/enmasse/systemtest/scale/ResultWriter.java
```diff
@@ -24,9 +27,9 @@ public ResultWriter(Path rootDir) {
     public void write(Map<String, Object> result) throws Exception {
         Files.createDirectories(rootDir);
         saveResultsFile(rootDir.resolve("results.json"), result);
-        Path plotDir = rootDir.resolve("plot-data");
+        var plotDir = Environment.getInstance().testLogDir().resolve("plot-data");
         Files.createDirectories(plotDir);
-        savePlotCSV(plotDir.resolve("plot.dat"),result);
```
lulf, line 29: IMO to make this class more easily reusable in different contexts, I think the plotDir and the name of the CSV to be passed in the constructor. Relying on global instances like makes things a bit unmanageable.

PR #4435, opened 2020-05-04
Test - Make systemtests build operator registry with olm manifests

famartinrh:
This is just a test, to workaround minikube issues


systemtests/src/main/java/io/enmasse/systemtest/info/TestInfo.java
```diff
@@ -155,6 +155,10 @@ public ExtensionContext getActualTest() {
         return currentTest;
     }
 
+    public ExtensionContext getActualTestClass() {
```
kornys, line 158: getActualTest returns extension context which contains actual test class I think

PR #4456, opened 2020-05-06
Tenant status update check

lulf:
* Add some more debug logging to tenant logging
* Rename to use new selector names and format as agreed.
* Ensure router and broker states are initialized.


lulf:

lulf:

PR #4479, opened 2020-05-06
Fix the primary key on the credentials table

ctron:
### Type of change

- Bugfix

### Description

There was a wrong primary key on the credentials table. It was (tenant, device) and should have been (tenant, type, auth id). This also makes the additional index unnecessary, as the lookup is always by (tenant, type, auth id).

Additionally we require a new index on (tenant, device id) for credential management operations, as they work "by device".

### Checklist

- [ ] Update documentation
- [ ] Update CHANGELOG.md


ctron:
@enmasse-ci run tests ocp4 profile=iot

PR #4480, opened 2020-05-06
Surefire remove workaround and try latest snapshot

kornys:
Seems surefire plugin M5 release is moved to unknown date with fix for handling exceptions in beforeAll/afterAll methods, so I tried to move surefire to snapshot release. My question is can we use snapshot release of maven surefire 3.0.0 or we should still use workaround. One issue with workaround is that we are not able to handle afterAll.


kornys:
it does not work, snapshot is not published in central, so closing

PR #4484, opened 2020-05-07
Adopt the new messaging infrastructure

ctron:
### Type of change

- Design

### Checklist

- [x] Update/write design documentation in `./documentation/design`
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/documentation/design/iot/README.adoc
+++ b/documentation/design/iot/README.adoc
@@ -1,18 +1,82 @@
+:icons: font
 
 = IoT support
 
+[.lead]
 This document describes how IoT is added into the EnMasse system.
+
+NOTE: This is the updated version for EnMasse Core v1.
 
 == Overview
 
-EnMasse IoT support is an additional layer on top of the EnMasse system. It supports IoT
-specific use cases and protocols, like "telemetry ingestion", "command & control", "MQTT", "HTTP", ...
+EnMasse IoT support is an additional layer on top of the
+EnMasse system. It supports IoT specific use cases and
+protocols, like "telemetry ingestion", "command & control",
+"MQTT", "HTTP", …
 
 This is implemented based on the Eclipse Hono project.
 
+== IoT Config CRD
+
+[.lead]
+An "IoT Config" is an IoT infrastructure which can serve multiple IoT tenants (IoTProject).
+
+An IoT Config:
+
+* Is part of the EnMasse infrastructure namespace.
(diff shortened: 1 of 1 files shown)


documentation/design/iot/README.adoc
```diff
@@ -1,38 +1,148 @@
...
+pointing to exactly one messaging infrastructure.
+* Serves all IoT projects assigned to the IoT infrastructure.
++
+This is an implicit assignment. An IoT project resides in a Kubernetes
+namespace, which is assigned to a messaging tenant, which is assigned
+to a messaging infrastructure.
+
+*Target persona:* IoT service admin
+
+*Proposal:* Change the name from `IoTConfig` to `IoTInfrastructure`
```
lulf, line 42: Agreed on renaming it, just noting we should either call it IoTInfra or rename MessagingInfra to MessagingInfrastructure

documentation/design/iot/README.adoc
```diff
@@ -1,38 +1,148 @@
...
+
+=== Missing messaging tenant
+
+*Aspect:* Getting started
+
+If an IoT project gets created in a namespace which does not have
+a _message tenant_ created, then the IoT project will:
+
+* Get the field `.status.phase` set to `Failed`
+* Provide a helpful message in the field `.status.phaseReason`
```
lulf, line 105: Again pointing out an inconsistency in iot vs enmasse with .status.message vs .status.phaseReason, where we should agree on one of them.

documentation/design/iot/README.adoc
ctron: I would propose then to go for the full name. I don't think it s very common to abbreviate in the Kubernetes world.

documentation/design/iot/README.adoc
ctron: Taking at look at the Kubernetes docs:

> Some resources in the v1 API contain fields called `phase`, and associated `message`, `reason`, and other status fields.

So I think it should be either `reason` or `message`, but not `phaseReason`. If we already have `message` elsewhere, I would propose to switch to `message` here as well.

ctron: Done.

documentation/design/iot/README.adoc
kornys: anyway add shortnames please into CRDs :)

PR #3970, opened 2020-02-24
Do not merge - investigation 504 gateway timeout on console login

k-wall:
openshift oauth-proxy image switched to k_wall/oauth-proxy:latest (debug added)

PR #4492, opened 2020-05-07
add failed status in filter dropdown of adress list

drcoolsanjeev:
### Type of change

- Bugfix

### Description
Include failed status in dropdown for address list.
image

### DIFF
--- a/console/console-init/ui/src/modules/address-space/utils/AddressSpaceFormatter.tsx
+++ b/console/console-init/ui/src/modules/address-space/utils/AddressSpaceFormatter.tsx
@@ -8,7 +8,8 @@ import {
   CheckCircleIcon,
   InProgressIcon,
   ExclamationCircleIcon,
-  BanIcon
+  BanIcon,
+  ErrorCircleOIcon
 } from "@patternfly/react-icons";
 import { Badge } from "@patternfly/react-core";
 import { getType } from "utils";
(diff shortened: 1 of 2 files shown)


drcoolsanjeev:
@enmasse-ci run tests testcase=shared.standard.web.FirefoxConsoleTest#testFilterAddressesByStatus crc

drcoolsanjeev:
@enmasse-ci run tests profile=shared testcase=web.**

console/console-init/ui/src/modules/address-space/utils/AddressSpaceFormatter.tsx
```diff
@@ -46,6 +48,9 @@ const statusToDisplay = (phase: string) => {
     case "pending":
       icon = <ExclamationCircleIcon color="red" />;
       break;
+    case "failed":
+      icon = <ErrorCircleOIcon color="red" />;
+      break;
     case "":
```
ajaypratap003, line 54: do we need this blank case if yes then move it in default case. And always add default case in switch to avoid warning

k-wall:
@ajaypratap003 I agree that testClientsMetrics fail is nothing to do with this change.   It seems from an OpenShift issue with oauth2-authentication (#3970).   I agree that testSortConnectionsBySenders looks fishy. I'll talk to you in a separate channel.

console/console-init/ui/src/modules/address-space/utils/AddressSpaceFormatter.tsx
drcoolsanjeev: @ajaypratap003 Yes, I think we need the case for blank in phase as initially for few moment server returns empty phase and once controller picks up the address it will update the status as pending. So that's why there is a case for blank string

PR #4493, opened 2020-05-08
Remove connected and running variables

lulf:
The code that previously required connection to be checked has been
made resilient to handle this error instead, and this should no longer
be needed, allowing the console and infra controller to work with the
old interface.

The lastErr variable is kept to handle the closing of requests that
occur during a reconnect.


lulf:

lulf:
@enmasse-ci run tests testcase=*ConsoleTest

PR #4495, opened 2020-05-08
Disregard incoming remote target on receiving links initiated by the plugin

k-wall:
### Type of change

- Bugfix
- Enhancement / new feature
- Refactoring
- Documentation

### Checklist

- [ ] Update documentation
- [x] Update CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
 * #4305: Inject OpenShift generated custom CA trust bundle into console pod so that console authentication works when a custom CA is in use.
 * #4342: address stuck unready - failed to retrieve addresses: null (##4345)
 * #4358: broker init.sh stop script execution of a script if a command has an error (#4359)
+* #4442: Disregard incoming remote target on receiving links initiated by the plugin (#4495)
 
 ## 0.31.1
 * #4046: add network-only as fetch policy for all queries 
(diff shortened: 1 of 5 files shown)


k-wall:
@lulf unfortunately I needed to use some java package trickery to get the access I needed to disregard the incoming attach source/targets for initiated links.  I don't see a neater way.

In the longer term, I think Artemis plugin connection plugin needs to somehow signal back to Artemis that the link was self-initiated, so that the target (of receiving link) and the source (of sending links) on the attach response can be ignored.  I think an easy way to do this is to add an Qpid Proton Attachment to the Link.  This can be used to allow Artemis to distinguish a self initiated link from one that was initiated by the partner.

https://qpid.apache.org/releases/qpid-proton-j-0.33.4/api/org/apache/qpid/proton/engine/Extendable.html

k-wall:

broker-plugin/plugin/src/main/resources/logging.properties
```diff
@@ -20,9 +20,9 @@
 loggers=org.eclipse.jetty,org.jboss.logging,org.apache.activemq.artemis.core.server,org.apache.activemq.artemis.utils,org.apache.activemq.artemis.journal,org.apache.activemq.artemis.jms.server,org.apache.activemq.artemis.integration.bootstrap,org.apache.activemq.audit.base,org.apache.activemq.audit.message
 
 # Root logger level
-logger.level=INFO
+logger.level=DEBUG
```
lulf, line 23: Should probably switch this back.

k-wall: Fixed

k-wall:
@lulf I addressed the review comment and made a few more tidy ups.  I also introduced a env-var that could be used to turn off the change.

PR #4505, opened 2020-05-11
Test: unit tests for DeviceListToolbar component

rkpattnaik780:
### Type of change

- Unit test

### Description

Unit tests for the `DeviceListToolbar` component.

### DIFF
--- a/console/console-init/ui/src/modules/device/components/DeviceListToolbar/DeviceListToolbar.test.tsx
+++ b/console/console-init/ui/src/modules/device/components/DeviceListToolbar/DeviceListToolbar.test.tsx
@@ -1,0 +1,88 @@
+/*
+ * Copyright 2020, EnMasse authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+
+import React from "react";
+import { render, fireEvent } from "@testing-library/react";
+import { MemoryRouter } from "react-router";
+import {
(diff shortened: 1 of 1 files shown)


console/console-init/ui/src/modules/device/components/DeviceListToolbar/DeviceListToolbar.test.tsx
```diff
@@ -0,0 +1,98 @@
...
+      <MemoryRouter>
+        <DeviceListToolbar {...props} />
+      </MemoryRouter>
+    );
+
+    const dropDownItem0 = getByText("Select none (0 items)");
+    const dropDownItem1 = getByText("Select page (10 items)");
+    const dropDownItem2 = getByText("Select all (100 items)");
+
+    expect(dropDownItem0).toBeInTheDocument();
```
drcoolsanjeev, line 94: expect is not required if you are using getByText. getByText internally manages this thing.

rkpattnaik780: Using expect to maintain consistency with the current codebase, also `toBeInTheDocument()` gives a better idea about purpose of the test.

drcoolsanjeev: we should remove expect for all cases if we are using getByText. It is like doing same thing twice.

PR #4516, opened 2020-05-13
Fix java cli test issue

kornys:


kornys:
@lulf fix tested with enmasse and also new enmasse with shared infra

PR #4522, opened 2020-05-13
Update yarn dependencies

ajaypratap003:
### Type of change

- Enhancement / new feature

### Description
Update yarn dependencies

### Checklist

- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/console/console-init/ui/yarn.lock
+++ b/console/console-init/ui/yarn.lock
@@ -11,9 +11,9 @@
     tslib "^1.10.0"
 
 "@apollo/react-hooks@^3.1.4":
-  version "3.1.4"
-  resolved "https://registry.yarnpkg.com/@apollo/react-hooks/-/react-hooks-3.1.4.tgz#2d966ed4fdf01729113850a72698faee490fd429"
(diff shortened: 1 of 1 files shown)


k-wall:
@ajaypratap003 When you say 'update yarn', are you updating yarn itself or just running `yarn upgrade`. If the former, then you ought to update `yarn.version` in `pom.xml` too.

ajaypratap003:
> @ajaypratap003 When you say 'update yarn', are you updating yarn itself or just running `yarn upgrade`. If the former, then you ought to update `yarn.version` in `pom.xml` too.

@k-wall yes, I updated yarn itself. I'll update yarn version in `pom.xlm` too.

k-wall:
> > @ajaypratap003 When you say 'update yarn', are you updating yarn itself or just running `yarn upgrade`. If the former, then you ought to update `yarn.version` in `pom.xml` too.
> 
> @k-wall I ran `yarn upgrade` command. I don't think so we need to update the yarn version in `pom.xml`. Please correct me if I'm wrong.

No imperative to upgrade that I know of!  It is just the issue title lead me to expect an "upgrade [of] yarn" was the task!  In general, I would suggest that keeping the build tooling version number in sync with your development tooling it best, and not getting too far behind.

PR #4527, opened 2020-05-13
Docs:Add note for viewing address spaces

hmaclean73:
### Type of change

- Documentation

### Description

Docs: Add note for viewing address spaces
### Checklist

- [ ] Update documentation

### DIFF
--- a/documentation/modules/proc-logging-in-console.adoc
+++ b/documentation/modules/proc-logging-in-console.adoc
@@ -29,7 +29,7 @@ endif::[]
 .Procedure
 . In a web browser, navigate to `https://_console-host-name_` where `_console-host-name_` is the {ConsoleName} host name.
 
-. Log in with your OpenShift user credentials. The {ConsoleName} opens.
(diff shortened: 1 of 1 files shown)


documentation/modules/proc-logging-in-console.adoc
```diff
@@ -29,7 +29,7 @@ endif::[]
 .Procedure
 . In a web browser, navigate to `https://_console-host-name_` where `_console-host-name_` is the {ConsoleName} host name.
 
-. Log in with your OpenShift user credentials. The {ConsoleName} opens.
+. Log in with your {KubePlatform} user credentials. The {ConsoleName} opens and lists all the address spaces that you can administer. For information on creating an address space, see link:{BookUrlBase}{BaseProductVersion}{BookNameUrl}#proc-create-address-space-console-messaging[Creating an address space using the {ConsoleName}].
```
k-wall, line 32: On the Kubernetes platform, there is no Kubernetes user credentials.  On Kubernetes the service installer configures EnMasse with an OIDC provider (proc-configure-openid-connect-for-kubernetes.adoc).  So I think we need to have a new variable KubePlatformCredentials which is defined OpenShift (master-openshift.adoc) and OIDC (master-kubernetes.adoc) respectively.

hmaclean73: Thanks for the feedback. I will make this change.

PR #4529, opened 2020-05-13
Modify dialog prompt component

drcoolsanjeev:
### Type of change

- Enhancement / new feature

### Description
Modify dialog prompt component to handle confirm label button and color and add icon in the header

image


drcoolsanjeev:
@anukritijha , No it will not affect other designs.

PR #4543, opened 2020-05-15
Broker readiness probe

lulf:
Adds a broker-probe binary that can be used to attach a consumer to an address defined by the environment.

For shared infra, the probe address is created by the enmasse-operator. The AMQP command client have also been fixed to track restarts, as the tests failed due to operator missing a broker restart.

For standard infra, the probe address is created by the agent.

For brokered, the probe address is statically defined in the broker.xml, as there are only 1 broker in that case.


lulf:

lulf:
@enmasse-ci run tests profile=isolated-shared-infra crc

lulf:
@enmasse-ci run tests profile=isolated-shared-infra crc

lulf:

lulf:

lulf:
@enmasse-ci run tests profile=isolated-shared-infra crc

lulf:

lulf:

cmd/broker-probe/main.go
```diff
@@ -0,0 +1,61 @@
...
+	client, err := amqp.Dial("amqp://127.0.0.1:5672", amqp.ConnSASLPlain(username, password), amqp.ConnConnectTimeout(probeTimeout), amqp.ConnProperty("product", "broker-probe"))
+	if err != nil {
+		log.Fatal(err, "Error dialing broker")
+	}
+	log.Println("Connected, creating session")
+
+	defer func() {
+		_ = client.Close()
+	}()
+
```
k-wall, line 47: We should wrap the session/link code with a https://golang.org/pkg/context/#WithTimeout.  This serves to minimise the risk that Probe invocations could stack up (adding to a storm) if things start to go wrong (say Artemis in GC death spiral).

broker-plugin/plugin/src/main/resources/brokered/broker.xml
```diff
@@ -108,6 +108,13 @@ under the License.
             <address-full-policy>${ADDRESS_FULL_POLICY}</address-full-policy>
          </address-setting>
       </address-settings>
+      <addresses>
```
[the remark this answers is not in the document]
lulf: Given that it is statically defined in the broker.xml, there is no need to create an address setting that prevents it from being auto-created.

address-space-controller/src/main/resources/templates/brokered-space-infra.yaml
```diff
@@ -112,6 +112,18 @@ objects:
             value: serverData
           - name: HOME
             value: /var/run/artemis/split-1/
+          - name: PROBE_ADDRESS
+            value: "!!READINESS"
+          - name: PROBE_USERNAME
```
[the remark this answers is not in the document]
lulf: With the new broker version, it should be possible to use the per-acceptor permissions. Maybe we should use that for this purpose then and create a localhost acceptor where the probe user can log in?

cmd/broker-probe/main.go
k-wall: Mmm.  I didn't notice they didn't take a cxt.  One way I suppose would be to schedule a close of the connection after n seconds.    Looking at the link code and the session is looks like if the session/connection is closed, it will stop awaiting the response.

btw.  I have a simple Schedule implementation in pkg/consolegraphql/server/utils.go:14.  I guess you'll need to use a` once()` to organise for the connection to be closed exactly one time.

lulf: I've attempted an approach where I spin off the probe in a goroutine and use a timeout on that in the main function.

lulf:

lulf:

PR #4545, opened 2020-05-16
Delivering management functionality for shared infra

k-wall:
### Type of change

- Design sketch

I was thinking about how we'd deliver management functionality in the shared world.
I had some thoughts about a possible solution that I've captured as an embryonic proposal.

https://github.com/k-wall/enmasse/blob/man-tasks-for-shared/documentation/design/proposals/management_tasks-for-shared.adoc

All comments welcome.

### Checklist

- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/documentation/design/proposals/management_tasks-for-shared.adoc
+++ b/documentation/design/proposals/management_tasks-for-shared.adoc
@@ -17,7 +17,7 @@ The current design has number of drawbacks:
    and HTTP proxies).  If the management request takes too long the response channel is lost meaning the user has no
    mechanism to monitor the request's completion.  Purging a deep queue or moving a significant number of messages
    is likely to breach timeouts. 
-1. The deisgn has no mechnaism to monitor the progress of a request
+1. The deisgn has no mechnaism to monitor the progress of a request or even know if it is still running.
 1. The deisgn has no mechnaism to cancel a management request.
 1. The Console UI uses modals UI elements to initiate the request and blocks awaiting the response. The user is unable
    to continue to interact with the app whilst the request is in progress.
--- a/documentation/design/proposals/management_tasks-for-shared.adoc
+++ b/documentation/design/proposals/management_tasks-for-shared.adoc
@@ -45,6 +45,9 @@ https://kubernetes.io/docs/concepts/workloads/controllers/jobs-run-to-completion
 The tenant admin would be able to CRUDWL `ManagementTask` from the CLI using `kubectl`/`oc` tooling (usual RBAC permissions
 apply), or later an EnMasse CLI tool might abstract away some details. 
 
+The operator would orchestrate the task.  The precise details of the orchestration would vary from task to task.   For some
+tasks we might delegate the work of the to a Kubernetes job in the infra namespace.
+
 From the Console. functions such as purge queue/subscription would, behind the button cause an instance of this object to
 be created.  There would be an area of the UI which lists running/recently finsihed tasks.  From this page, you'd be able
 to cancel running tasks too.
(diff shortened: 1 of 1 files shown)


documentation/design/proposals/management_tasks-for-shared.adoc
```diff
@@ -0,0 +1,91 @@
...
+  phase: New | Running | Failed | Completed
+  message: ..
+  conditions:
+    ...
+```
+
+Open Questions:
+
+* The ability of create a ManagementTask would be typically granted to an Tenant Admin.  This would give that user
+  the permission to do any type of management functiion permitted by the task.  Would we need to be more discriminating?
```
lulf, line 89: We could add another 'verb' to the MessagingRole called 'manage', which, when granted would allow you to perform purge and move operations on addresses where you have that role, which is similar to what we have today. That would allow creating more fine-grained RBAC rules. However, I don't think that would change the above design fundamentally.

documentation/design/proposals/management_tasks-for-shared.adoc
```diff
@@ -0,0 +1,91 @@
...
+* View message management function - how would that be delivered?
```
lulf, line 90: We could similarly define a 'view' verb on the MessagingRole to handle this I think.

documentation/design/proposals/management_tasks-for-shared.adoc
k-wall: I was seeing this as being permissioned by the Kubernetes domain rather than the EnMasse Messaging permissions model (MessagingRole/MessagingRoleBinding).   What is your idea?

I was wondering in [RBAC subresources](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#referring-to-resources) could help.  In order to be able to purge an address, the EnMasse Operator would check for a permission granting `address/purge` on the target resource.  My difficulty is I don't see a way to find out who created the ManagementTask.

lulf: @k-wall Yeah, I was thinking to use the messaging permissions model, but maybe it would complicate things too much. It could be that RBAC on ManagementTask is sufficient given that a namespace only have 1 messaging tenant anyway. I'm not too familiar of who is typically given these permissions in an organization.

It would be nice to use subresources for this, I'm not sure if it would be allowed to create generic CRD subresources that are not `scale` or `status`.

k-wall: @lulf My gut feel is that RBAC would be sufficient.  In other words, I think as tenant admin should be able to create any type of ManagementTask.

documentation/design/proposals/management_tasks-for-shared.adoc
```diff
@@ -0,0 +1,88 @@
...
+be created.  There would be an area of the UI which lists running/recently finsihed tasks.  From this page, you'd be able
+to cancel the task too.
+
+=== CRD sketch
+
+```
+apiVersion: enmasse.io/v1
+kind: ManagementTask
+metadata:
+  name: mytenantnamespace
```
k-wall, line 58: name is arbitrary - the tenant would create this resource in his messaging project.

documentation/design/proposals/management_tasks-for-shared.adoc
```diff
@@ -0,0 +1,88 @@
...
+1. The Console UI uses modals UI elements to initiate the request and blocks awaiting the response. The user is unable
+   to continue to interact with the app whilst the request is in progress.
+1. There is no provision for a command line interface.
+1. Poorly defined RBAC model.
+
+== High Level Proposal
+
+Introduce a new CRD `ManagementTask`.  A `ManagementTask` executes a management function (such as a purge) in a
+tenant namespace within EnMasse.  It would be rather similar to
+https://kubernetes.io/docs/concepts/workloads/controllers/jobs-run-to-completion/[Jobs] from Kubernetes.
```
k-wall, line 32: @ctron said "could we actually implement the running of the management task as Kubernetes jobs.

documentation/design/proposals/management_tasks-for-shared.adoc
k-wall: Done.

documentation/design/proposals/management_tasks-for-shared.adoc
k-wall: Note in document.

PR #4562, opened 2020-05-19
User stories update 2

lulf:


documentation/design/user-stories/shared-infra.adoc
```diff
@@ -760,28 +756,30 @@ metadata:
...
-        storage: 20Mi
+        queueSizeMemory: 5Mi
+        queueSizeStorage: 20Mi
       limits:
-        memory: 10Mi
-        storage: 30Mi
-
-  # Per-address type properties
-  queue:
+        # Amount of memory queues with this plan can use on the broker
```
k-wall, line 766: How will the limits be used?

lulf: They translate to the per-address settings (max-size-bytes) for the address. When configuring an address, the operator will look at the sum of limits for all addresses and check that the new address does not cause it to go above the limits set in the MessagingPlan. If it does, then status section would be updated with QuotaExceeded error, similar to what we do for address plans vs address space plans today.

lulf: I've added a comment to address this. My thinking was that limits would translate to broker address settings. Requests would be used for scheduling. If requests < limits, the scheduler may move this queue to a different broker to improve space efficiency/density. This is modelled after how requests/limits works for pods.

PR #4565, opened 2020-05-19
Add validations on metadata form

anukritijha:
### Type of change

- Enhancement / new feature

### Description

This PR adds the following features to metadata form

- Adds dynamic child rows, its key, values and types in metadata json 
- Enables/Disables the value field in case of Array or Object
- Enables/Disables the child row addition button
- Removes the generic typeahead component to patternfly select component
- Adds validations to metadata fields

image

### DIFF
--- a/console/console-init/ui/src/modules/address-detail/components/AddressLinks/AddressLinks.tsx
+++ b/console/console-init/ui/src/modules/address-detail/components/AddressLinks/AddressLinks.tsx
@@ -17,7 +17,7 @@ import { ExternalLinkAltIcon } from "@patternfly/react-icons";
 import { css } from "@patternfly/react-styles";
 import { Link } from "react-router-dom";
 import { Tooltip, TooltipPosition } from "@patternfly/react-core";
-import { StyleForTable } from "modules/address-space/components/AddressSpaceList/AddressSpaceList";
+import { StyleForTable } from "modules/msg-and-iot";
 
 interface IAddressLinksProps {
   rows: IAddressLink[];
--- a/console/console-init/ui/src/modules/address-space/components/AddressSpaceList/AddressSpaceList.tsx
+++ b/console/console-init/ui/src/modules/address-space/components/AddressSpaceList/AddressSpaceList.tsx
@@ -14,9 +14,8 @@ import {
   SortByDirection
 } from "@patternfly/react-table";
 import { StyleSheet, css } from "@patternfly/react-styles";
-import { ProjectToolbar } from "modules/project/components/ProjectToolbar";
 
-export const StyleForTable = StyleSheet.create({
+const StyleForTable = StyleSheet.create({
   scroll_overflow: {
     overflowY: "auto",
     paddingBottom: 100
--- a/console/console-init/ui/src/modules/address/components/AddressList/AddressList.tsx
+++ b/console/console-init/ui/src/modules/address/components/AddressList/AddressList.tsx
@@ -18,7 +18,7 @@ import {
 import { css } from "@patternfly/react-styles";
 import { FormatDistance } from "use-patternfly";
 import { useWindowDimensions } from "components";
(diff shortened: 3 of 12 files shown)


console/console-init/ui/src/components/TypeAheadSelect/TypeAheadSelect.tsx
```diff
@@ -75,6 +75,7 @@ const TypeAheadSelect: React.FunctionComponent<ITypeAheadSelectProps> = ({
           } else {
             setOptions([
               <SelectOption
+                key={0}
```
drcoolsanjeev, line 78: can we replace the key with dynamic value like index?

console/console-init/ui/src/modules/device/dialogs/CreateDevice/MetaDataRow.tsx
```diff
@@ -11,108 +11,165 @@ import {
...
-            />
-            <Button
-              variant="control"
-              aria-label="Add child on button click"
-              onClick={e => handleAddChildRow(e, "context")}
-            >
-              <PlusIcon />
-            </Button>
+            ></Select>
+            {isChildAdditionEnabled === true ? (
```
drcoolsanjeev, line 139: we can replace ternary operator by && like {isChildAdditionEnabled && (
<Button
                variant="control"
                aria-label="Add child on button click"
                onClick={e => {
                  let parentValue: string = metadataRow.key;
                  handleAddChildRow(e, parentValue);
                }}
              >
                <PlusIcon />
              </Button>
)}

console/console-init/ui/src/modules/device/utils/constant.ts
```diff
@@ -60,4 +60,17 @@ const getInitialFilter = () => {
   return filter;
 };
 
-export { deviceTypeOptions, deviceStatusOptions, getInitialFilter };
+const deviceRegistrationTypeOptions: ISelectOption[] = [
+  { key: "string", label: "String", value: "String" },
+  { key: "number", label: "Number", value: "Number" },
```
drcoolsanjeev, line 65: label should be Numeric

drcoolsanjeev: also value will be in lower case

console/console-init/ui/src/components/TypeAheadSelect/TypeAheadSelect.tsx
anukritijha: There is no dynamic value here, as there is only one element.

console/console-init/ui/src/modules/device/dialogs/CreateDevice/MetaDataRow.tsx
```diff
@@ -11,108 +11,165 @@ import {
...
+    updatedValueMetadata[rowIndex].value = propertyValue;
+    setMetadataList(updatedValueMetadata);
   };
 
   const onPropertySelect = (e: any, selection: SelectOptionObject) => {
     setPropertySelected(selection.toString());
-    setPropertyInput(undefined);
   };
 
   const onPropertyClear = () => {
```
rkpattnaik780, line 110: Not able to clear the input using the close icon.

console/console-init/ui/src/modules/device/dialogs/CreateDevice/MetaData.tsx
```diff
@@ -4,36 +4,39 @@
...
+      ...metadataList,
+      { key: "", value: [], type: defaultType }
+    ]);
   };
 
   return (
     <>
       <MetaDataHeader sectionName="Default properties" />
-      {formData.map((data, index) => (
+      {metadataList.map((matadataRow, index) => (
```
rkpattnaik780, line 34: `metadatarow` is not being used anywhere.

PR #4577, opened 2020-05-20
Fix logging

lulf:


lulf:

kornys:
@lulf does it require CSV change as well?

lulf:
@kornys no, there was just permissions i forgot to remove in the bundle a long time ago. They are not in the CSV so nothing to remove.

PR #4595, opened 2020-05-23
Push documentation directly to website

lulf:
* Use asciidoctor maven plugin
* Remove ruby dependency in build
* Update developer and documentation contributor guide
* Restructure documentation source
* Publish docs to enmasse.io website on every master and release build.

Must be merged after https://github.com/EnMasseProject/enmasseproject.github.io/pull/4


.github/scripts/copy_docs.sh
```diff
@@ -0,0 +1,26 @@
...
+
+rm -rf  ${TO}/documentation/${DOCVERSION}/images
+cp -vrL ${FROM}/documentation/target/generated-docs/htmlnoheader/* ${TO}/documentation/${DOCVERSION}
+
+pushd ${TO}
+if [[ -z $(git status -s) ]]; then
+    echo "No changes to the output on this push; exiting."
+    exit 0
+fi
+
```
kornys, line 19: @lulf Im not sure how checkout/clone works in GH action but in Jenkins it detach after clone/checkout with plugin, so if GH actions do the same, you need to call `git checkout master`.

lulf: @kornys I believe this is done by the checkout plugin (in the workflow file I specify the branch). Do you know, @obabec ?

obabec: @kornys @lulf As far as I know (from v1 checkout) it checkout to branch you specify. And repo stays on that branch until you again checkout with plugin. I can try it and let you know later today :)

lulf: Ok. Lets discuss it and merge tomorrow, its not a huge problem if it doesn't work right away and I can fix it if it fails.

PR #4605, opened 2020-05-26
Lower default globalmax

### LINKED ISSUE #4574: Lower default globalMaxSize of broker to 1/4 of broker heap
**Description**

Using the default setting of -1 causes the globalMaxSize to be adjusted to 1/2 of the JVM size, which is not sufficient to prevent the broker from going out of memory.

The address space controller/standard-controller should adjust this setting based on the container memory limits (1/8 of container memory, 1/4 of jvm memory).

lulf:
### Type of change

- Bugfix

### Description

Modifies examples to use a lower default global max as well as make the init script set the global max to 1/8 of container memory if it was unset (-1).

Fixes #4574 

### Checklist

- [x] Update documentation
- [x] Reference relevant issue(s) and close them after merging
- [x] Update CHANGELOG.md


lulf:

lulf:

Issue #3157, opened 2019-09-05
Add support for per-address limits in broker


lulf:
**Overview**

At present, the facility for doing per-address configuration in the broker is not able to handle large amount of addresses. This issue is for tracking the state of that, and make sure we enable this capability once it is supported in the broker

**Design**

Per-address settings will be created on the broker.  These will be updated when an address is added / removed.  And synchronized when the agent starts up.

Set the maxSizeBytes in the per-address settings.  This will be calculated from globalMaxSize and the addressplan.spec.resource.broker.  The globalMaxSize is taken from infraconfig.spec.broker.globalMaxSize.  If it isn't set on the infraconfig, request it from the broker to get the default.  The globalMaxSize is multiplied by the addressplan.spec.resource.broker to get the maxSizeBytes, if addressplan.spec.resource.broker is > 0 & < 1.

The feature will be implemented for broker and standard.

**Tasklist**

- [ ] Extend agent to set maxSizeBytes from infraconfig.spec.broker.globalMaxSize & addressplan.spec.resource.broker to the address-settings
- [ ] When starting up agent, sync the address-settings.
- [ ] When creating / deleting addresses, update the address-settings.
- [ ] When address plans are changed, update the address-settings if the settings are different.
- [ ] Write system tests

PR #4611, opened 2020-05-26
created settings per address in the broker

vbusch:
### Type of change

- Enhancement / new feature

### Description
issue #3157 
- in the standard address space and brokered these are updated when we an address plan is changed, or when an address is created / deleted / modified

### Checklist

- [ ] Update documentation
- [ ] Update CHANGELOG.md


agent/lib/broker_address_settings.js
```diff
@@ -31,52 +43,100 @@ function extract_address_plan(object) {
...
-    return new Promise(function (resolve, reject) {
-        self.watcher.once('updated', function () {
-            log.info('plans have been retrieved');
+BrokerAddressSettings.prototype.get_changes = function (name, addressplans, unchanged) {
+    var c = myutils.changes(this.last[name], addressplans, addressplan_compare, unchanged, description);
+    this.last[name] = addresses;
+    return c;
+};
+
+BrokerAddressSettings.prototype.get_global_max_size = function () {
```
lulf, line 97: I think a simpler way to get this would be to inject the value as an environment variable in the template..

agent/lib/broker_address_settings.js
```diff
@@ -18,9 +18,21 @@
 var log = require("./log.js").logger();
 var kubernetes = require('./kubernetes.js');
 var myutils = require('./utils.js');
+var events = require('events');
+var util = require('util');
 
 function kubernetes_resource_compare(a, b) {
-    return myutils.string_compare(a.metadata.name, b.metadata.name);
+    return myutils.string_compare(a.name, b.name);
```
k-wall, line 25: I think this is used to compare address settings, rather than Kubernetes resources, if so, it should be renamed.

agent/lib/broker_address_settings.js
```diff
@@ -18,9 +18,21 @@
...
+}
+
+function extract_spec(def) {
```
k-wall, line 28: name does not reflect the implementation well.  How about something like flatten_plan?

agent/lib/broker_address_settings.js
```diff
@@ -86,7 +146,7 @@ BrokerAddressSettings.prototype.delete_settings = function (plan) {
 };
 
 BrokerAddressSettings.prototype.updated = function (objects) {
-    var plans = objects.map(extract_address_plan).filter(required_broker_resource);
+    var plans = objects.filter((plan) => plan !== undefined).map(extract_spec).filter(required_broker_resource);
```
k-wall, line 149: is it really necessary to filter for undefined objects? sounds fishy to me.

agent/lib/broker_controller.js
```diff
@@ -89,6 +100,28 @@ BrokerController.prototype.sync_addresses = function (addresses) {
...
+    addressplans.forEach(function (addressplan) {
+
+        for (var address in self.addresses) {
+            if (self.addresses[address].plan === addressplan.name) {
+                addrs.push(self.addresses[address]);
+            }
+        }
+    });
+        if (addrs.length) {
+            return Promise.all(addrs.map(function (addr) {
```
k-wall, line 115: Should we apply the 'p-limit' pattern here?

agent/lib/broker_controller.js
```diff
@@ -89,6 +100,28 @@ BrokerController.prototype.sync_addresses = function (addresses) {
...
+                return self.update_address_setting(addr);
+            })).then(function () {log.info('updated all ')})
```
k-wall, line 117: make this logging more specific, or remove it.

agent/lib/broker_controller.js
```diff
@@ -248,6 +281,17 @@ function is_not_internal(conn) {
     return conn.user !== undefined && conn.user.match('^agent.[a-z0-9]+$') == null; //Can't get properties or anything else on which to base decision yet
 }
 
+function compare_address_settings(orig_address_settings, new_address_settings) {
+    for (var name in orig_address_settings) {
+        log.debug('comparing %s: %s %s', name, orig_address_settings[name], new_address_settings[name])
+        if (new_address_settings[name] && myutils.string_compare(orig_address_settings[name], new_address_settings[name]) !=0){
```
k-wall, line 287: recompilation of string_compare in the false case.

agent/lib/broker_controller.js
```diff
@@ -248,6 +281,17 @@ function is_not_internal(conn) {
     return conn.user !== undefined && conn.user.match('^agent.[a-z0-9]+$') == null; //Can't get properties or anything else on which to base decision yet
 }
 
+function compare_address_settings(orig_address_settings, new_address_settings) {
+    for (var name in orig_address_settings) {
```
k-wall, line 285: this method won't catch the case where the new settings object has a value that the old  does not.  Is this important?

agent/lib/broker_controller.js
```diff
@@ -383,40 +427,58 @@ function translate(addresses_in, excluded_names, excluded_types) {
     return addresses_out;
 }
 
+BrokerController.prototype.update_address_setting = function (a) {
+    var self = this;
+    return self.broker.getAddressSettings(a.address).then(function (orig_settings) {
+
+        return self.get_address_settings(a).then(function (new_settings) {
+            if (compare_address_settings(orig_settings, new_settings) != 0) {
+                self.broker.removeAddressSettings(a.address).then(function () {
```
k-wall, line 436: error handling missing

agent/lib/broker_controller.js
```diff
@@ -383,40 +427,58 @@ function translate(addresses_in, excluded_names, excluded_types) {
...
-        log.info('[%s] Deleting queue "%s"...', self.id, a.address);
-        return self.broker.destroyQueue(a.address).then(function () {
-            log.info('[%s] Deleted queue "%s"', self.id, a.address);
-            self.post_event(myevents.address_delete(a));
-        }).catch(function (error) {
-            log.error('[%s] Failed to delete queue %s: %s', self.id, a.address, error);
-            return self.broker.deleteAddress(a.address).then(function () {
-                log.info('[%s] Deleted anycast address %s', self.id, a.address);
+
+    return self.broker.getAddressSettings(a.address).then(function (settings) {
```
k-wall, line 447: why do we need to lookup the address settings here?  also, if the address settings are missing is it correct to not delete the queues etc?

agent/lib/broker_controller.js
```diff
@@ -426,14 +488,17 @@ BrokerController.prototype.delete_address_and_settings = function (a) {
         log.info('[%s] Deleted address-settings for "%s"', self.id, a.address);
         return self.delete_address(a);
     }).catch(function (error) {
-        log.error('[%s] Failed to delete address settings for "%s": %s', self.id, a.address, error);
+        log.error('[%s] Failed to delete address setting for "%s": %s', self.id, a.address, error);
```
k-wall, line 491: I don't understand the need for the two catches.  Would moving the self.delete_address(a) to a .finally() make the code simpler?

agent/lib/broker_controller.js
```diff
@@ -476,18 +541,50 @@ BrokerController.prototype.create_address = function (a) {
...
+                        })
+                        .catch(function (error) {
+                            log.error('[%s] Failed to create address settings for "%s": %s', self.id, a.address, error);
+                            return self.create_address(a);
+                        });
+                });
+        });
 };
 
+BrokerController.prototype.ensure_connector = function (a) {
```
k-wall, line 561: Can you tell me more about this part of the change?

agent/lib/broker_address_settings.js
```diff
@@ -31,52 +43,100 @@ function extract_address_plan(object) {
     }
 }
 
+function addressplan_compare(a, b) {
+    return myutils.string_compare(a.name, b.name);
+}
+
```
k-wall, line 49: let's name this function so the name doesn't collide with the map.

k-wall:
@lulf @vbusch I don't understand how this feature is supposed to work with PAGE policy.  If I infer the intent properly from the code we allocate a plan defined proportion of the globalMaxSize to each address, so in the PAGE case, the address' size would be constrained by RAM.

agent/lib/broker_address_settings.js
```diff
@@ -31,52 +43,100 @@ function extract_address_plan(object) {
...
-        var r = this.required_broker_resource[plan];
-        if (r) {
-            return {
-                maxSizeBytes: r * global_max_size,
-                addressFullMessagePolicy: 'FAIL'
-            };
-        } else {
-            log.info('no broker resource required for %s, therefore not applying address settings', plan);
-        }
+    var r = this.required_broker_resource[plan];
```
k-wall, line 132: Doesn't this calculation need to factor in partitions?

agent/lib/broker_address_settings.js
```diff
@@ -31,52 +43,100 @@ function extract_address_plan(object) {
...
+    if (r) {
+        return {
+            maxSizeBytes: Math.round(r * global_max_size),
+            addressFullMessagePolicy: 'FAIL'
+        };
     } else {
-        log.info('no global max, therefore not applying address settings');
+        log.info('no broker resource required for %s, therefore not applying address settings', plan);
     }
 };
```
k-wall, line 142: in delete_settings, I think you are passing the result of extract_spec, so plan.metadata will be undefined and will crash.

agent/lib/broker_address_settings.js
vbusch: Renamed to 'addressplan_name_compare'

agent/lib/broker_address_settings.js
vbusch: was able to remove the filter

agent/lib/broker_controller.js
vbusch: added p-limit

agent/lib/broker_controller.js
vbusch: removed

agent/lib/broker_controller.js
vbusch: updated

agent/lib/broker_controller.js
vbusch: The broker returns all the properties with default values.  It is important that we only compare the ones in the new_settings.

agent/lib/broker_controller.js
vbusch: added

agent/lib/broker_controller.js
vbusch: removed the extra

agent/lib/broker_controller.js
vbusch: wasn't required

vbusch:

agent/lib/artemis.js
```diff
@@ -499,7 +499,7 @@ var address_settings_fields = {
     'addressFullMessagePolicy': 'FAIL',
     'slowConsumerThreshold':-1,
     'slowConsumerCheckPeriod':-1,
-    'slowConsumerPolicy':'DROP',
+    'slowConsumerPolicy':'KILL',
```
lulf, line 502: Why this change?

agent/lib/broker_address_settings.js
```diff
@@ -18,65 +18,90 @@
 var log = require("./log.js").logger();
 var kubernetes = require('./kubernetes.js');
 var myutils = require('./utils.js');
+var events = require('events');
+var util = require('util');
 
-function kubernetes_resource_compare(a, b) {
-    return myutils.string_compare(a.metadata.name, b.metadata.name);
+function extract_flatten_plan(def) {
```
lulf, line 24: Is there any real need to 'flatten' the definition? Iirc the reason this is done for addresses is because of legacy configuration format from looong ago.

agent/lib/utils.js
```diff
@@ -286,3 +286,26 @@ module.exports.get = function (object, fields, default_value) {
...
+var multipliers = {};
+multipliers.B = 1;
+multipliers.KB = ( multipliers.B * 1024 );
+multipliers.MB = ( multipliers.KB * 1024 );
+multipliers.GB = ( multipliers.MB * 1024 );
+module.exports.parseToBytes = function (input) {
+    var unit = input.toUpperCase().match(/([KMGTPEZ]I?)?B$/);
+
+    if (!unit) {
+        log.warn( "parseToBytes input does not contain a supported unit of measurement.  Using 'B': %s", input );
```
lulf, line 299: We should probably use 'k' or both 'k' and 'K', because 'k' is the official.

agent/lib/artemis.js
vbusch: DROP has been removed.  The available options are kill & notify
https://github.com/apache/activemq-artemis/blob/master/artemis-server/src/main/java/org/apache/activemq/artemis/core/settings/impl/SlowConsumerPolicy.java

agent/lib/broker_address_settings.js
vbusch: ah, I'll change that.  I saw it was done elsewhere and followed.

agent/lib/utils.js
vbusch: by switching it to upper case(toUpperCase) both will be allowed.

lulf: Ah, i didn't catch that

agent/lib/artemis.js
```diff
@@ -523,6 +523,22 @@ Artemis.prototype.removeAddressSettings = function (match) {
...
+Artemis.prototype.getAddressSettings = function (match) {
+    return this._request('broker', 'getAddressSettingsAsJSON', [match]).then(function (result) {
+        return JSON.parse(result);
+    });
+};
+
+Artemis.prototype.listAddressSettings = function () {
+    var self = this;
+    return this.listAddresses().then(function (addresses) {
+        return Promise.all(addresses.map(function (a) {
```
k-wall, line 535: suggest using the `plimit` pattern to reduce the inflight.

vbusch: I'll remove that method.  broker_controller already has the list of addresses, and is using that list to call getAddressSettings (with p-limit).  
This one was the original, that is no longer used.

agent/lib/broker_address_settings.js
k-wall: @lulf @vbusch if you have a plan that requires 0.5 broker and you are partitioning across 2, I think that means you should have 0.25 per broker.  I therefore think the calculation needs to divide by patitions.  Is that right?

agent/lib/broker_address_settings.js
```diff
@@ -86,34 +111,35 @@ BrokerAddressSettings.prototype.delete_settings = function (plan) {
...
-        return this.wait_for_plans().then(function () {
-            return global_max_size_promise.then(function (global_max_size) {
-                var settings = self.generate_address_settings(address.plan, global_max_size);
-                log.info('using settings %j for %s', settings, address.address);
-                return settings;
-            });
-        });
-    } else {
+
+    var max_size_as_bytes = this.config.BROKER_GLOBAL_MAX_SIZE ? myutils.parseToBytes(this.config.BROKER_GLOBAL_MAX_SIZE) : 0;
```
k-wall, line 132: Would it be cleaner to read GLOBAL_MAX_SIZE/request max size from the Broker once in `start()`  After all the value is not going to change.  

something like:

```
BrokerAddressSettings.prototype.start = function () {
...
read_global_max_size().then((size) => {
self.global_max_szie = size;
self.watcher.on('updated', self.updated.bind(this));
});
```

agent/lib/broker_controller.js
```diff
@@ -383,8 +427,38 @@ function translate(addresses_in, excluded_names, excluded_types) {
...
+
+                    return self.broker.removeAddressSettings(a.address).then(function () {
+
+                        return self.broker.addAddressSettings(a.address, new_settings);
+                    }).catch(function (error) {
+                        log.error('[%s] Adding the new address setting (removal failed )%s: %s', self.id, a.address, error);
+                        return self.broker.addAddressSettings(a.address, new_settings)
+                    });
+                } else {
+                    log.info('[%s] Address settings match for %s: not updating', self.id, a.address);
```
k-wall, line 446: To verbose (systems that have many addresses). Change to debug.

agent/lib/utils.js
```diff
@@ -286,3 +286,26 @@ module.exports.get = function (object, fields, default_value) {
     }
     return o || default_value;
 };
+
+var multipliers = {};
+multipliers.B = 1;
+multipliers.KB = ( multipliers.B * 1024 );
+multipliers.MB = ( multipliers.KB * 1024 );
+multipliers.GB = ( multipliers.MB * 1024 );
+module.exports.parseToBytes = function (input) {
```
k-wall, line 295: it seems weird that the regex allows GTPEZ but the multipler map does not.

agent/lib/broker_controller.js
```diff
@@ -383,8 +427,38 @@ function translate(addresses_in, excluded_names, excluded_types) {
...
+                }
+            } else {
+                //Settings weren't created, and reason is already logged. Most likely broker resource not required.
+            }
+        }).catch(function (error) {
+            log.error('[%s] Failed to create new address setting %s: %s', self.id, a.address, error);
```
k-wall, line 452:
```suggestion
            log.error('[%s] Failed to create new address setting %s: %s', self.id, a.address, error);
            return Promise.reject(error);
```

agent/lib/broker_controller.js
```diff
@@ -383,8 +427,38 @@ function translate(addresses_in, excluded_names, excluded_types) {
...
+        });
+    }).catch(function (error) {
+        log.error('[%s] Failed to retrieve brokers address setting: %s', self.id, a.address, error);
```
k-wall, line 455:
```suggestion
        log.error('[%s] Failed to retrieve brokers address setting: %s', self.id, a.address, error);
        return Promise.reject(error);
```

agent/lib/broker_controller.js
k-wall, line 446:
```suggestion
                    log.info('[%s] Address settings match for %s: not updating', self.id, a.address);
                    return Promise.resolve();
```

agent/lib/broker_controller.js
```diff
@@ -383,8 +427,38 @@ function translate(addresses_in, excluded_names, excluded_types) {
...
+                        return self.broker.addAddressSettings(a.address, new_settings);
+                    }).catch(function (error) {
+                        log.error('[%s] Adding the new address setting (removal failed )%s: %s', self.id, a.address, error);
+                        return self.broker.addAddressSettings(a.address, new_settings)
+                    });
+                } else {
+                    log.info('[%s] Address settings match for %s: not updating', self.id, a.address);
+                }
+            } else {
+                //Settings weren't created, and reason is already logged. Most likely broker resource not required.
```
k-wall, line 449:
```suggestion
                //Settings weren't created, and reason is already logged. Most likely broker resource not required.
                return Promise.resolve();
```

agent/lib/artemis.js
k-wall: ok

agent/lib/broker_address_settings.js
k-wall: @lulf @vbusch is my comment invalid?

lulf: Sorry, I have forgotten to respond to this. Yes, the credits should be divided by the number of partitions.

vbusch:

vbusch:

vbusch:
Investigated the failing testcase "isolated.CommonTest#testRestartComponents"
It passes on master.  But the logs on master show the same issue.
There are 2 address spaces, brokered & standard.  Each address space has the same queue test-queue.  When restarting all of enmasse (restarting single components works), the agent of the standard address space has an issue with configuring its router.
It is getting all the addresses (for both address spaces), then it tries to create the test-queue address twice.  And fails the 2nd time (ERROR creating address test-queue on qdrouterd-6c2f4cf-0 => Name conflicts with an existing entity)
It does retry and then gets (ERROR configuration update for qdrouterd-6c2f4cf-0 not up to date (attempt 3 of 3), agent ERROR Unable to apply desired configuration; gave up after 3 attempts)
I haven't determined yet, why the test passes on master.  But it doesn't seem right that the standard address space is creating addresses that belong in the brokered address space.

k-wall:
@lulf @vbusch reflecting in this change, I'm not sure we have done the right thing.   I think we should be driving the address settings based on the `status.planStatus` section of each address and _not_ the address plan directly. Taking the it from planStatus is the correct thing to do because it means that the address's controller has had chance to apply the validations.  Also it has an implementation benefit - the additional events that @vbusch has had to introduce to deal with plan updates make for complicated and complex code that is hard to reason about.  Making this change would get rid of this complication.

For standard - the status.planStatus is already in place.  For brokered, this needs an extension to the controller (agent), but this is code I most have already on the TTL change.

vbusch:
@k-wall This morning, I want to quickly confirm whether or not we need to delete the per address settings, before re-adding them for the update.  I saw some behaviour last night, that suggests the delete this might be problematic.  But could be simply solved, if an add just overwrites.
  Driving it from the `status.planStatus` is an interesting idea.  I'll have a look at that for the standard.  If it simplifies the code, that would be great.

k-wall:
@vbusch see my comment on #4612.  I think with the way you chain the creation of address settings before addresses, that route is more like to create the address.

vbusch:
Pushed the change with the system tests + refactoring to use the addresses.status.planStatus.
I broke the update on my final cleanup.  I believe it might be a change I made to BrokerController.prototype.on_connection_open or the addresses structure.  Will look at it first thing in the am.

vbusch:

systemtests/src/test/java/io/enmasse/systemtest/bases/plans/PlansTestBase.java
```diff
@@ -131,6 +144,65 @@ public void doTestUnknownAddressPlan(AddressSpace addressSpace, List<StageHolder
         } while(true);
     }
 
+    public void doTestUpdatePlanBrokerCredit_changesPerAddressMaxSize(AddressSpace addressSpace, Address queueDest, AddressPlan phase1, AddressPlan phase2, AddressPlan redefinedPhase2, AmqpClientFactory amqpClientFactory) throws Exception {
```
kornys, line 147: Please use java camel syntax instead of `_`

systemtests/src/test/java/io/enmasse/systemtest/isolated/plans/brokered/PlansTestBrokered.java
```diff
@@ -149,4 +152,75 @@ void testUnknownAddressPlan() throws Exception {
         doTestUnknownAddressPlan(addressSpace, stageHolders);
     }
 
+    @Test
+    void testUpdatePlanBrokerCredit_changesPerAddressMaxSize() throws Exception {
```
kornys, line 156: same here please avoid using `_`

systemtests/src/test/java/io/enmasse/systemtest/isolated/plans/standard/PlansTestStandard.java
```diff
@@ -1168,6 +1168,88 @@ void testUnknownAddressPlan() throws Exception {
         doTestUnknownAddressPlan(addressSpace, stageHolders);
     }
 
+    @Test
+    void testUpdatePlanBrokerCredit_changesPerAddressMaxSize() throws Exception {
```
kornys, line 1172: same here please avoid using `_` and also add this test into acceptance profile then we can see behavior on ocp4

vbusch:
Some initial times for creating queues
```
				PR	Master
create 2500 queue:		7m30s	6m59s
create additional 1:		55s	58s
create additional 2500:		15:30	5:52+(missed recording the time where they were all in ready state)	
create additional 1:		2m15s	2m32s
delete all queues		14m56s	20m2s	

```
Recorded on a 3.11 cluster

vbusch:

systemtests/src/test/java/io/enmasse/systemtest/bases/plans/PlansTestBase.java
k-wall: Fixed by @vbusch

PR #4614, opened 2020-05-26
Add validations to metadata and fix review comments

anukritijha:
### Type of change

Add validation on device registration fields.

- Enhancement / new feature

### Description

This PR adds the following features to metadata form

- Adds dynamic child rows, its key, values and types in metadata json
- Enables/Disables the value field in case of Array or Object
- Enables/Disables the child row addition button
- Removes the generic typeahead component to patternfly select component
- Adds validations to metadata fields

image

### Checklist

- [ ] Resolve all warnings
- [x] Update license

### DIFF
--- a/console/console-init/ui/src/modules/device/dialogs/CreateDevice/MetaDataRow.tsx
+++ b/console/console-init/ui/src/modules/device/dialogs/CreateDevice/MetaDataRow.tsx
@@ -11,108 +11,172 @@ import {
   Button,
   DropdownPosition,
   TextInput,
-  DropdownItem,
-  SelectOptionObject
+  SelectOptionObject,
+  SelectVariant,
+  Select,
+  SelectOption
 } from "@patternfly/react-core";
-import { TypeAheadSelect, DropdownWithToggle } from "components";
+import { DropdownWithToggle } from "components";
 import { PlusIcon, MinusCircleIcon } from "@patternfly/react-icons";
-import { ISelectOption } from "utils";
+import { deviceRegistrationTypeOptions, getLabelByValue } from "modules/device";
+import { DataType } from "constant";
 
 export interface IMetaDataRow {
-  onChangePropertyInput?: (value: string) => Promise<any>;
-  formData: any;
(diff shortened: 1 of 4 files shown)


console/console-init/ui/src/modules/device/utils/constant.ts
```diff
@@ -75,11 +75,30 @@ const getInitialFilter = () => {
 const HIDE_ADVANCE_SETTING = "Hide advance setting";
 const SHOW_ADVANCE_SETTING = "Show advanced setting";
 
+const deviceRegistrationTypeOptions: ISelectOption[] = [
```
drcoolsanjeev, line 78: Date is missing

anukritijha: Noted.

console/console-init/ui/src/modules/device/dialogs/CreateDevice/MetaDataRow.tsx
```diff
@@ -11,108 +11,172 @@ import {
...
-    let metadataList = [...formData];
-    // metadataList[index][name]=value;
-    setFormData(metadataList);
+  const handleValueChange = (propertyValue: string, e: any) => {
+    let updatedValueMetadata = [...metadataList];
+    updatedValueMetadata[rowIndex].value = propertyValue;
+    setMetadataList(updatedValueMetadata);
   };
 
   const onPropertySelect = (e: any, selection: SelectOptionObject) => {
```
ajaypratap003, line 107: I think we can have function/event name onSelectProperty, onClearProperty. event name should come first then postfix. 
onSelect, onClear, onChange etc. wdyt?

anukritijha: Agreed.

anukritijha:
> The handleDeleteRow method in MetaDataRow has a bug, we need to pass the index. Currently the event object is being passed which is deleting the first row and not the selected row.

Thanks for pointing this out, will fix this.

PR #4625, opened 2020-05-27
Resolve build warnings on Console UI

anukritijha:
### Type of change

- Refactoring

### Description

Remove Warnings from Console UI


anukritijha:
@enmasse-ci run tests testcase=FirefoxConsoleTest

console/console-init/ui/src/modules/address-space/dialogs/CreateAddressSpace/Configuration.tsx
```diff
@@ -147,7 +147,7 @@ export const Configuration: React.FunctionComponent<IConfiguration> = ({
                 description:
                   plan.spec.shortDescription || plan.spec.longDescription
               };
-            }
+            } else return {};
```
ajaypratap003, line 150: Not sure, why did you add else return {}, I'm asuming that this is added for remove warning.
It might be cause to add empty object {} in addressSpacePlans in case else condition execute. Can you please verify this case what will happned if else part execute.

PR #4633, opened 2020-05-28
Allow specifying CPU limits in the infra configs

lulf:
* Add cpu field to admin, broker and router resources spec.
* Remove hardcoding of values in template, rely on system defaults if no
  settings are provided
* Set CPU and Memory requests and limits if defined in infra config.
* Allow overriding via pod templates


lulf:

kornys:
@lulf I think this is require also CSV change, or not?

lulf:
@kornys I don't think so. The build fails on some unit test code I will fix

lulf:

api-model/src/main/java/io/enmasse/admin/model/v1/BrokeredInfraConfigSpecAdminResources.java
```diff
@@ -19,44 +19,47 @@
...
+    private String cpu;
 
-    public BrokeredInfraConfigSpecAdminResources() {
+    public void setMemory(String memory) {
+        this.memory = memory;
     }
 
-    public BrokeredInfraConfigSpecAdminResources(final String memory) {
-        setMemory(memory);
+    public String getMemory() {
```
k-wall, line 32: methods reordered

k-wall:
@lulf code looks good - just a niggle about the delta and the question about scope.

lulf:
@k-wall console and authservice were already supporting CPU limits afaict. Added ability to set for address space controller

pkg/controller/address_space_controller/controller.go
```diff
@@ -212,19 +212,24 @@ func ApplyDeployment(deployment *appsv1.Deployment) error {
...
-		if err != nil {
-			return err
+		cpuEnv, ok := os.LookupEnv("ADDRESS_SPACE_CONTROLLER_CPU_LIMIT")
+		if ok {
+			cpuLimit, err := resource.ParseQuantity(cpuEnv)
+			if err != nil {
+				return err
+			}
+			container.Resources.Requests[corev1.ResourceCPU] = cpuLimit
+			container.Resources.Limits[corev1.ResourceCPU] = cpuLimit
```
k-wall, line 222: if the user removes the ADDRESS_SPACE_CONTROLLER_CPU_LIMIT var, I think the existing container would still contain the stale values.  we should remove them on the false path.

k-wall:
> @k-wall console and authservice were already supporting CPU limits afaict. Added ability to set for address space controller

agreed, I see console and standard are covered, but I think they should also include an else branch to cover the case where the resources are removed from the CRD.

```
		if authservice.Spec.Standard.Resources != nil {
			container.Resources = *authservice.Spec.Standard.Resources
		} else {
			container.Resources = corev1.ResourceRequirements{}
		}
```

none auth service omits to use its  `authservice.Spec.None.Resources`.  Might as well fix...

pkg/controller/address_space_controller/controller.go
lulf: Fixed.

lulf:

PR #4639, opened 2020-05-28
Docs:Add procedure for viewing endpoint console

hmaclean73:
### Type of change

- Documentation

### Description

Docs: Add procedure for viewing endpoint console, update text for creating an address space and fix table heading placement.
### Checklist

- [x] Update documentation
- [ ] Update CHANGELOG.md


documentation/modules/proc-create-address-space-console.adoc
```diff
@@ -5,17 +5,16 @@
 [id='proc-create-address-space-console-{context}']
 = Creating an address space using the {ConsoleName}
 
-You can create a new address space using the {ConsoleName}.
+You can create a new address space using the {ConsoleName}, including configuring the endpoints of an address space using different certificate providers and creating endpoints to make the address space available for messaging applications to use.
```
k-wall, line 8: FYI if the user choses not to configure the endpoints, the system creates a default set (which are probably good for many use-cases).

documentation/modules/ref-view-endpoint-info.adoc
```diff
@@ -0,0 +1,57 @@
...
+.Messaging endpoint information reference table
+[cols="50%a,50%a",options="header"]
+|===
+|Column |Description
+|Name |Displays the name of the endpoint.
+|Type |Displays the type of endpoint. Valid values include:
+
+`cluster`:: An endpoint accessible on the {KubePlatform} cluster through a cluster IP address.
+
+`route`:: Creates a cluster endpoint available through the {KubePlatform} route.
```
k-wall, line 31: surround `route` with the if openshift idiom so it is excluded from the Kubernetes docs.

k-wall, line 31:
```suggestion
`route`:: An {KubePlatform} route available outside the cluster.
```

documentation/modules/ref-view-endpoint-info.adoc
```diff
@@ -0,0 +1,57 @@
...
+
+`loadbalancer`:: Creates a cluster endpoint available through the `LoadBalancer` service.
```
k-wall, line 33:
```suggestion
`loadbalancer`:: A `LoadBalancer` service integrating with an external load balancer.
```

documentation/modules/ref-view-endpoint-info.adoc
```diff
@@ -0,0 +1,57 @@
...
+
+ifeval::["{cmdcli}" == "oc"]
+For more information, see the following {KubePlatform} documentation:
+
+* link:https://docs.openshift.com/container-platform/4.4/networking/routes/route-configuration.html[Route Configuration]
```
k-wall, line 38: Let's use vars for the urls.  That way it'll be less work when there is a version bump.

documentation/modules/proc-create-address-space-console.adoc
hmaclean73: Should I add a note to this section stating that the system creates a default set if no endpoints are configured?

k-wall: Yes, I think this is a good idea.  On OpenShift, by the default will give the user openshift routes for AMQPS and AMQP WSS, secured with a system generated (self-signed) certificate and a cluster service.  On Kubernetes you get the cluster service only.

hmaclean73: Are the following notes sufficient?

**OpenShift note**:
If you choose not to configure the endpoints for your address space, the system creates a default set of endpoints as OpenShift routes for AMQPS and AMQP-WSS, secured with a system generated (self-signed) certificate and a cluster service.
**Kubernetes note**:
If you choose not to configure the endpoints for your address space, the system creates a default set of cluster service endpoints.

k-wall: For kubernetes, the default is a (single) cluster service.

PR #4646, opened 2020-05-29
Fix K8s IoT deployment

dejanb:
### Type of change

- Bugfix

### Description

Fixes currently broken K8s IoT getting started guide


dejanb:
In my cluster this default token works we expected. I'll check what's going on in minishift next.

PR #4675, opened 2020-06-03
add-routes

drcoolsanjeev:
### Type of change
- Discussion
- New feature
- Refactoring

### Description

Add routes path for iot screens

### DIFF
--- a/console/console-init/ui/src/Routes.tsx
+++ b/console/console-init/ui/src/Routes.tsx
@@ -17,6 +17,46 @@ const getConnectionDetail = () =>
   import("modules/connection-detail/ConnectionDetailPage");
 
 export const AppRoutes = () => (
+  // <SwitchWith404>
+  //   {/* Redirect route */}
+  //   <Redirect path="/" to="/projects" exact={true} />
+  //   {/* messaging project list page route */}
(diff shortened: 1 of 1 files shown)


console/console-init/ui/src/Routes.tsx
```diff
@@ -17,6 +17,46 @@ const getConnectionDetail = () =>
...
+  //   />
+  //   {/* iot-project detail page route */}
+  //   <LazyRoute
+  //     path="/iot-projects/:namespace/:projectname/:sublist"
+  //     getComponent={getIoTProjectDetailPage}
+  //     exact={true}
+  //   />
+  //   {/* iot-project device detail page route */}
+  //   <LazyRoute
+  //     path="/iot-projects/:namespace/:projectname/devices/:deviceid/:sublist"
```
ajaypratap003, line 55: do we need to add devices in path ?
It can be path="/iot-projects/:namespace/:projectname/:deviceid/:sublist"

drcoolsanjeev: I think yes.just like the path for address detail path 
It is path="/address-spaces/:namespace/:name/:type/addresses/:addressname"

PR #4681, opened 2020-06-03
Ensure that memory settings are applied if minReplicas is not specified

lulf:
Fixes the infra tests.


lulf:

lulf:
Test failures seem unrelated to this change (image pull backoff, probably quay.io instability)

PR #4686, opened 2020-06-04
Set maxUnavailable=1 as default since it is not preventing a single b…

lulf:
…roker from being upgraded

### DIFF
--- a/templates/example-plans/020-StandardInfraConfig-default-minimal.yaml
+++ b/templates/example-plans/020-StandardInfraConfig-default-minimal.yaml
@@ -9,6 +9,7 @@ spec:
     resources:
       memory: 512Mi
   broker:
+    maxUnavailable: 1
     resources:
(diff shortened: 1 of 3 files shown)


k-wall:
> …roker from being upgraded

@lulf could you expand on why this becomes an issue now?  what use case is affected?

lulf:
@k-wall Originally this was added to ensure that when upgrading Kubernetes when you have HA queues (>= 2 brokers), the PodDisruptionBudget created would ensure that  brokers got moved one at a time to maintain availability. However, some time ago there was a new systemtest added that would drain all nodes and run into an issue with the last node drain to fail, so I removed this based on an assumption that this was the cause. However, after testing it manually I found that it handles the case where maxUnavailable: 1 and having 0 or 1 broker (i.e. it will still allow pod to be evicted), so I think we can safely add it back.

PR #4701, opened 2020-06-08
Delete enmasse in case installation failed

kornys:
Delete enmasse in case beforeall install fails, this will prevent situation where enmasse is in half installed state.


kornys:
@lulf enmasse OLM uninstall and bundle uninstall is almost same :)

PR #4702, opened 2020-06-08
Fix deployment of secrets for plain Kubernetes

ctron:
### Type of change

- Bugfix

### Description

This fixes the missing secrets for the device registry, when testing on plain kubernetes.

### Checklist

- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/api-model/src/main/java/io/enmasse/iot/model/v1/DeviceRegistryServiceConfig.java
+++ b/api-model/src/main/java/io/enmasse/iot/model/v1/DeviceRegistryServiceConfig.java
@@ -23,6 +23,7 @@ import io.sundr.builder.annotations.Inline;
 public class DeviceRegistryServiceConfig {
     private InfinispanDeviceRegistry infinispan;
     private JdbcDeviceRegistry jdbc;
+    private ManagementConfig management;
 
     public InfinispanDeviceRegistry getInfinispan() {
         return infinispan;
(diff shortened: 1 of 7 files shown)


ctron:
@enmasse-ci run tests ocp4 profile=iot testcase=io.enmasse.systemtest.iot.shared.http.HttpAdapterTest

kornys:
@ctron if you use beforeall, you need to apply same workaround as we apply here -> https://github.com/EnMasseProject/enmasse/blob/f3eed1fbedf3250baa4f6b18944407d47638c7d7/systemtests/src/main/java/io/enmasse/systemtest/listener/JunitCallbackListener.java#L96 where you need to catch exception and raise it in before each. it is a workaround until they release new surefire plugin version  with fix.

kornys:
@ctron and also you triggered jenkins with wrong test, seem io.enmasse.systemtest.iot.shared.http.HttpAdapterTest no longer exists

ctron:
> @ctron and also you triggered jenkins with wrong test, seem io.enmasse.systemtest.iot.shared.http.HttpAdapterTest no longer exists

Indeed … it got moved recently. Still, everything is green :grin:

ctron:
> @ctron if you use beforeall, you need to apply same workaround as we apply here ->
> 
> https://github.com/EnMasseProject/enmasse/blob/f3eed1fbedf3250baa4f6b18944407d47638c7d7/systemtests/src/main/java/io/enmasse/systemtest/listener/JunitCallbackListener.java#L96
> where you need to catch exception and raise it in before each. it is a workaround until they release new surefire plugin version with fix.

Is that really necessary? This is an interface, extending a test case. But not a "test extension". And the `JunitCallbackListener` should still be in play.

kornys:
@ctron anyway if some exception is thrown in beforeall method, test is not triggered and also surefire report that as green because of issue in surefire plugin

ctron:
@kornys I added a aroundaround

kornys:
@ctron thank you, looks really good

PR #4707, opened 2020-06-09
Gracefully handle missing Prometheus CRDs in some cases

ctron:
If monitoring is disabled, the IoT reconciler tries to "delete" the prometheus CRs. However that will fail if Prometheus CRDs are not installed.

It would be possible to add another flag or check, however I think the `ENABLE_MONTORING` env-var should already be enough. If monitoring is enabled, the prometheus CRD must be installed, otherwise it is an error. If monitoring is disabled, deleting the CR due to a missing CRD can be ignored. If the user chooses to enable monitoring without having the CRDs installed, an error is shown. It is still possible to install prometheus CRDs afterwards. The reconciler will try again and succeed the next time, as there are no stale caches. Using the `HasApi` methods would cache the outcome, which is ok for system resources, but wrong for CRDs.

In addition, the systemtests enable monitoring, although monitoring is not installed. This is fixed as well.

### Type of change

- Bugfix

### Checklist

- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/pkg/controller/iotconfig/monitoring.go
+++ b/pkg/controller/iotconfig/monitoring.go
@@ -10,10 +10,13 @@ import (
 	promv1 "github.com/coreos/prometheus-operator/pkg/apis/monitoring/v1"
 	iotv1alpha1 "github.com/enmasseproject/enmasse/pkg/apis/iot/v1alpha1"
 	"github.com/enmasseproject/enmasse/pkg/monitoring"
+	"github.com/enmasseproject/enmasse/pkg/util"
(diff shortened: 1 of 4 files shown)


ctron:
@enmasse-ci run tests ocp4 profile=iot testcase=io.enmasse.systemtest.iot.isolated.http.HttpAdapterTest

kornys:
AFAIK we enable monitoring only in scope of monitoring tests

lulf:
@kornys It is set to "true" by default in the operator deployment yaml.

PR #4708, opened 2020-06-09
Replace CSV image env at build time

lulf:


kornys:
@enmasse-ci run tests profile=olm crc

PR #4709, opened 2020-06-09
Fix runUntilPass catch also assertion exception

kornys:
### Type of change

- Bugfix

### Description

Catch also AssertionError

### Checklist

- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/systemtests/src/main/java/io/enmasse/systemtest/utils/TestUtils.java
+++ b/systemtests/src/main/java/io/enmasse/systemtest/utils/TestUtils.java
@@ -350,7 +350,7 @@ public class TestUtils {
             try {
                 log.debug("Running command, attempt: {}", i);
                 return fn.call();
-            } catch (Exception ex) {
+            } catch (Exception | AssertionError ex) {
                 log.warn("Command failed", ex);
             }
             Thread.sleep(1000);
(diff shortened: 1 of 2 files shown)


systemtests/src/main/java/io/enmasse/systemtest/utils/TestUtils.java
```diff
@@ -350,7 +350,7 @@ public static void waitForNamespaceDeleted(Kubernetes kubernetes, String namespa
             try {
                 log.debug("Running command, attempt: {}", i);
                 return fn.call();
-            } catch (Exception ex) {
+            } catch (Exception | AssertionError ex) {
```
lulf, line 353: Won't this actually mask errors?

kornys: I don't think so, the reason why I did this, is runUntullPass does not catch exception in method passed into because Exception is not a parent of AssertionError

lulf: If there is an error and it will still be flagged at the last failure, thats fine I suppose.

kornys: yes

PR #4710, opened 2020-06-09
Add upgrade pre-check when installing via ansible

### LINKED ISSUE #4705: Upgrade to 0.31.z with address space created by service broker in a non-existent namespace fails
**Describe the bug**
When upgrading from EnMasse 0.30.z or older to 0.31.z or newer, and having an address space created via the service-broker, it was possible to specify a non-existing namespace for the service instance. If such an invalid address space have been created, the upgrade may fail with the address-space-controller not being able to convert an address space from the ConfigMap representation to the AddressSpace CRD representation due to conflict: already exists.

image

lulf:
Fixes #4705

### DIFF
--- a/ansible/playbooks/openshift/precheck.yml
+++ b/ansible/playbooks/openshift/precheck.yml
@@ -1,0 +1,4 @@
+- name: "Upgrade precheck"
+  hosts: enmasse
+  roles:
+    - upgrade_precheck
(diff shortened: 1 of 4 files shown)


kornys:
@enmasse-ci run tests profile=upgrade testcase=UpgradeTest#testAnsiblePrecheck test_only

PR #4723, opened 2020-06-11
Ensure that addresses_ready event does not include addresses belonging to other address spaces.

k-wall:
### Type of change

- Bugfix

### Description

addresses_ready event erroneously included addresses for other addressspaces colocated in the same namespace.

### Checklist

- [ ] Update documentation
- [x] Update CHANGELOG.md

### DIFF
--- a/agent/lib/internal_address_source.js
+++ b/agent/lib/internal_address_source.js
@@ -21,6 +21,7 @@ var events = require('events');
 var kubernetes = require('./kubernetes.js');
 var log = require('./log.js').logger();
 var myutils = require('./utils.js');
+var clone = require('clone');
 
 function extract_spec(def, env) {
     if (def.spec === undefined) {
@@ -33,7 +34,7 @@ function extract_spec(def, env) {
     o.addressSpaceNamespace = def.metadata ? def.metadata.namespace : process.env.ADDRESS_SPACE_NAMESPACE;
 
     if (def.status && def.status.brokerStatuses) {
-        o.allocated_to = def.status.brokerStatuses;
+        o.allocated_to = clone(def.status.brokerStatuses);
     }
     return o;
 }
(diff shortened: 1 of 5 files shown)


k-wall:

agent/lib/internal_address_source.js
```diff
@@ -193,7 +193,7 @@ AddressSource.prototype.updated = function (objects) {
     if (changes) {
         this.update_readiness(changes);
         this.dispatch('addresses_defined', addresses, changes.description);
-        this.dispatch_if_changed('addresses_ready', objects.filter(ready).map(extract_spec), same_address_definition);
+        this.dispatch_if_changed('addresses_ready', addresses.filter(ready), same_address_definition);
```
lulf, line 196: I'm worried this might cause a regression, where the entries in the addresses array would have their state mutated, causing the readiness to be wrong (I vaguely remember or imagining fixing an issue like that).

lulf: I've looked through the commit history, and I think i was confusing this with some other fixes that I made to the readiness logic. I don't see this affecting it so if tests passes, +1.

k-wall: @lulf  I  don't see a problem, but I don't find it particularly easy code to work with.  What is it that mutates the address record?   I don't think its this.update_readiness() and it maintains a separate structure (readiness - which is used only by brokered).

The cowardly approach would be to repeat the filter, which I can do if preferred.

k-wall:

k-wall:
@lulf you were right, this change did expose another issue.   The durable subscription tests (such as ChromeConsoleTest#testCreateDeleteDurableSubscription) were consistently failing.  The issue came down to a live ref to the broker allocation being store in the `last` view.  This meant that the address wasn't considered ready when the allocation had in fact changed.

k-wall:

k-wall:

PR #4734, opened 2020-06-15
Fix for issue-4612

vbusch:
### Type of change
- Bugfix

### Description
When the router creates a queue on the broker it should be durable

### Checklist

- [ ] Update documentation
- [ ] Update CHANGELOG.md

Signed-off-by: Vanessa Busch <vbusch@redhat.com>

### DIFF
--- a/agent/lib/router_config.js
+++ b/agent/lib/router_config.js
@@ -374,7 +374,7 @@ function desired_address_config(high_level_address_definitions, brokers) {
     for (var i in high_level_address_definitions) {
         var def = high_level_address_definitions[i];
         if (def.type === 'queue') {
-            config.add_address({prefix:def.address, distribution:'balanced', waypoint:true});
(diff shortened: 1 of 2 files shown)


k-wall:
@vbusch rather than the approach you've taken (which I'm not sure would address the whole issue), I think we need to organise that agent always creates the broker resources.   @lulf gave me some ideas on depriving the router messaging user of the create permissions in order to prevent this defect happing (which is the way he's solved the problem in shared).  I'll experiment with this approach this morning.

vbusch:
@k-wall great.  I had tested it with the wrong image.  And was thrilled when everything worked.  Even after I started removing extra lines that I thought I didn't need.  After creating the PR, I realized my mistake, and removed you as reviewer and switched it to draft, because I was back where I started.

vbusch:
This won't work.  Replaced by #4739

PR #4735, opened 2020-06-15
Add version and commit revision metadata to operator metadata

lulf:
This allows the operator to be upgraded by applying the new templates,
which will also update any operands like the address space controller,
authentication service and console.

### DIFF
--- a/Makefile.common
+++ b/Makefile.common
@@ -35,7 +35,7 @@ buildpush:
 	$(MAKE) docker_push
 
 docker_build: package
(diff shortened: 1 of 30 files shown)


lulf:

lulf:

lulf:

lulf:
@k-wall We could do that, but it would mean that if a new image for address-space-controller was created with a major.minor tag (but would contain different revision), and it was restarted with the new revision, it would potentially end up auto-restart the router and broker as well, removing the 'choice' of when to do the restart of the messaging infrastructure. However, from what I can tell, it would not apply the new templates unless the infra configs or address space have changed, so it could be it would work as before.

k-wall:
Sorry, I take back my suggestion - I think it makes for confusion.  The revision and version will only ever correspond to the operator's build, this is fine for address space controller/console/authservice since the operator is directly responsible for these deployments.  This is not the case where the admin/route/broker.  If we were to say the rev address space controller image or the standard-controller image. the propagated information would be stale.  It could lead someone to conclude that then update hadn't worked.

I suggest popping off the last commit. Sorry!

PR #4739, opened 2020-06-15
Prevent the router from racing with the agent and causing addresses/queues to be autocreated.

k-wall:
### Type of change
- Bugfix

### Description

Prevent the router from racing with the agent and causing ddresses/queues to be autocreated.

Uses @lulf's idea to assign the router's user a role that denies it create/delete address and create/delete durable queue.  This means if the router attaches its link before the agent has done its work, the Broker's autocreate will fail to cause the address/queue to come into existence with the wrong properties.

### Checklist

- [ ] Update documentation
- [ ] Update CHANGELOG.md


k-wall:

PR #4741, opened 2020-06-15
Set global address space plan for create address buttons

anukritijha:
### Type of change

- Bugfix

### Description

Populate Plan dropdown on Create Address Wizard from Empty Address's Create Button.

image


lulf:

PR #4742, opened 2020-06-15
Feat: wizards for device list page

rkpattnaik780:
### Type of change

- New feature

### Description

Wizards for bulk operations and delete operations. Show NoResult component if no data is obtained after applying filters.

### Checklist

- [ ] Resolve all warnings
- [x] Update license


console/console-init/ui/src/modules/iot-device/utils/constant.ts
```diff
@@ -97,6 +97,12 @@ const getLabelByValue = (typeValue: string) => {
   return filteredType.label as string;
 };
 
+enum DialogTypes {
```
anukritijha, line 100: We should also add Update/Edit here and make this component generic across all modules.

rkpattnaik780: Okay, that sounds fair.

anukritijha:
@rkpattnaik780 Please resolve the warnings, also if some variables are kept for future use, please comment them.

PR #4758, opened 2020-06-17
Enforce not treating target as directory

lulf:


Makefile.common
```diff
@@ -5,12 +5,12 @@ MOCHA_FILE      ?= build/test-results/test/TEST-$(PROJECT_NAME).xml
...
 MAVEN_BATCH     ?= true
 UNAME_S := $(shell uname -s)
-LN=ln
+LN=ln -T
 BUILD_GOOS=linux
 BUILD_GOARCH=amd64
 
 ifeq ($(UNAME_S),Darwin)
-	LN = gln
+	LN = gln -h
```
kornys, line 13:
```
gln -h                                                                                                        
gln: invalid option -- 'h'
Try 'gln --help' for more information.
```

lulf: Sorry, I missed the fact that it switches to GNU ln on mac

templates/Makefile
```diff
@@ -111,7 +111,7 @@ component_install: replace_images
 
 ansible_install: component_install
 	cp -r $(PACKAGE_ANSIBLE_DIR) $(INSTALLDIR)/
-	$(LN) -srf $(INSTALLDIR)/install/components $(INSTALLDIR)/ansible/playbooks/openshift/components
+	$(LN) -srfT $(INSTALLDIR)/install/components $(INSTALLDIR)/ansible/playbooks/openshift/components
```
kornys, line 114: also does not work on mac :)
```
 ln -srfT pokus pokus2                                                                                      
ln: illegal option -- r
usage: ln [-Ffhinsv] source_file [target_file]
       ln [-Ffhinsv] source_file ... target_dir
       link source_file target_file

ln --help                                                                                                 
ln: illegal option -- -
usage: ln [-Ffhinsv] source_file [target_file]
       ln [-Ffhinsv] source_file ... target_dir
       link source_file target_file

     ~  gln --help                                                                                                    1 ↵  default/10-37-144-48:8443/developer ⎈  15:23:12 
Usage: gln [OPTION]... [-T] TARGET LINK_NAME
  or:  gln [OPTION]... TARGET
  or:  gln [OPTION]... TARGET... DIRECTORY
  or:  gln [OPTION]... -t DIRECTORY TARGET...
In the 1st form, create a link to TARGET with the name LINK_NAME.
In the 2nd form, create a link to TARGET in the current directory.
In the 3rd and 4th forms, create links to each TARGET in DIRECTORY.
Create hard links by default, symbolic links with --symbolic.
By default, each destination (name of new link) should not already exist.
When creating hard links, each TARGET must exist.  Symbolic links
can hold arbitrary text; if later resolved, a relative link is
interpreted in relation to its parent directory.

Mandatory arguments to long options are mandatory for short options too.
      --backup[=CONTROL]      make a backup of each existing destination file
  -b                          like --backup but does not accept an argument
  -d, -F, --directory         allow the superuser to attempt to hard link
                                directories (note: will probably fail due to
                                system restrictions, even for the superuser)
  -f, --force                 remove existing destination files
  -i, --interactive           prompt whether to remove destinations
  -L, --logical               dereference TARGETs that are symbolic links
  -n, --no-dereference        treat LINK_NAME as a normal file if
                                it is a symbolic link to a directory
  -P, --physical              make hard links directly to symbolic links
  -r, --relative              create symbolic links relative to link location
  -s, --symbolic              make symbolic links instead of hard links
  -S, --suffix=SUFFIX         override the usual backup suffix
  -t, --target-directory=DIRECTORY  specify the DIRECTORY in which to create
                                the links
  -T, --no-target-directory   treat LINK_NAME as a normal file always
  -v, --verbose               print name of each linked file
      --help     display this help and exit
      --version  output version information and exit

The backup suffix is '~', unless set with --suffix or SIMPLE_BACKUP_SUFFIX.
The version control method may be selected via the --backup option or through
the VERSION_CONTROL environment variable.  Here are the values:

  none, off       never make backups (even if --backup is given)
  numbered, t     make numbered backups
  existing, nil   numbered if numbered backups exist, simple otherwise
  simple, never   always make simple backups

Using -s ignores -L and -P.  Otherwise, the last option specified controls
behavior when a TARGET is a symbolic link, defaulting to -L.

GNU coreutils online help: <https://www.gnu.org/software/coreutils/>
Full documentation <https://www.gnu.org/software/coreutils/ln>
or available locally via: info '(coreutils) ln invocation'
```

lulf: @kornys I'm not sure I understand... the gln seems to work, if you look in Makefile.common, you can see that LN=gln on mac.

PR #4759, opened 2020-06-17
[systemtests] Check if resources are removed

kornys:
### Type of change

- Bugfix

### Checklist


kornys:
@enmasse-ci run tests profile=isolated test_only

PR #4760, opened 2020-06-17
Fail in case iot cannot deploy certs

kornys:
### Type of change

- Bugfix

### Description

Currently if create/deploy script fail system ignores that and continue, so this should prevent it.

### Checklist


kornys:
@enmasse-ci run tests profile=iot testcase=iot.isolated.http.HttpAdapterTest crc test_only

PR #4763, opened 2020-06-17
Ensure connection and senders are closed before resetting state

lulf:


lulf:

standard-controller/src/main/java/io/enmasse/controller/standard/AddressProbeClient.java
```diff
@@ -144,32 +144,36 @@ public void connect(String host, int port, ProtonClientOptions clientOptions, Co
 
     @Override
     public void close() throws Exception {
+        CountDownLatch latch = new CountDownLatch((connection != null ? 1 : 0) + senders.size() + receivers.size());
```
k-wall, line 147: There's a better way.  `runOnContext` returns a completable future.  Gather these and use a CompletableFuture.allOf().  It is more succinct and handles the error case where sender.close() throws an exception.

lulf: Hmm, isn't that what it was doing already? I something attempting to use the ProtonConnection after it had been set to null, and I thought that it didn't actually wait for the close to finish. Then it must be another race here...

PR #4790, opened 2020-06-22
Iot cert auth shared infra

lulf:
This PR creates an internal listener + service for the shared infra that can be used by cluster-internal components such as iot adapters (it could also reuse the same port as the operator management, but it feels cleaner to separate that). 

The second commit that changes the iotconfig controller is just an example on how you can reconcile the client certificate and get the information that you need to inject into the adapters. I'm not sure where you think this should belong, but this is anyway how it would look. I can split out the first part in a PR if you think this approach looks good.

### DIFF
--- a/pkg/controller/iotconfig/auth.go
+++ b/pkg/controller/iotconfig/auth.go
@@ -7,7 +7,11 @@ package iotconfig
 
 import (
 	"context"
+	"fmt"
+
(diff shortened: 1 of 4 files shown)


ctron:
This look great, and simple! I agree that this should have its own endpoint. Let's go ahead with this proposal. I will take a look at Hono once we merged to 1.3, so I can check out if we can directly connect Hono, or need to go through QDR.

lulf:

lulf:

PR #4806, opened 2020-06-23
Fix iot ALM examples in CSV

kornys:
Fix alm examples to deploy iot without any other secrets needed.


kornys:
I have to test it before we merge it :)

kornys:
@lulf @ctron do I need to update also spec.customresourcedefinitions in CSV? Because I see only count of adapters and not enabled field is specified there

kornys:
and I think that this is a way how the new form UI on ocp4.5 works, it reads values from spec.customresourcedefinitions

PR #4807, opened 2020-06-23
Fix: undesired behavior of CreateAddress wizard

rkpattnaik780:
### Type of change

- Bugfix

### Description

This PR fixes the two bugs reported in the Create address wizard (changing plan from subscription to any other type does not fetch the respective plan and changing type for brokered address creation does not get new plan options). This was arising due to the asynchronous state changes made in the CreateAddressWizard. If we had a type which has only one plan, switching from it made a re-render with old planOptions. If only one element is passed as options in DropdownWithToggle, it selects it as default value. This has been fixed by a simple re-order for state change, that unselects the plan upon type change.

### Checklist

- [ ] Resolve all warnings
- [ ] Update license


console/console-init/ui/src/modules/address/containers/AddressDefinitionContainer/AddressDefinitionContainer.tsx
```diff
@@ -106,9 +106,9 @@ export const AddressDefinitionContainer: React.FunctionComponent<IAddressDefinit
             description: plan.spec.shortDescription || plan.spec.longDescription
           };
         });
-        setPlan(" ");
         setTopic(" ");
         setPlanOptions(planOptions);
+        setPlan(" ");
```
ajaypratap003, line 111: is there any reasoing of passing space in states functions?

rkpattnaik780: Yeah, I have not checked for this PF version, but earlier we had to give a space other wise the dropdown toggle was getting misplaced.

PR #4813, opened 2020-06-24
Propagate openshift detection result

lulf:


lulf:
@enmasse-ci run tests profile=smoke crc

pkg/controller/address_space_controller/controller.go
```diff
@@ -205,6 +205,9 @@ func ApplyDeployment(deployment *appsv1.Deployment) error {
 		applyImageEnv(container, "TOPIC_FORWARDER_IMAGE", "topic-forwarder")
 		applyImageEnv(container, "MQTT_GATEWAY_IMAGE", "mqtt-gateway")
 		applyImageEnv(container, "MQTT_LWT_IMAGE", "mqtt-lwt")
+		if util.IsOpenshift() {
+			applyImageEnv(container, "ENMASSE_OPENSHIFT", "true")
```
k-wall, line 209:
```suggestion
			applyImageEnv(container, util.EnMasseOpenshiftEnvVar, "true")
```

lulf:
@k-wall Even today, the operator takes a while to decide its not on openshift and produce a warning, I don't see how this is much different. It could be the case where it is in the process of attaching an aggregated API server, so a 404 could be worth retrying for.

lulf:
@k-wall I've removed the change to the java code. I agree just changing so that the operator sets the env is sufficient, and less chances of causing issues.

lulf:

PR #4816, opened 2020-06-24
Add support for cost field in connector

lulf:


lulf:
TODO: Systemtests

kornys:
I don't understand what this feature should do and how to test it

PR #4836, opened 2020-06-26
Remove non-1.0 modules

lulf:
* Remove modules and components not part of 1.0
* Update smoke test to run isolated-shared-infra profile

Templates are kept and will be updated in a separate PR. This means it
will still be possible to run the old version using public images.

Systemtests still remain, and will have to be updated separately.


templates/shared-infra/enmasse.io_messagingaddresses.yaml
```diff
@@ -4,7 +4,7 @@ apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.3.0
```
kornys, line 7: @lulf this will be neverending battle... someone will generate crds with version 0.3.0 and someone with 0.2.5

lulf: Sorry, I will fix my local setup :)

lulf: Fixed

ctron:
So, just to understand … this will mean that IoT tests will fail immediately and only come back green once we migrated to shared infra?!

kornys:
@ctron yes I think so, so after this PR I have to push my branch where tests are removed and iot tests are disabled

lulf:
@ctron yes, that was what I was thinking. I’m not sure there is any point in testing/building anything else than shared infra now that we have branches

systemtests/src/test/java/io/enmasse/systemtest/sharedinfra/MessagingAddressTest.java
```diff
@@ -34,10 +34,7 @@
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
-import java.util.concurrent.Callable;
-import java.util.concurrent.ForkJoinPool;
-import java.util.concurrent.Future;
-import java.util.concurrent.TimeUnit;
+import java.util.concurrent.*;
```
kornys, line 37: remove star imports

systemtests/src/test/java/io/enmasse/systemtest/sharedinfra/MessagingEndpointTest.java
```diff
@@ -42,9 +42,7 @@
 
 import java.util.Collections;
 import java.util.List;
-import java.util.concurrent.ForkJoinPool;
-import java.util.concurrent.Future;
-import java.util.concurrent.TimeUnit;
+import java.util.concurrent.*;
```
kornys, line 45: same here

PR #4842, opened 2020-06-26
Remove templates

lulf:
Next phase after #4836.

* Removes pre-1.0 templates
* Makes shared infra the 'main' bundle
* Adds IoT to main bundle (IMO no need to keep separate step for IoT when its just an integrated part in the operator anyway)


kornys:
I agree iot bundle should be part of enmasse bundle by default :-)

PR #4843, opened 2020-06-29
Try to run CI using kind

lulf:


obabec:
I would be happy to migrate to kind but I think if we move we should still run (for example once a week) master job also on minikube to be sure, that we have covered this platform too. WDYT? @lulf

lulf:
@obabec Yeah, another master workflow that doesn't push images to quay.io then?

obabec:
@lulf Exactly, I'll handle after kind if working, sounds reasonable for you?

obabec:
@lulf I think you are missing registry. That needs to be deployed too.

PR #4849, opened 2020-06-29
Remove metrics api and remove nonexisting project deps

lulf:


.github/scripts/build.sh
```diff
@@ -7,7 +7,7 @@ if [[ -v PUSH_REGISTRY ]]
 then
     echo "Logging in to registry"
     docker login -u "${REGISTRY_USER}" -p "${REGISTRY_PASS}" "${DOCKER_REGISTRY}"
-    make TAG="${VERSION}" buildpush
+    make buildpush
```
kornys, line 10: why we do not need specify tag?

lulf: Its a bug after my previous refactoring. Previously we would push images with both ${TAG} and ${VERSION}, and this was just a leftover from when we tried to use floating tags and were still using travis multiplexing different parameters in this biuld script. With github actions, I think we can safely just use the TAG specified in the action as the source of truth rather than doing extra manipulation here.

PR #4854, opened 2020-06-30
Use new XML syntax for loading metrics plugin

lulf:


k-wall:
Don't we need this on master too?

lulf:
@kornys I have only tested it on a standalone broker.

@k-wall I will raise a separate PR, because the code bases have diverged :)

PR #4855, opened 2020-06-30
Initial proposal for console backend 1.0

lulf:
Rendered version: https://github.com/EnMasseProject/enmasse/blob/console-1.0-proposal/documentation/design/proposals/console-1.0-backend.adoc


documentation/design/proposals/console-1.0-backend.adoc
```diff
@@ -0,0 +1,117 @@
...
+
+At the moment console-server queries connections/links from the address space agents.
+
+For 1.0, the agent disappears, so information must be gathered differently:
+
+* Deploy an EnMasse-owned ephemeral Prometheus instance as planned in the initial console proposal.
+* Configure prometheus to scrape router and broker pods
+* Console-server querying the routers directly to build the list of connections and their links
+* Console server querying the prometheus instance for address and connection metrics
+* Console-server eventually also becomes backend for the CLI.
```
k-wall, line 86: I wonder if we should put the exposing the graphql service (cURL with preemptive auth) as part of this release.  This will allow us to test some features directly against the graphql interface rather than writing selenium based tests.    This will allow us to reduce the number of slow selenium tests (slow both in terms of their writing and running).  We'll still need to make sure the whole UI is covered, but we should be able to do this with fewer tests.

documentation/design/proposals/console-1.0-backend.adoc
```diff
@@ -0,0 +1,117 @@
...
+Given the timelines, service admin console should not be in scope for 1.0
+
+=== Console API/UI changes
+
+==== AddressSpace => MessagingTenant
+
+image::console_1.0_overview.png[]
+
+* `AddressSpace` is removed and replaced by `MessagingTenant`, and there can now only be a single MessagingTenant named 'default' per Kubernetes namespace. Most of the properties like plan and authentication service (now identity provider) is now optional, and the minimum information needed to create a tenant is its namespace. In addition, a tenant may have a set of capabilities such as 'transactional'.
+
```
k-wall, line 41: When in the UI will the user find the IoTProjects?

documentation/design/proposals/console-1.0-backend.adoc
```diff
@@ -0,0 +1,117 @@
...
+For 1.0, the agent disappears, so information must be gathered differently:
+
+* Deploy an EnMasse-owned ephemeral Prometheus instance as planned in the initial console proposal.
+* Configure prometheus to scrape router and broker pods
+* Console-server querying the routers directly to build the list of connections and their links
+* Console server querying the prometheus instance for address and connection metrics
+* Console-server eventually also becomes backend for the CLI.
+* Future: use metrics from prometheus to drive router/broker scaling
+* Future: use prometheus to drive IoT/Hono scaling and request throttling
+
```
k-wall, line 89: Would prometheus federation be supported initially?  I.e. as a user would I be able to point my prometheus instance at EnMasse's instance and scrape metrics from it so that I may do my own analysis, configure graphana dashes etc.

documentation/design/proposals/console-1.0-backend.adoc
jbtrystram: the IoTProjects will be displayed in the same list as the MessagingTenants

documentation/design/proposals/console-1.0-backend.adoc
lulf: I think thats a reasonable future task, but not something we'll be able to do before 1.0

PR #4863, opened 2020-07-01
[Shared Infra] [Identity] Basic Access Control Server

k-wall:
### Type of change

- Enhancement / new feature

### Description

Create a basic access controller server that implements the AMQP SASL protocol.  This task will:

* put the mechanics of the AMQP SASL server in place
* implement the SASL ANONYMOUS mech (and grant access to all addresses)
* deploy an instance of accessControlServer per messaging infra
* configure the router listeners to delegate to the access controller server as an authServicePlugin

### Checklist

- [ ] Update documentation
- [ ] Update CHANGELOG.md


lulf:
@k-wall Regarding router <-> access control server, I was wondering if it would make sense to have the operator issue a cert to the access control server, and reuse the per-messaginginfra CA trust of the router when connecting to the access control server. IMO this would be more robust than relying on OpenShift specific vs . Kubernetes specific mechanisms for certificate management.

k-wall:
> @k-wall Regarding router <-> access control server, I was wondering if it would make sense to have the operator issue a cert to the access control server, and reuse the per-messaginginfra CA trust of the router when connecting to the access control server. IMO this would be more robust than relying on OpenShift specific vs . Kubernetes specific mechanisms for certificate management.

That makes sense to me.

templates/enmasse-operator/enmasse-operator.deployment.yaml
```diff
@@ -110,6 +110,8 @@ spec:
           value: ${CONSOLE_PROXY_OPENSHIFT_IMAGE}
         - name: RELATED_IMAGE_CONSOLE_PROXY_KUBERNETES
           value: ${CONSOLE_PROXY_KUBERNETES_IMAGE}
+        - name: RELATED_IMAGE_ACCESS_CONTROL_SERVER
+          value: ${ACCESS_CONTROL_SERVER_IMAGE}
```
lulf, line 114: @k-wall Note that the env needs to be set in the olm-manifest as well :)

k-wall:
@lulf I've got a problem with the tests that I'm still investigating, but my hope is that the pieces for this change are now in place.

k-wall:
@enmasse-ci run tests

k-wall:
@enmasse-ci run tests

go.mod
```diff
@@ -30,7 +30,7 @@ replace (
 	k8s.io/legacy-cloud-providers => k8s.io/legacy-cloud-providers v0.0.0-20191016115753-cf0698c3a16b
 	k8s.io/metrics => k8s.io/metrics v0.0.0-20191016113814-3b1a734dba6e
 	k8s.io/sample-apiserver => k8s.io/sample-apiserver v0.0.0-20191016112829-06bb3c9d77c9
-	pack.ag/amqp => github.com/interconnectedcloud/go-amqp v0.12.6-0.20200506124159-f51e540008b5 // indirect
+	pack.ag/amqp => github.com/k-wall/go-amqp v0.12.6-0.20200706113411-abc925d8ef14
```
lulf, line 33: Should the changes be submitted as a PR to interconnected cloud you think?

pkg/controller/messaginginfra/messaginginfra_controller.go
```diff
@@ -332,6 +337,7 @@ func (r *ReconcileMessagingInfra) Reconcile(request reconcile.Request) (reconcil
 		ready = infra.Status.GetMessagingInfrastructureCondition(v1.MessagingInfrastructureReady)
 		caCreated = infra.Status.GetMessagingInfrastructureCondition(v1.MessagingInfrastructureCaCreated)
 		certCreated = infra.Status.GetMessagingInfrastructureCondition(v1.MessagingInfrastructureCertCreated)
+		accessControlCreated = infra.Status.GetMessagingInfrastructureCondition(v1.MessagingInfrastructureCertCreated)
```
lulf, line 340: Wrong condition passed (CertCreated)

k-wall:
@enmasse-ci run tests

pkg/state/infra.go
```diff
@@ -86,16 +88,17 @@ type infraClient struct {
...
 
-func NewInfra(routerFactory routerStateFunc, brokerFactory brokerStateFunc, clock Clock) *infraClient {
+func NewInfra(i *v1.MessagingInfrastructure, routerFactory routerStateFunc, brokerFactory brokerStateFunc, clock Clock) *infraClient {
 	// TODO: Make constants and expand range
 	portmap := make(map[int]*string, 0)
 	for i := 40000; i < 40100; i++ {
 		portmap[i] = nil
 	}
 	client := &infraClient{
+		infrastructure:      i,
```
lulf, line 101: I see that only the name and namespace of the infrastructure is actually used. I wonder if it would be better to store just that, because this is not synced, so the rest of the infra could get stale if the user changes the spec for instance. Just to future proofing myself from looking at the infrastructure and thinking i can pull some data from the spec.

go.mod
k-wall: There's one raised already.

pkg/state/infra.go
k-wall: Thanks, I hadn't considered a staleness problem.  Fixed.

k-wall:
@enmasse-ci run tests

k-wall:
@enmasse-ci run tests

k-wall:
@enmasse-ci run tests

kornys:
@k-wall please new shared infra should be supported only on ocp4 so run your tests on crc :)

k-wall:
@enmasse-ci run tests ocp4

PR #4865, opened 2020-07-01
Fix issue converting map to sorted map

ctron:
### Type of change

- Bugfix

### Checklist

- [ ] Update documentation
- [ ] Update CHANGELOG.md


k-wall:
is this a bug fix to released functionality.  If so, it should at least a have a short description explaining how the defect impacts this user?

ctron:
> is this a bug fix to released functionality. If so, it should at least a have a short description explaining how the defect impacts this user?

No, it shouldn't have any impact on the user. I only noticed the issue, when copying the pattern for another functionality. And there it broke the logic. I still is wrong and fixing it seems the right thing to do.

PR #4868, opened 2020-07-02
Use same log format in all operator components

lulf:


pkg/state/infra.go
```diff
@@ -960,7 +962,7 @@ func (i *infraClient) buildRouterAddressEntities(endpoint *v1beta2.MessagingEndp
 	if tenant.Status.Broker != nil {
 		// Transactional means that we create a link route based on the tenant prefix,
 		// and all broker addresses go through that route.
-		log.Println("Tenant borkre set, using ")
+		log.Info("Tenant borkre set, using ")
```
kornys, line 965: typo

lulf: Fixed

PR #4875, opened 2020-07-02
Change API versions to v1

lulf:


kornys:
@lulf finally :) but iot won't be updated to v1?

lulf:
@kornys I was thinking to do that in another sweep to avoid too much conflict with ongoing work.

kornys:
@lulf Ok makes sense

PR #4881, opened 2020-07-03
Remove messaging func from test classes

obabec:
### Type of change

- Enhancement / new feature

### Description
Proposal how to get rid of messaging do test methods from testClasses. I am not 100% confident with naming, but this decomposition will also help us to get more useful logger messages with good format for debug (fe. [INFO] [testName -> MessagignClientRunner] Hello world). This patter could be applied for rest of test classes.

### Checklist

- [ ] Update documentation
- [ ] Update CHANGELOG.md


systemtests/src/main/java/io/enmasse/systemtest/messagingclients/ExternalMessagingClient.java
```diff
@@ -142,6 +142,13 @@ public String getId() {
         }
     }
 
+    public boolean isSender() {
+        ClientType clientType = this.client.getClientType();
```
kornys, line 146: Instead of checking this enum we should implement getCapability in clients, which returns SENDER/RECEIVER enum value and just check it.

systemtests/src/main/java/io/enmasse/systemtest/messagingclients/MessagingClientRunner.java
```diff
@@ -0,0 +1,111 @@
...
+        clients.add(senderClient);
+        assertTrue(senderResult.get(1, TimeUnit.MINUTES), "Sender failed, expected return code 0");
+        for (Future<Boolean> receiverResult : receiverResults) {
+            assertTrue(receiverResult.get(1, TimeUnit.MINUTES), "Receiver failed, expected return code 0");
+        }
+        return clients;
+    }
+
+
+    public int getPort(String protocol, MessagingEndpoint endpoint) {
```
kornys, line 103: I think this method is more common and it should be in MessagingEndpointResource class as a static

systemtests/src/main/java/io/enmasse/systemtest/messagingclients/ExternalMessagingClient.java
obabec: Agree, that was one of the things, that i wanted to discuss. There is also option to add some abstraction such as Sender/Receiver interface. But in this case I agree, get Capability could be best option.

systemtests/src/main/java/io/enmasse/systemtest/messagingclients/MessagingClientRunner.java
obabec: I've moved this to MessagingAdressResouce as static.

systemtests/src/main/java/io/enmasse/systemtest/messagingclients/MessagingClientRunner.java
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    /**
+     * Send 10 messages on sender address, and receive 10 messages on each receiver address.
+     */
+    public List<ExternalMessagingClient> sendAndReceive(MessagingEndpoint endpoint, boolean waitReceivers, Map<ClientArgument, Object> extraSenderArgs,
+                           Map<ClientArgument, Object> extraReceiverArgs, String senderAddress, String ... receiverAddresses) throws Exception {
+        int expectedMsgCount = 10;
+
+        ExecutorService executor = Executors.newFixedThreadPool(1 + receiverAddresses.length);
```
lulf, line 53: This should be shutdown and awaited after clients are finished (try - finally) so that resources are cleaned up.

systemtests/src/main/java/io/enmasse/systemtest/messaginginfra/resources/MessagingAddressResourceType.java
```diff
@@ -70,4 +72,13 @@ public static MessagingAddressCondition getCondition(List<MessagingAddressCondit
         }
         return null;
     }
+
+    public static int getPort(String protocol, MessagingEndpoint endpoint) {
```
lulf, line 76: MessagingAddressResourceType seems like the wrong place for this, put it in MessagingEndpointResourceType.

systemtests/src/test/java/io/enmasse/systemtest/sharedinfra/MessagingAddressTest.java
```diff
@@ -57,7 +51,8 @@
 
     private MessagingTenant tenant;
     private MessagingEndpoint endpoint;
-
+    private MessagingClientRunner clientRunner = new MessagingClientRunner();
```
lulf, line 54: Maybe rename this to MessagingClientManager and put it in the TestBase? (CC @kornys ). This should be setup and teardown for each test method.

systemtests/src/main/java/io/enmasse/systemtest/messaginginfra/resources/MessagingAddressResourceType.java
obabec: Wait, we already have this method, but written differently. I'll remove this and change tests to use that one in EndpointResource.

systemtests/src/main/java/io/enmasse/systemtest/messagingclients/MessagingClientRunner.java
obabec: Good catch, I have forget about that!

systemtests/src/test/java/io/enmasse/systemtest/sharedinfra/MessagingAddressTest.java
obabec: I am kind of scary to name it manager because that would way to start thinking about it as manager and I really don't know if I want to face situation that we have singleton manager for everything. But I am open to discussion, naming is hard :D

lulf: Yeah, I don't like the manager name maybe, but I think it should be something in the test base similar to the resource manager, that is properly cleaned up between tests etc.

systemtests/src/main/java/io/enmasse/systemtest/messagingclients/MessagingClientRunner.java
obabec: Should be fixed now :) Can you check it @lulf  ?

systemtests/src/test/java/io/enmasse/systemtest/sharedinfra/MessagingAddressTest.java
kornys: I agree with testBase, there is a space for this cleaning in before each/after each. Anyway please keep testBase clean :) I don't want to paste here common methods or asserts methods etc... :)

systemtests/src/main/java/io/enmasse/systemtest/messagingclients/MessagingClientRunner.java
```diff
@@ -5,31 +5,31 @@
...
 import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 
-import static io.enmasse.systemtest.messaginginfra.resources.MessagingAddressResourceType.getPort;
+import static io.enmasse.systemtest.messaginginfra.resources.MessagingEndpointResourceType.getPort;
```
kornys, line 26: I don't like this kind of imports, it causes bad readability in code becasue nobody know from which class this static method is, so please remove this import and use `MessagingEndpointResourceType.getPort()`

systemtests/src/test/java/io/enmasse/systemtest/sharedinfra/MessagingAddressTest.java
lulf: Are you saying assertSendMessageForServiceBrokerAddress does not belong there @kornys ? :rofl:

obabec: I am thinking that if we do it right we don't need Manager for clients because cleaning is usually not hard, it's just shutdown for executorService, so I would rather stay with this layer and see how it's working. I think that there will be more useful resources that would need that wdyt @lulf ?

lulf: @obabec Call it runner or whatever, but we need something per test that is maintaining the list of clients that got created, and 'clears' the list between test methods.

systemtests/src/test/java/io/enmasse/systemtest/TestBase.java
```diff
@@ -65,6 +68,7 @@
 
     @AfterEach
     void cleanup() throws Exception {
+        clientRunner.clean_clients();
```
lulf, line 71: Underscore

obabec: Aghh, school C destroyed me!

systemtests/src/main/java/io/enmasse/systemtest/messagingclients/AbstractClient.java
```diff
@@ -105,6 +107,14 @@ public JsonArray getMessages() {
         return messages;
     }
 
+    /**
+     * Get client capability such as sender/receiver.
```
lulf, line 111: I wonder, would 'role' be a better term than capability. When seeing 'capability' i think about AMQP source/target capabilities that you can set like 'queue' or 'topic'. I think in the future we may want to be able to pass this on to the client as well, so perhaps rename this now. The term 'role' is used throughout the AMQP 1.0 spec.

obabec: Okay, that sounds reasonable. Going to change it.

kornys:
@lulf @obabec maybe also move methods from MessagingEndpointTest like `doTestSendReceiveOnCluster` etc to new manager?

obabec:
@kornys I would like to do it in next PR. I don't want to let this PR stuck during weekend. If that is not a problem :)

PR #4884, opened 2020-07-03
[Console iot] - Add metrics for Iot projects

jbtrystram:
I am unsure how we can format the dates here. See general info section in this screen : https://marvelapp.com/5gbji22/screen/69215979


console/console-init/ui/mock-console-server/mock-console-server.js
```diff
@@ -1804,6 +1804,62 @@ function makeAddressSpaceMetrics(as) {
...
+      // value is limited to a float ?
+      value: "11/01/2020",
+      units: "Date"
+    },
+    {
+      name: "End date",
+      type: "counter",
+      // value is limited to a float ?
+      value: "11/01/2021",
+      units: "Date"
```
jbtrystram, line 1840: @k-wall any idea on how we could encode a date in there ? The graphQL type is float

k-wall: In the metadata we are exposing timestamps as String! formatted in the ISO-8601 manner, can you reuse the same format.

Sorry I missed this review request.

jbtrystram: Thanks

jbtrystram: what I mean is that in the Metric garphQL object the `value` field is typed as a float : 
```
type Metric_consoleapi_enmasse_io_v1beta1 {
  name: String!
  type: MetricType!
  value: Float!
  units: String!
}
```
https://github.com/EnMasseProject/enmasse/blob/2f1c035b1118d825dc66a8ee8e859e5ab0471315/console/console-server/src/main/resources/schema.graphql#L99

PR #4889, opened 2020-07-04
Misc cleanup

lulf:
* Update CSV
    * Remove old manifests
    * Use centos base images for init containers
    * Use static compile and scratch images for Go images
    * Remove references to removed components


kornys:
@lulf I think we can remove also inventory for ansible from systetests package

console/console-server/Dockerfile
```diff
@@ -3,7 +3,7 @@
 # License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
 #
 
-FROM centos:7
+FROM scratch
```
k-wall, line 6: Why scratch?

lulf: It would, I can leave the console server at centos:8 if you wish, the main motivation was to avoid any external dependency and reduce image size and need for respins etc.

lulf: Note that in the case of the console, the oauth proxy container would still have curl available and access to the same localhost ports.

lulf: An alternative would be to use some basic alpine image that has curl etc, but with a much smaller footprint than centos...

lulf:
I'll merge this now to unbreak master, and we can change the base image back to something that has the needed tools but with less size.

PR #4892, opened 2020-07-05
Add test to verify messages can be sent accross endpoints

lulf:


lulf:
@kornys Well, it takes advantage of the fact that all the logic in sendAndReceive copes with the fact that you can run it without receivers.

PR #4893, opened 2020-07-05
Update readme and intro gif

lulf:


kornys:
@lulf looks good, but I'm not sure if I like switching from white background to black and then back :-)

PR #4895, opened 2020-07-06
Rename MessagingTenant to MessagingProject

lulf:


kornys:
@lulf just stupid question why renaming from tenant to project? It sounds really openshift specific. From my POV is better naming MessagingTenant and IoTTenant

lulf:
@kornys The development team voted for MessagingProject and IoTTenant as the new names:

* Throughout the documentation, we have a persona named 'messaging tenant' which can create many MessagingTenant resources - this is a source of confusion we'd like to remove.
* For IoT, the term 'tenant' is used in Hono, so IoT should use IoTTenant. However, using tenant for both IoT and Messaging is confusing because they mean different things, as there are multiple IoT tenants per messaging tenant.

I also think project is more correct - because it is easier to understand the 1:1 mapping to a kubernetes namespace. I don't think its getting openshift-specific by just naming it project. The term also makes sense in a Kubernetes context IMO.

kornys:
I didnt vote🙂

kornys:
Maybe also messaging namespace is the good compromise, and then it is really 1:1 with kubernetes

lulf:
@kornys We discussed MessagingNamespace and came to the conclusion it would be a lot more confusing than project, because namespace is such a widely establised Kubernetes term that means a different thing.

kornys:
Ok then.

PR #4898, opened 2020-07-06
Add initial helm chart

lulf:
Not a lot of customization other than namespace right now, but its probably a nice installation option to have given that Helm is a widely used tool. We can probably expand options later on.


kornys:
I made simillar changes in my helm playground, anyway I think the folder should be named helm3 because it is more visible which version it is.

kornys:
And yes I agree that helm is widely used tool, for example strimzi users mainly use helm3 instead of bundle or operator hub.

kornys:
@lulf also there are missing some values in chart like icon, description, keywords, mainteners, source, home

PR #4900, opened 2020-07-06
Support selecting on namespaces

lulf:
* Add support for using label matching for messaging infrastructures
* Add support for using same match logic for messaging plans
* Add initial settings to plan spec as the first properties to be wired into
  the reconcile logic.
* Add initial logic in project reconciler to be able to select the
  desired plan.

There is not yet any logic to apply the plan, that is intended as a
phase 2.

@k-wall FWIW, it should be possible to use the matching logic for the identity provider match as well I suppose.


pkg/controller/messagingproject/messagingproject_controller_test.go
```diff
@@ -0,0 +1,142 @@
...
+	objs := []runtime.Object{
+		ns,
+	}
+	cl := fake.NewFakeClientWithScheme(s, objs...)
+	r := &ReconcileMessagingProject{client: cl}
+	return r
+}
+
+func assertSelector(t *testing.T, expected bool, selector *v1.NamespaceSelector) {
+	r := setup(t, &corev1.Namespace{
```
k-wall, line 35: I found the fact that `assertSelector` was producing/passing some of the test data pretty confusing at first sight.   I'd move the setup call and extract arg 2 of the r.matchesSelector call.

pkg/controller/messagingproject/messagingproject_controller.go
```diff
@@ -362,37 +431,144 @@ func (r *ReconcileMessagingProject) Reconcile(request reconcile.Request) (reconc
...
+	var bestMatch v1.Selectable
+	var bestMatchSelector *v1.NamespaceSelector
+
+	// Try to find a selector matching a specific name before needing to lookup namespace
+	for _, object := range objects {
+		selector := object.GetSelector()
 		// If there is a global one without a selector, use it
 		if selector == nil && bestMatch == nil {
-			bestMatch = &infra
+			bestMatch = object
```
k-wall, line 494: I don't think there should be more than one global on, but what happens if there are?

lulf: It would just select the first one. I can add so that it orders by creation date which is the resolution used for the case where multiple infras match on the labels or names.

k-wall: I think ordering by creation date would be best from consistency point of view.

pkg/controller/messagingproject/messagingproject_controller_test.go
lulf: Not sure what you meant with that last sentence, but let me know if the refactoring improves thing.

pkg/controller/messagingproject/messagingproject_controller.go
lulf: Done, also added tests for this.

PR #4906, opened 2020-07-07
Initial migration of IoT to shared infrastructure

ctron:
### Type of change

- Refactoring


obabec:
Nice work @ctron but I think that IoT Config and IoT Tenant should use resources manager same as other resources so IoTUtils class can drop all these things like create/delete/edit or whatever and test writer can rely on resources manager to delete config,tenant. WDYT @kornys ?

systemtests/src/main/java/io/enmasse/systemtest/iot/IoTUtils.java
```diff
@@ -0,0 +1,306 @@
...
+                .map(IoTConfigSpec::getMesh)
+                .map(MeshConfig::getServiceConfig)
+                .map(ServiceConfig::getReplicas)
+                .orElse(1);
+
+        TestUtils.waitUntilCondition("IoT Config to deploy", (phase) -> allThingsReadyPresent(expectedDeployments, expectedStatefulSets), budget);
+        TestUtils.waitForNReplicas(expectedDeployments.length + meshReplicas, config.getMetadata().getNamespace(), IOT_LABELS, budget);
+    }
+
+    public static void deleteIoTConfigAndWait(IoTConfig config) throws Exception {
```
obabec, line 99: As follow up to my comment. I think for example this method should be moved to resource manager with created proper resources

systemtests/src/main/java/io/enmasse/systemtest/iot/IoTUtils.java
```diff
@@ -0,0 +1,306 @@
...
+                }
+                return updated == null;
+            }, Duration.ofMinutes(5), Duration.ofSeconds(10), () -> "IoT project failed to delete in time");
+            log.info("IoTProject {} deleted", projectName);
+
+        }
+
+    }
+
+    public static void createIoTConfig(IoTConfig config) throws Exception {
```
obabec, line 249: Same with this one.

systemtests/src/main/java/io/enmasse/systemtest/iot/IoTUtils.java
kornys: resource manager is not here for storing methods like before, if you want to proper use new resource manager approach you need to create <CR>ResouceType where CR is name of resource like we did for all shared infra resources and implement common methods.

obabec: That is what I meant, sorry for my bad naming. I thought that with creation of proper Resource type which is afterwards handled by ResourceManager such as it is now for all other tests..

ctron:
@enmasse-ci run tests ocp4 profile=iot testcase=io.enmasse.systemtest.iot.isolated.http.HttpAdapterTest

> Nice work @ctron but I think that IoT Config and IoT Tenant should use resources manager same as other resources so IoTUtils class can drop all these things like create/delete/edit or whatever and test writer can rely on resources manager to delete config,tenant. WDYT @kornys ?

The problem with this is that you cannot re-configure the infrastructure for different use cases. Things like the "max payload" tests would require different test classes, and spread a single test aspect across different classes.

Also was it super hard in the past to find out who cleans up what, and why things don't get cleaned up properly. So the concept of setting up tests, and tracking cleanup actions, compared to a huge uber-cleaner was much easier to handle.

I am fine with inlining the methods from the IoTUtils class, so that only the IoTTestSession has access to that. Because that is something that shouldn't be spread through systemtests.

kornys:
@ctron if you check  new resource manager you will see that right now it is super easy to work with that and it is readable etc...

ctron:
> @ctron if you check new resource manager you will see that right now it is super easy to work with that and it is readable etc...

I just did. I think it still has the same problems as the old concept. There is single class class that has everything laid out as a single instance. I wouldn't be able to run multiple infrastructure configurations in a single test. There are no safeguards and cleanups for steps that fail. And it doesn't track any state that is relevant. So the only thing it would provide, would be a list of things that it runs at the end of a test.

Besides, we have tests that work. And side from a few early and weird ones, I would really like to keep them as we had them. I don't see much benefit in re-writing tests that work and are easy to understand.

If you have a nice concept for transforming them into some other pattern, then maybe make a proposal for that. But I think that is a "nice to have", compared to the task of creating something that works.

kornys:
@ctron I agree with do not use resource manager at this moment, but what I don't want is adding back interfaces which we used in systemtests previously like IoTTest, StandardIoTTest etc...

ctron:
> @ctron I agree with do not use resource manager at this moment, but what I don't want is adding back interfaces which we used in systemtests previously like IoTTest, StandardIoTTest etc...

Can you explain why?

kornys:
Because as you said on meeting, you dont like the fact you dont know where resource are deleted etc... I dont like the fact listing interfaces to find where we deploy default certs etc...

kornys:
And also I dont think we need shared iot tests anymore

ctron:
> Because as you said on meeting, you dont like the fact you dont know where resource are deleted etc... I dont like the fact listing interfaces to find where we deploy default certs etc...

So, what is your requirement here? There is a single class for IoT deploying the certs?

There are two classes `IoTTestSession` and the `IoTTests` base interface. I am happy to drop the latter, because that can easily go into the `IoTTestSession`.

> And also I dont think we need shared iot tests anymore

There are none anymore.

Both of the requests are fine with me, but I fail to see how that refers to the `StandardIoTTest`.

ctron:
@enmasse-ci run tests ocp4 profile=iot testcase=io.enmasse.systemtest.iot.isolated.http.HttpAdapterTest

ctron:
@enmasse-ci run tests ocp4 profile=iot testcase=io.enmasse.systemtest.iot.isolated.http.HttpAdapterTest

kornys:
> > Because as you said on meeting, you dont like the fact you dont know where resource are deleted etc... I dont like the fact listing interfaces to find where we deploy default certs etc...
> 
> So, what is your requirement here? There is a single class for IoT deploying the certs?
> 
I think IoTTestSession should deploy certs

> There are two classes `IoTTestSession` and the `IoTTests` base interface. I am happy to drop the latter, because that can easily go into the `IoTTestSession`.
> 
> > And also I dont think we need shared iot tests anymore
> 
> There are none anymore.
> 
ok perfect

> Both of the requests are fine with me, but I fail to see how that refers to the `StandardIoTTest`.

ctron:
@enmasse-ci run tests ocp4 profile=iot testcase=io.enmasse.systemtest.iot.isolated.http.HttpAdapterTest

ctron:
@enmasse-ci run tests ocp4 profile=iot

kornys:
@ctron just one question, could you point me please where you delete deployments of postgress and infinispan etc... ?

ctron:
@enmasse-ci run tests ocp4 profile=iot

ctron:
> @ctron just one question, could you point me please where you delete deployments of postgress and infinispan etc... ?

https://github.com/ctron/enmasse/blob/53aadc2736bc28ad0f99bb75f2fb4ec1b4cbb59d/systemtests/src/main/java/io/enmasse/systemtest/iot/IoTTestSession.java#L1073-L1082

kornys:
@ctron thank you! I don't have any objection against merging into master

PR #4908, opened 2020-07-07
Migrate do test clients

obabec:
### Type of change

- Refactoring

### Description
Refactoring continue on doTest methods with client uses. Migrated rest of doTest methods from current tests to clientRunner.

### Checklist

- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/systemtests/src/main/java/io/enmasse/systemtest/messagingclients/MessagingClientRunner.java
+++ b/systemtests/src/main/java/io/enmasse/systemtest/messagingclients/MessagingClientRunner.java
@@ -119,7 +119,7 @@ public class MessagingClientRunner {
         }
     }
 
-    public void endpointSendAndReceiveOnCluster(String host, int port, String address, boolean enableTls, boolean websockets) throws InterruptedException {
+    public void sendAndReceiveOnCluster(String host, int port, String address, boolean enableTls, boolean websockets) throws InterruptedException {
         assertTrue(port > 0);
         int expectedMsgCount = 10;
         executor = Executors.newFixedThreadPool(2);
--- a/systemtests/src/main/java/io/enmasse/systemtest/messagingclients/MessagingClientRunner.java
+++ b/systemtests/src/main/java/io/enmasse/systemtest/messagingclients/MessagingClientRunner.java
@@ -7,12 +7,25 @@ package io.enmasse.systemtest.messagingclients;
 import io.enmasse.api.model.MessagingEndpoint;
 import io.enmasse.systemtest.Endpoint;
 import io.enmasse.systemtest.framework.LoggerUtils;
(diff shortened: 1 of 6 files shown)


systemtests/src/test/java/io/enmasse/systemtest/TestBase.java
```diff
@@ -86,4 +89,18 @@ void cleanup() throws Exception {
             throw exception;
         }
     }
+
+    protected void assertDefaultMessaging() {
```
kornys, line 93: remove this. Test base is not for storing assertMethods anymore

systemtests/src/main/java/io/enmasse/systemtest/messagingclients/MessagingClientRunner.java
```diff
@@ -105,6 +119,69 @@ public void sendAndReceive(MessagingEndpoint endpoint, boolean waitReceivers, Ma
         }
     }
 
+    public void endpointSendAndReceiveOnCluster(String host, int port, String address, boolean enableTls, boolean websockets) throws InterruptedException {
```
kornys, line 122: please change name, I don't understand why it is named endpointSendEndReceive

systemtests/src/test/java/io/enmasse/systemtest/TestBase.java
obabec: Where you suggest to store this method afterwards? I don't see point why common assertMethod which is used in most of the tests shouldn't be there, moving it somewhere like testUtils is not right in my opinion.

kornys: AssertionUtils is a best approach I guess, create static class and store special Asserts there

kornys: and TestBase is not for storing common methods, we have testbase only for extend junit extension and configuration and basic clean in some case

systemtests/src/main/java/io/enmasse/systemtest/messagingclients/MessagingClientRunner.java
obabec: Changed, but I am thinking that we could maybe squash sendAndReceive with this method in some way. Need to take a look on that little bit more.

systemtests/src/test/java/io/enmasse/systemtest/TestBase.java
obabec: OK, thanks for explanation of current situation, I have create AssertionUtils for these as you suggested. Although I think that we should use as least static classes as possible.

obabec:
@lulf Sure they are and I think in process of creating new tests we will come up with options we want/need to configure and we will change functions to handle them :)

PR #4913, opened 2020-07-07
Requeue finalizer if resource is in use

lulf:
### Type of change

- Bugfix

### Description

This ensures that reconciler is called again to recheck finalization of
the resource.

@kornys I've not tested if this fixes your issue, but I could see from the operator logs that it failed to remove first because address was still present, but it didn't attempt to reconcile the endpoint again, so this should ensure that it tries again after 10 seconds. Usually it tries again on errors, but maybe that is not guaranteed by framework, so explicit scheduling is probably better.

### Checklist

- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/pkg/controller/messagingaddress/messagingaddress_controller.go
+++ b/pkg/controller/messagingaddress/messagingaddress_controller.go
@@ -217,7 +217,9 @@ func (r *ReconcileMessagingAddress) Reconcile(request reconcile.Request) (reconc
 					}
 					client := r.clientManager.GetClient(infra)
 					err = client.DeleteAddress(address)
-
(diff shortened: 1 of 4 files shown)


kornys:
@lulf so from code I see, if it is requeue, that means if will try again after 10 seconds and if it fails again it requeue it again? Like a infinite recursion?

lulf:
@kornys Correct, it will attempt indefinitely. IMO it shouldn't try to be clever and give up and go ahead with deletion, because it can lead to an inconsistent state.

kornys:
@lulf makes sense :-)

PR #4920, opened 2020-07-08
Fix endpoint management

lulf:
### Type of change

- Bugfix

### Description

Improve on the fix in  #4913 so that RequeueAfter is propagated so that reconcile loop does not continue if resource is in use which leaves it in a flip-flop state.

When having multiple endpoints with TLS certs, the secret containing the certificates get rewritten for each endpoint. The fix ensures that the secret is only initialized if it was not already.

When running tests with many endpoints and queues, the broker becomes unresponsive, so the default broker memory will be set to 1GB with a TODO to make this configurable in the MessagingInfrastructure later.

### Checklist

- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/pkg/controller/messagingendpoint/messagingendpoint_controller.go
+++ b/pkg/controller/messagingendpoint/messagingendpoint_controller.go
@@ -344,6 +344,7 @@ func (r *ReconcileMessagingEndpoint) reconcileFinalizer(ctx context.Context, log
 				err = client.DeleteEndpoint(endpoint)
 				if err != nil {
 					if errors.Is(err, stateerrors.ResourceInUseError) {
+						logger.Info("[Finalizer] Endpoint is still in use, rescheduling")
 						return reconcile.Result{RequeueAfter: 10 * time.Second}, nil
 					} else {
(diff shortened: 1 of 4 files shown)


pkg/controller/messagingendpoint/messagingendpoint_controller.go
```diff
@@ -344,6 +344,7 @@ func (r *ReconcileMessagingEndpoint) reconcileFinalizer(ctx context.Context, log
 				err = client.DeleteEndpoint(endpoint)
 				if err != nil {
 					if errors.Is(err, stateerrors.ResourceInUseError) {
+						logger.Info("[Finalizer] Endpoint is still in use, rescheduling")
```
kornys, line 347: maybe it should be warn? I'm not sure

lulf: There is no warn, only info

kornys: aha ok :D so then merge it! merge it! :D

PR #4930, opened 2020-07-08
Parallelization actions proposal

obabec:
Parallelization actions PR flow
### Type of change

- Enhancement / new feature

### Description
- Separate unit and systemtests
- Run them in parallel
- Unit tests kill systemtests run if failed
- We will have to change job requiements for merging after this PR

Next plan is to add IoT jobs run for jobs with IoT tag and also play with caching api-model from unit tests and somehow store images :) across jobs.

### Checklist

- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/.github/workflows/smoke.yml
+++ b/.github/workflows/smoke.yml
@@ -9,16 +9,56 @@ env:
 on: [pull_request]
 
 jobs:
-  pr_job:
-
+  unittests:
     runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@master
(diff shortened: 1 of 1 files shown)


obabec:
@lulf I have to play with that a little bit, because I am not sure, how cooperation with cache across these 2 jobs will go. Ideally if something changes in unit tests it would be good to change on system tests as well. So I can add that to unittests and maven for systemtests but I think we will have to redo it a little bit in future. Sounds good @lulf ?

PR #4940, opened 2020-07-09
[console-iot] Fix finished step issue

drcoolsanjeev:
### Type of change

- Bugfix

### Description

- Fix issue in last step of create project wizard. The flickering page between success and error state of creation 
- Add command query to render xml command in editor for iot project
image


console/console-init/ui/src/components/FinishedStep/FinishedStep.tsx
```diff
@@ -50,59 +50,72 @@ const FinishedStep: React.FunctionComponent<IFinishedStepProps> = ({
...
-            Enter your {projectType} Project for management, or return to
-            homepage to view all projects.
-          </EmptyStateBody>
-          <br />
-          <Link to={projectDetailUrl()}>
-            <Button variant={ButtonVariant.primary} component="a">
-              View the project
+      {success !== undefined &&
+        (!success ? (
+          <EmptyState
```
ajaypratap003, line 55: can we wrap this code into two seperate functions success and failure cases.
functions: renderSuccessState and renderFailedState. i think it would more cleaner code. wdyt ?

drcoolsanjeev: Agreed. Will make this change

drcoolsanjeev: Made the changes as suggested

PR #4942, opened 2020-07-09
Upgrade fabric8 to 4.10.2

kornys:


pom.xml
```diff
@@ -42,7 +42,7 @@
         <maven.compiler.target>11</maven.compiler.target>
 
         <avro.version>1.9.0</avro.version>
-        <fabric8.kubernetes-client.version>4.6.4</fabric8.kubernetes-client.version>
+        <fabric8.kubernetes-client.version>4.10.2</fabric8.kubernetes-client.version>
```
k-wall, line 45: I notice that fabric8 bumped its sundrio, shouldn't we stay in lockstep?
    <sundrio.version>0.21.0</sundrio.version>

kornys: ah, good point, thank you I will update that as well

PR #4944, opened 2020-07-10
Check conditions of addresses and endpoints as well

lulf:


systemtests/src/test/java/io/enmasse/systemtest/sharedinfra/MessagingInfrastructureTest.java
```diff
@@ -183,23 +183,29 @@ public void testInfraRestart(ExtensionContext extensionContext) throws Exception
...
 
         clientRunner.sendAndReceiveOnCluster(endpoint.getStatus().getHost(), endpoint.getStatus().getPorts().get(0).getPort(), anycast.getMetadata().getName(), false, false);
         clientRunner.sendAndReceiveOnCluster(endpoint.getStatus().getHost(), endpoint.getStatus().getPorts().get(0).getPort(), queue.getMetadata().getName(), false, false);
         AssertionUtils.assertDefaultMessaging(clientRunner);
     }
 
-    private void waitForConditionsTrue(MessagingInfrastructure infra) {
-        for (String condition : List.of("CaCreated", "RoutersCreated", "BrokersCreated", "BrokersConnected", "Ready")) {
-            waitForCondition(infra, condition, "True");
+    private void waitForConditionsTrue(MessagingInfrastructure infra, MessagingEndpoint endpoint, MessagingAddress ... addresses) {
```
kornys, line 193: @lulf can you move this method maybe into MessagingInfrastructureResoruceType class? I think we can also use it in other future tests

lulf: I think its a bit test specific now, maybe we can move it once it has proven to be useful just to make sure we keep the resource types a bit clean? I'm not sure what should/should not be in the resource type classes and if maybe we can make that more generic rather than having to create this for each type.

kornys: ah ok

PR #4947, opened 2020-07-10
Add acceptance profile into pom

kornys:
Added acceptance profile into systemtests pom


kornys:
@obabec yup sure, I will open PR where I need to rewrite whole Testing.md so I will write also about profiles etc...

obabec:
@kornys amazing :)

PR #4955, opened 2020-07-10
Proper openshift detection in Kubernetes class

kornys:


kornys:
@enmasse-ci run tests testcase=MessagingEndpointTest test_only crc

PR #4966, opened 2020-07-13
Move to new go

obabec:
### Type of change

- Enhancement / new feature
### Checklist

- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -38,7 +38,7 @@ jobs:
 
       - uses: actions/setup-go@v2
         with:
-          go-version: '1.13.1'
+          go-version: '1.14.3'
       - run: go version
 
       - name: setup-docker
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -40,7 +40,7 @@ jobs:
 
       - uses: actions/setup-go@v2
         with:
-          go-version: '1.13.1'
+          go-version: '1.14.3'
       - run: go version
 
       - name: setup-docker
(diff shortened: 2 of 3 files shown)


ctron:
I think you need to update the "go.mod"/"go.sum" file as well

obabec:
@ctron Why? This is just change on installed dependency on our machines. Updating go dependencies have nothing to do with actions I think.

obabec:
@ctron I can surely update go.mod and go.sum too, but I don't want to dig to "your" stuff and just prepare everything so that you can change that any time you want :)

ctron:
> @ctron I can surely update go.mod and go.sum too, but I don't want to dig into "your" stuff and just prepare everything so that you can change that any time you want :)

Its all "our" stuff :grin: … I think it makes sense to switch to the new version in a single PR … @k-wall what do you think?

obabec:
@ctron Ok, I don't have problem with that, I can bump go version here :)

ctron:
Ok, it looks like there is more than just the upgrade … go dependencies are the new NPM hell :roll_eyes: … I will take a look at using Go 1.14 and come back to this PR

ctron:
Ok … so #4974 would do the switch to use Go 1.14

So you've been right, we should have split up things :)

obabec:
@ctron ok cool, I'll revert one commit and this PR should be ready :)

kornys:
@ctron @obabec let me know if you need to merge it ASAP I can merge it before GH actions test complete

PR #4968, opened 2020-07-13
[Console-iot] - Unify all projects list results

jbtrystram:
### Type of change

Unifying the results to create a list of resources that contains `iotptojects` and `addressSpaces` and can be filtered and paged through.
This is done with adding a `kind` field to both of the objects.

- Enhancement

See README for how to use it.

### DIFF
--- a/console/console-init/ui/mock-console-server/README.md
+++ b/console/console-init/ui/mock-console-server/README.md
@@ -416,89 +416,115 @@ query single_address_with_links_and_metrics {
 }
 ```
 
-## Retrieve all Iot projects
+## Retrieve list of Iot tenants and messaging projects
 
 ```
-query allProjects {
-  allProjects(
-    projectType: iotProject
-  ) {
+query {
+  allProjects {
     total
-    iotProjects {
-      metadata {
-        name
-        namespace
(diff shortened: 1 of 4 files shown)


anukritijha:
@drcoolsanjeev FYI

console/console-init/ui/mock-console-server/README.md
```diff
@@ -416,89 +416,115 @@ query single_address_with_links_and_metrics {
...
-  ) {
+query {
+  allProjects {
     total
-    iotProjects {
-      metadata {
-        name
-        namespace
-        creationTimestamp
+    objects {
```
anukritijha, line 425: We would also need a property that signifies if a project is 'Ready'|'Configuring'|'Pending'|'Failed'

k-wall: I would expect that you'll get access to the .status section of both underlying objects.

jbtrystram:
> I think the approach looks reasonable, have you tried a POC in Go just to make sure there are no gotchas? I know @lulf was looking at this too, so he might have comments.

Starting a Go POC is next in my list :)

PR #4977, opened 2020-07-14
Remove nodejs installation on actions

obabec:
### Type of change

- Bugfix

### Checklist

- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/.github/scripts/install_dep.sh
+++ b/.github/scripts/install_dep.sh
@@ -12,9 +12,8 @@ sudo apt install maven
 echo "Install nodeJS"
 
 sudo apt-get install curl
-curl -sL https://deb.nodesource.com/setup_13.x
-sudo -E bash -
-sudo apt-get install nodejs
+curl -sL https://deb.nodesource.com/setup_12.x | sudo -E bash -
+sudo apt-get install -y nodejs yarn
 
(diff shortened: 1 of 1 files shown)


.github/scripts/install_dep.sh
```diff
@@ -12,9 +12,8 @@ sudo apt install maven
 echo "Install nodeJS"
 
 sudo apt-get install curl
-curl -sL https://deb.nodesource.com/setup_13.x
-sudo -E bash -
-sudo apt-get install nodejs
+curl -sL https://deb.nodesource.com/setup_12.x | sudo -E bash -
+sudo apt-get install -y nodejs yarn
```
k-wall, line 16: The Maven build actually installs node and yarn itself (and it controls the versions).  It is those that are used for the build.  What is this node actually used for?

obabec: @k-wall Really? I didn't know that, actually I am wondering what is bad on my local machine in that case because I've hit several times issue when my build fails with error missing nodejs :/

obabec: Ah I see via maven-front-end plugin, in that case I should remove this nodeJS installing all the way.

PR #4989, opened 2020-07-15
Console iot - graphQL API Device refactoring

jbtrystram:
That should make things easier !


console/console-server/src/main/resources/iot.schema.graphql
```diff
@@ -73,11 +74,20 @@ type IotEndpoint {
...
+  memberOf: [String!]
+  status: Device_status!
+  ext: String! #The Json representation of the ext field for this device.
   credentials: String! #A Json array with the devices credentials.
 }
 
+type Device_status {
+  created: String!
+  updated: String!
+  last_user: String!
```
rkpattnaik780, line 88: `lastSeen` field is missing, unable to get it in the device query.

jbtrystram: good catch ! It's updated

PR #5017, opened 2020-07-20
Validate metadata value based on type

anukritijha:
### Type of change

- Enhancement / new feature

### Description

Validate the value of metadata value based on Type during device Registration.

### Checklist

- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/console/console-init/ui/src/modules/iot-device/components/CreateMetadata/MetaDataRow.tsx
+++ b/console/console-init/ui/src/modules/iot-device/components/CreateMetadata/MetaDataRow.tsx
@@ -16,6 +16,7 @@ import { DropdownWithToggle } from "components";
 import { PlusIcon, MinusCircleIcon } from "@patternfly/react-icons";
 import { deviceRegistrationTypeOptions } from "modules/iot-device";
 import { DataType } from "constant";
+import { ValidationStatusType } from "modules/iot-device/utils";
 
 export interface IMetaDataRow {
   metadataList: any;
@@ -29,9 +30,9 @@ export const MetaDataRow: React.FC<IMetaDataRow> = ({
   rowIndex
 }) => {
   const metadataRow = metadataList[rowIndex];
-  const [validationStatus, setValidationStatus] = useState<"default" | "error">(
-    "default"
-  );
+  const [validationStatus, setValidationStatus] = useState<
+    ValidationStatusType.DEFAULT | ValidationStatusType.ERROR
+  >(ValidationStatusType.DEFAULT);
 
   const onSelectType = (typeValue: string) => {
     const validationStatus = getValidationStatus(typeValue, metadataRow.value);
@@ -44,22 +45,32 @@ export const MetaDataRow: React.FC<IMetaDataRow> = ({
   };
 
   const getValidationStatus = (type: string, value: string) => {
-    let validationStatus: "default" | "error" = "default";
+    let validationStatus:
+      | ValidationStatusType.DEFAULT
+      | ValidationStatusType.ERROR = ValidationStatusType.DEFAULT;
     switch (type) {
       case "string":
-        validationStatus = typeof value === "string" ? "default" : "error";
+        validationStatus =
+          typeof value === "string"
+            ? ValidationStatusType.DEFAULT
+            : ValidationStatusType.ERROR;
         break;
       case "number":
-        validationStatus = !isNaN(Number(value)) ? "default" : "error";
+        validationStatus = !isNaN(Number(value))
+          ? ValidationStatusType.DEFAULT
+          : ValidationStatusType.ERROR;
         break;
       case "boolean":
         validationStatus =
-          value.toLowerCase() === "true" || value.toLowerCase() === "false"
-            ? "default"
-            : "error";
+          value &&
+          (value?.toLowerCase() === "true" || value?.toLowerCase() === "false")
+            ? ValidationStatusType.DEFAULT
+            : ValidationStatusType.ERROR;
         break;
       case "datetime":
-        validationStatus = !isNaN(Date.parse(value)) ? "default" : "error";
+        validationStatus = !isNaN(Date.parse(value))
+          ? ValidationStatusType.DEFAULT
+          : ValidationStatusType.ERROR;
         break;
       default:
     }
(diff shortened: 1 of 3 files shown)


anukritijha:
@suyash-naithani FYI

console/console-init/ui/src/modules/iot-device/components/CreateMetadata/MetaDataRow.tsx
```diff
@@ -29,16 +29,43 @@ export const MetaDataRow: React.FC<IMetaDataRow> = ({
...
     if (isObjectOrArray(typeValue as any))
       updatedTypeMetadata[rowIndex].value = "";
     setMetadataList(updatedTypeMetadata);
   };
 
+  const getValidationStatus = (type: string, value: string) => {
+    let validationStatus: "default" | "error" = "default";
+    switch (type) {
+      case "string":
+        validationStatus = typeof value === "string" ? "default" : "error";
```
rkpattnaik780, line 50: A query here, if we want to put a number having the metadata type as string, it is still validating it, is that expected?

console/console-init/ui/src/modules/iot-device/components/CreateMetadata/MetaDataRow.tsx
```diff
@@ -29,16 +29,43 @@ export const MetaDataRow: React.FC<IMetaDataRow> = ({
   rowIndex
 }) => {
   const metadataRow = metadataList[rowIndex];
-  // console.log("metadataList", metadataList, "type", metadataRow.type);
+  const [validationStatus, setValidationStatus] = useState<"default" | "error">(
```
drcoolsanjeev, line 32: I think we should create a `enum` like `validStatusType` having two status for `default` and `error`

console/console-init/ui/src/modules/iot-device/components/CreateMetadata/MetaDataRow.tsx
drcoolsanjeev: I think in string we should skip validation. we can just do empty string check as everything in input value can be a string

console/console-init/ui/src/modules/iot-device/components/CreateMetadata/MetaDataRow.tsx
ajaypratap003, line 32: I think you can use ValidatedOptions for default/error/sucess. 
import {ValidatedOptions} from "@patternfly/react-core";

console/console-init/ui/src/modules/iot-device/components/CreateMetadata/MetaDataRow.tsx
```diff
@@ -29,16 +29,43 @@ export const MetaDataRow: React.FC<IMetaDataRow> = ({
...
+    const validationStatus = getValidationStatus(typeValue, metadataRow.value);
+    setValidationStatus(validationStatus);
     let updatedTypeMetadata = [...metadataList];
     updatedTypeMetadata[rowIndex].type = typeValue;
     if (isObjectOrArray(typeValue as any))
       updatedTypeMetadata[rowIndex].value = "";
     setMetadataList(updatedTypeMetadata);
   };
 
+  const getValidationStatus = (type: string, value: string) => {
```
ajaypratap003, line 46: I think validator name would be more meaningful. wdyt ?
This function should be part of top level utils.

console/console-init/ui/src/modules/iot-device/components/CreateMetadata/MetaDataRow.tsx
```diff
@@ -29,16 +29,43 @@ export const MetaDataRow: React.FC<IMetaDataRow> = ({
...
+    let validationStatus: "default" | "error" = "default";
+    switch (type) {
```
ajaypratap003, line 48: convert type into lower case for safer side. 
switch(type?.toLowerCase())

console/console-init/ui/src/modules/iot-device/components/CreateMetadata/MetaDataRow.tsx
```diff
@@ -29,16 +29,43 @@ export const MetaDataRow: React.FC<IMetaDataRow> = ({
...
+      case "number":
+        validationStatus = !isNaN(Number(value)) ? "default" : "error";
+        break;
+      case "boolean":
+        validationStatus =
+          value.toLowerCase() === "true" || value.toLowerCase() === "false"
+            ? "default"
+            : "error";
+        break;
+      case "datetime":
```
ajaypratap003, line 61: not sure, this `datetime` type check will work or not. pleas re-check for datetime.

console/console-init/ui/src/modules/iot-device/components/CreateMetadata/MetaDataRow.tsx
anukritijha: @drcoolsanjeev The check for string is needed here because if user first chose the type as number by mistake and entered a string as value, there would be a validation error shown on the screen. With this check in place, the validation error would be removed if the user changes the type back to string. Although, I agree that we should have an empty string check somewhere.

console/console-init/ui/src/modules/iot-device/components/CreateMetadata/MetaDataRow.tsx
anukritijha: String values are accepted values in 'validated' props.

console/console-init/ui/src/modules/iot-device/components/CreateMetadata/MetaDataRow.tsx
anukritijha: Not sure if this function has any use case beyond the metadata component, so I chose to keep it here the component itself.

console/console-init/ui/src/modules/iot-device/components/CreateMetadata/MetaDataRow.tsx
anukritijha: As this is not user input data, so there is no possibility for the case to change.

anukritijha: I have also used enum here, so it will further eliminate such issues.

console/console-init/ui/src/modules/iot-device/components/CreateMetadata/MetaDataRow.tsx
anukritijha: Agreed

anukritijha:
> The app gets broken if we change the metadata type to boolean before putting values inside the TextInputs. It throws this error message: `TypeError: value.toLowerCase is not a function`

@rkpattnaik780 This error occurred because the initial value is `[]` . I have fixed this issue in my next PR as it restructures the metadata Json.

PR #5018, opened 2020-07-20
Mock console - add metrics to iot projects

jbtrystram:
### Type of change

- Enhancement / new feature

### Description

Add some mock metrics to the mock server


console/console-init/ui/mock-console-server/mock-console-server.js
```diff
@@ -1794,6 +1794,60 @@ function makeAddressSpaceMetrics(as) {
...
+      type: "counter",
+      value: getRandomCreationDate(),
+      units: "Date"
+    },
+    {
+      name: "End date",
+      type: "counter",
+      value: getRandomCreationDate(),
+      units: "Date"
+    },
```
jbtrystram, line 1829: @k-wall As I stated on the previous PR : 

what I mean is that in the Metric garphQL object the `value` field is typed as a float : 

https://github.com/EnMasseProject/enmasse/blob/2f1c035b1118d825dc66a8ee8e859e5ab0471315/console/console-server/src/main/resources/schema.graphql#L99-L104

Should I add a string type there ?

PR #5024, opened 2020-07-20
[Console-IoT] Add ID to generic components

suyash-naithani:
### Type of change

- Refactoring

### Description

Add ID to generic components

### Checklist

- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/console/console-init/ui/src/components/About/About.tsx
+++ b/console/console-init/ui/src/components/About/About.tsx
@@ -59,7 +59,7 @@ export const About: React.FunctionComponent<IAboutProps> = ({
   if (docs) {
     docsElement = (
       <>
-        <TextListItem component="dt" id="about-documentation-dropdown">
+        <TextListItem component="dt" id="about-documentation-textlistitem">
           Documentation
         </TextListItem>
         <TextListItem component="dd">
--- a/console/console-init/ui/src/components/About/About.tsx
+++ b/console/console-init/ui/src/components/About/About.tsx
@@ -59,7 +59,7 @@ export const About: React.FunctionComponent<IAboutProps> = ({
   if (docs) {
     docsElement = (
       <>
-        <TextListItem component="dt" id="about-documentation-textlistitem">
+        <TextListItem component="dt" id="about-documentation-dropdown">
           Documentation
         </TextListItem>
         <TextListItem component="dd">
--- a/console/console-init/ui/src/components/About/About.tsx
+++ b/console/console-init/ui/src/components/About/About.tsx
@@ -59,7 +59,9 @@ export const About: React.FunctionComponent<IAboutProps> = ({
   if (docs) {
     docsElement = (
       <>
-        <TextListItem component="dt">Documentation</TextListItem>
+        <TextListItem component="dt" id="about-documentation-textlistitem">
+          Documentation
+        </TextListItem>
         <TextListItem component="dd">
           <a href={process.env.REACT_APP_DOCS}>{process.env.REACT_APP_DOCS}</a>
         </TextListItem>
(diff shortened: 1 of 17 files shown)


console/console-init/ui/src/components/About/About.tsx
```diff
@@ -59,7 +59,9 @@ export const About: React.FunctionComponent<IAboutProps> = ({
   if (docs) {
     docsElement = (
       <>
-        <TextListItem component="dt">Documentation</TextListItem>
+        <TextListItem component="dt" id="about-documentation-textlistitem">
```
anukritijha, line 62: I think in the UI it looks like a dropdown, so it should be better named as 'about-documentation-dropdown'.

console/console-init/ui/src/components/JsonEditor/JsonEditor.tsx
```diff
@@ -39,7 +39,7 @@ const JsonEditor: React.FunctionComponent<IJsonEditorProps> = ({
       <Flex>
         <FlexItem align={{ default: "alignRight" }}>
           <Tooltip
-            id={`tooltip-feedback-for-success-copy-${tooltipKey}`}
+            id={`json-editor-successfuly-copied-tooltip-${tooltipKey}`}
```
anukritijha, line 42: typo 'successfully'

console/console-init/ui/src/components/MetadataListTable/MetadataListTable.tsx
```diff
@@ -68,7 +68,10 @@ const TableHeader: React.FC<IHeaderProps> = ({
   return (
     <DataListItem aria-labelledby={ariaLabelledby}>
       <DataListItemRow>
-        <DataListItemCells dataListCells={dataListCells()}></DataListItemCells>
+        <DataListItemCells
+          id="metadata-list-item-cells"
```
anukritijha, line 72: If this is a row in the metadata list, it would be better named as 'metadata-list-row'. The idea is to name id as it appears in the UI, instead of using PF names.

console/console-init/ui/src/components/ServerMessageAlert/ServerMessageAlert.tsx
```diff
@@ -74,6 +74,7 @@ export const ServerMessageAlert: React.FC = () => {
     return (
       <PageSection>
         <Alert
+          id="server-mesage-error-alert"
```
anukritijha, line 77: Typo "message"

console/console-init/ui/src/components/About/About.tsx
```diff
@@ -59,7 +59,9 @@ export const About: React.FunctionComponent<IAboutProps> = ({
   if (docs) {
     docsElement = (
       <>
-        <TextListItem component="dt">Documentation</TextListItem>
+        <TextListItem component="dt" id="about-documentation-dropdown">
```
obabec, line 62: I think we should follow same pattern everywhere so if it's text list item it should be about-documentation-textlistitem

console/console-init/ui/src/components/FormAndJson/FormAndJson.tsx
```diff
@@ -41,7 +41,11 @@ const FormAndJson: React.FunctionComponent<IFormAndJsonProps> = ({
   return (
     <>
       {!isValid && <h1 style={{ color: "red" }}>inValid</h1>}
-      <Switch id="iot-enable-switch" isChecked={showJson} onChange={onToggle} />
+      <Switch
+        id="form-json-toggle-switch-button"
```
obabec, line 45: I don't know if it's necessary to have `toggle` keyword there because it's from my POV kind of irrelevant from point of testing, but if that makes sense to you go ahead and let it be with `toggle`

suyash-naithani: You're right. I'll make changes to this id

console/console-init/ui/src/components/About/About.tsx
suyash-naithani: Would it be more appropriate to have 'dropdown' at all places instead of 'textlistitem'?

obabec: I don't really care about that, that is decision what you in UI team have to do. I am just following other ID's which are already using textlistitem :) But if you want know my personal opinion I would go with dropdown in all places :)

suyash-naithani: I'd let this stay textlistitem then, just to make sure we have consistency throughout instead of finding all textlistitem and editing their id's :)

ajaypratap003: @suyash-naithani Please follow the consistency

PR #5040, opened 2020-07-21
Update console backend filter to parse IN query

jbtrystram:
This is an attempt to add `IN` as an operation in the console backend filtering engine but at the moment the go tests I added fail , it looks like the parser does not parse the IN statement as a legal expression.


k-wall:
The filter's grammar leans on JMS selector syntax, which in turn leans on SQL-92 specification.     JMS selector syntax was chosen as it has good traction in the messaging world so should be familiar to anyone working with EnMasse.

IN is defined by 3.8.1.1 Message Selector Syntax as:

```
identifier [NOT] IN (string-literal1, string-literal2,...) comparison operator
where identifier has a String or NULL value.
– Country IN (’ UK’, ’US’, ’France’) is true for ‘UK’ and false for ‘Peru’ it is equivalent to the expression (Country = ’ UK’) OR (Country = ’ US’) OR (Country = ’ France’)
– Country NOT IN (’ UK’, ’US’, ’France’) is false for ‘UK’ and true for ‘Peru’ it is equivalent to the expression NOT ((Country = ’ UK’) OR (Country = ’ US’) OR (Country = ’ France’))
– If identifier of an IN or NOT IN operation is NULL the value of the operation is unknown.
```

SQL-92 allows types other than strings literals:

https://docs.oracle.com/javadb/10.8.3.0/ref/rrefsqlj23075.html#rrefsqlj23075__sqlj34517

```
Expression [ NOT ]
    IN ( Expression
    [, Expression ]* )
```

So my first comment on the PR is that you are not implementing the SQL-92 grammar.  The parentheses are required and NOT IN should be supported too.

I also notice the PR seems to want to support for checking for an element present in an array argument.   As arrays types don't exist in SQL, I would suggest that trying to extend IN to accommodate those types would be breaking the rule of least surprise.  If you need that power, I'd suggest that implementing a separate functions is the way to go, something like:

CONTAINS(<value>, <array>)

pkg/consolegraphql/filter/parser.y
```diff
@@ -193,6 +204,16 @@ value:
     $$ = NewNullVal()
   }
 
+value_list:
+ '(' value ')'
+  {
+    $$ = NewValueList($2)
+  }
+| '(' value_list ',' value ')'
```
k-wall, line 212: production rule `'(' value_list ',' value ')'` looks odd.   wouldn't that imply nesting of the brackets was required i.e.  (('A'), 'B') rather than ('A','B').

PR #5044, opened 2020-07-22
Chore: handle order by dates for DeviceList

rkpattnaik780:
### Type of change

- Enhancement

### Description

Enable device list to be ordered by dates.

### Checklist

- [ ] Resolve all warnings
- [ ] Update license

### DIFF
--- a/console/console-init/ui/src/graphql-module/queries/iot_device.ts
+++ b/console/console-init/ui/src/graphql-module/queries/iot_device.ts
@@ -117,6 +117,8 @@ const SORT_RETURN_ALL_DEVICES_FOR_IOT_PROJECT = (sortBy?: ISortBy) => {
       case 1:
         orderBy = "`$.deviceId` ";
         break;
+      case 2:
+        break;
       case 3:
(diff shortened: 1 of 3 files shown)


console/console-init/ui/src/graphql-module/queries/iot_device.ts
```diff
@@ -117,12 +117,18 @@ const SORT_RETURN_ALL_DEVICES_FOR_IOT_PROJECT = (sortBy?: ISortBy) => {
       case 1:
         orderBy = "`$.deviceId` ";
         break;
-      case 2:
```
drcoolsanjeev, line 120: what will happen for case 2?

drcoolsanjeev: If we are not having sorting for that column. I would suggest to keep the case i.e `case 2: break;`

rkpattnaik780: Done.

PR #5051, opened 2020-07-23
Remove okhttp3 workaround for fabric8

kornys:
### Type of change

- Refactoring

### Description

Remove fabric8 workaround, this should be fixed in latest fabric8 client which we use

### Checklist

- [ ] Update documentation
- [ ] Update CHANGELOG.md


k-wall:
-1.  the defect we worked around was in okhttp3 not fabric8 and we still pull in an affected version.

[INFO] +- io.fabric8:kubernetes-client:jar:4.6.4:compile
[INFO] |  +- com.squareup.okhttp3:okhttp:jar:3.12.6:compile
[INFO] |  |  \- com.squareup.okio:okio:jar:1.15.0:compile

I think what I said here still holds:

https://github.com/EnMasseProject/enmasse/pull/4399#discussion_r417239705

PR #5054, opened 2020-07-23
Allow excluding IoT CRDs during Ansible installation (fixes #5046)

### LINKED ISSUE #5046: Ability to exclude EnMasse IoT components from Ansible install
**Overview**

For EnMasse 0.x only, when deploying from Ansible there needs to be an option to not to install the IoT components. 

**Design**

The Ansible manifest should accept a new parameter that governs if the IoT CRDs should be installed.  It should default to true to retain existing behaviour.   If set false the IoT CRDs should be omitted and the EnMasse operator configured disable the IoT controllers.

**Tasklist**

If/when you can, enumerate the tasks required to complete the feature in order to track progress as the feature is implemented.

ctron:
### Type of change

- Enhancement / new feature

### Checklist

- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/ansible/inventory/with-standard-authservice.txt
+++ b/ansible/inventory/with-standard-authservice.txt
@@ -4,6 +4,7 @@ localhost ansible_connection=local
 [enmasse:vars]
 namespace=enmasse-infra
 enable_rbac=True
+enable_iot=True
 service_catalog=False
 keycloak_admin_password=admin
 authentication_services=["standard"]
--- a/ansible/roles/cluster_admin/defaults/main.yml
+++ b/ansible/roles/cluster_admin/defaults/main.yml
@@ -1,4 +1,4 @@
 ---
 monitoring_operator: False
 monitoring: False
-enable_iot: False
+enable_iot: True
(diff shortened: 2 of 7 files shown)


k-wall:
@ctron tested locally and had to fix a couple of bugs.   it would for me now.

kornys:
right now this PR does not makes sense why we set defaults to enable iot?

k-wall:
> right now this PR does not makes sense why we set defaults to enable iot?

why would the default be to disable?  If I install from yaml bundle or OLM, I get IoT.  If I install from Ansible, I would expect the same.   We've added the option so that users of Ansible can elect to disable.  I don't think it should be the default.

kornys:
@k-wall because ansible is used only by integratly I think, and for testing they use our defaults inventory files, so it will cause troubles I guess if we change old behavior

k-wall:
@kornys - I'm just arguing against a change in the behaviour of of our release.  Previous releases defaulted to install IoT, so we should not break that.   If users such as Integrately need to disable the IoT install, they will have to create their own inventory - or pass additional options to the ansible-playbook command line to inject additional parameters.

lulf:
Release-0.33 branch was renamed to master: Please re-create PR against master branch.

PR #5063, opened 2020-07-24
Fix FirefoxConsoleTest#testOpenShiftWithCustomCert to work on newer OpenShift

k-wall:
### Type of change

- Bugfix

### Description

On newer OpenShift versions it seems that the oauth proxy endpoint gets the replacement
cert a significant time before the EnMasse console endpoint.  This causes the test to fail.

Test now awaits both endpoints to present the new cert.  Also guards against a stale selenium session.

### Checklist

- [ ] Update documentation
- [ ] Update CHANGELOG.md


k-wall:
@enmasse-ci run tests testcase=*brokered.web.*ConsoleTest tests_only crc

PR #5081, opened 2020-07-28
Console Iot - proposal to align device input schema with JSON editor view

jbtrystram:
The idea is to have the JSON editor for device to have the following template : 

```
{
  "id": "<optional-id>",
  "registration" : 
    {
      "enabled": true, 
      "defaults": {},
      "ext": {},
    },
  "credentials": []
}
```

This align the graphQL schemas to match this.

### DIFF
--- a/console/console-init/ui/mock-console-server/README.md
+++ b/console/console-init/ui/mock-console-server/README.md
@@ -1116,9 +1116,11 @@ mutation {
        namespace: "app1_ns"
     },
     device: {
-      deviceId: "Jens-phone"
-      enabled: true
-      ext: "{brand: samsung}"
+      deviceId: "Jens-phone",
+      registration : {
+        enabled: true
+        ext: "{brand: samsung}"
+      },
       credentials: "[{auth-id: \"pin\", type: \"password\", pwd-plain: \"1234\"}]"
     }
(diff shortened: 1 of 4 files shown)


jbtrystram:
So I was implementing this in the mock server and I have concerns because it leads to have transaction support in the backend server :

In hono credentials and device management are done through different API calls. It means that to execute the graphQL query the console backend needs to do two calls to the device registry.

So the most sensible way to implement this is to register the device data, register a callback to the result and register the credentials if the device was registered successfully. If the credentials call then fails, then should we delete the device that was created or forward an error message to the console frontend ? 
Can we make sure that we cover all the issues that can happens in error messages ?

ctron:
I wouldn't care too much about the mock server.

For the real backend, I would propose to create a new API in the Hono device management API, which allows to do a composite update of registration and credentials at the same time. And let the graphql server only access this API call.

jbtrystram:
> For the real backend, I would propose to create a new API in the Hono device management API, which allows to do a composite update of registration and credentials at the same time. And let the graphql server only access this API call.

and then the transaction work is delegated to the device registry implementation which is far easier. sounds good to me. Yet another API addition in Hono.

PR #5084, opened 2020-07-28
Prevent stale record from cache leading to incomplete address space deletion

k-wall:
### Type of change

- Bugfix

### Description

A race condition meant that stale values from the `EventCache` could cause the AddressSpaceController to partially recreate the address space controller during deletion.

The issue occurs as follows:

1. the Controller thread blocks trying to acquire mutex in method `ResourceChecker#onUpdate` (called from `EventCache#pop`).
2. The mutex is already held by `ResourceChecker#doWork` which is the controller's main loop.
3. By the time the controller's main loop terminates, the records in `io.enmasse.k8s.api.cache.EventCache#store` are stale.  Watcher updates *have* arrived (stored in the intermediate io.enmasse.k8s.api.cache.EventCache#queue) but these can't be processed as the Controller thread is blocked awaiting the mutex.
4. In the unlucky case, after the controller's loop terminates, `ResourceChecker#doWork` immediate reacquires the mutex.  It then releases the mutex as `Object#wait` is called.  This allows the Controller thread to acquire the mutex and notify the ResourceChecker. The ResourceChecker calls io.enmasse.k8s.api.ResourceCache#getItems which returns stale items which is then used as input to the Controller.
5. As the stale input has `metadata.deletionTimestamp` null, the controller proceeds to recreate secrets/configmaps of the address space.

The fix avoids the need for the Controller thread to block.  This will mean that the incoming updates from the watch are processed in timely way, and `#getItems` will be fresh. 

Also made two small changes to avoid noise in the log:

* prevent realm recreation for terminating address spaces
* prevent router status for terminating address spaces to avoid logged warnings

### Checklist

- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/address-space-controller/src/main/java/io/enmasse/controller/ControllerChain.java
+++ b/address-space-controller/src/main/java/io/enmasse/controller/ControllerChain.java
@@ -85,11 +85,9 @@ public class ControllerChain implements Watcher<AddressSpace> {
             return;
         }
 
-        List<AddressSpace> updatedResources = new ArrayList<>();
-
-
         boolean requeue;
         do {
+            List<AddressSpace> updatedResources = new ArrayList<>();
             requeue = false;
             for (final AddressSpace original : resources) {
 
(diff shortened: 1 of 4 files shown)


k-wall:
@enmasse-ci run tests ocp4

k-wall:
@lulf no urgency on the review of this change.

PR #5087, opened 2020-07-29
Log executor only if logging is enabled

kornys:
### Type of change

- Enhancement

### Description

Logging of running commands cause really mess in our logs, so I changed that behavior to log that command only if logging is enabled.

### DIFF
--- a/systemtests/src/main/java/io/enmasse/systemtest/executor/Exec.java
+++ b/systemtests/src/main/java/io/enmasse/systemtest/executor/Exec.java
@@ -35,7 +35,7 @@ import java.util.regex.Pattern;
  * Class provide execution of external command
  */
 public class Exec {
-    private static Logger log = LoggerUtils.getLogger();
+    private static final Logger log = LoggerUtils.getLogger();
     private Process process;
     private String stdOut;
     private String stdErr;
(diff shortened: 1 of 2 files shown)


systemtests/src/main/java/io/enmasse/systemtest/executor/Exec.java
```diff
@@ -35,7 +35,7 @@
  * Class provide execution of external command
  */
 public class Exec {
-    private static Logger log = LoggerUtils.getLogger();
+    private static final Logger LOGGER = LoggerUtils.getLogger();
```
ctron, line 38: I would prefer to keep this `log` as we have it in all the other classes as well.

kornys: in many classes it was already renamed to a LOGGER, so if I see class with log I try ti refactor that

systemtests/src/main/java/io/enmasse/systemtest/platform/Kubernetes.java
```diff
@@ -173,11 +173,11 @@ public static Kubernetes getInstance() {
             isOpenshift = Exec.execute(Arrays.asList(Kubernetes.getInstance().getCluster().getKubeCmd(), "api-resources"), false).getStdOut().contains("openshift.io");
             ocpVersion = instance.getOcpVersion();
 
-            log.info("Is CRC: {}", isCRC());
-            log.info("Is OpenShift: {}", isOpenShift());
-            log.info("Is OpenShiftCompatible: {}", isOpenShiftCompatible());
+            log.debug("Is CRC: {}", isCRC());
```
ctron, line 176: Because down here it is still `log`

kornys: yeah it is a different class :) ok I will move back to a log

PR #5106, opened 2020-08-04
Fix: change IDevice interface according to mock-server

rkpattnaik780:
### Type of change

- Bugfix

### Description

In IDevice interface, credential field has been removed and viaGroups have been added.

### Checklist

- [ ] Resolve all warnings
- [ ] Update license

### DIFF
--- a/console/console-init/ui/src/modules/iot-device/utils/TableFormatter.tsx
+++ b/console/console-init/ui/src/modules/iot-device/utils/TableFormatter.tsx
@@ -10,6 +10,7 @@ import { FormatDistance } from "use-patternfly";
 import { IRowData } from "@patternfly/react-table";
 import { IDevice } from "modules/iot-device/components";
 import { Label } from "@patternfly/react-core";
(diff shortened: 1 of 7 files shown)


console/console-init/ui/src/modules/iot-device/utils/TableFormatter.tsx
```diff
@@ -24,26 +25,14 @@ const renderDeviceType = (type: string) => {
...
-  const viaGateway = via && via?.length > 0;
+  let deviceType: DeviceConnectionType;
 
-  let deviceType: string;
-
-  if (!via?.length === !credentials?.length) {
-    deviceType = "N/A";
-  } else if (viaGateway) {
-    deviceType = "Via gateways";
+  if (via?.length || viaGroups?.length) {
```
rkpattnaik780, line 32: @jbtrystram since we can't get the credentials in the devices query, there is no way to check if a device has conflicting type, we will just be able to check the lengths of via & viaGroups and then determine the deviceType, is that okay.

PR #5111, opened 2020-08-05
Improve openshift 4 detection logic

dejanb:
### Type of change

- Bugfix

### Description

Server version call can return either Kubernetes or Openshift version, so we must account for both cases.


k-wall:
Relates to  https://bugzilla.redhat.com/show_bug.cgi?id=1861097

dejanb:
Underlying issue fixed in Openshift.

PR #5121, opened 2020-08-06
iot console mock server - generate device ID

jbtrystram:


console/console-init/ui/mock-console-server/mock-console-server.js
```diff
@@ -2552,10 +2552,12 @@ const resolvers = {
     },
 
     createIotDevice: (parent, args) => {
-      let reg = args.device.registration;
-      reg.deviceId = args.device.deviceId;
       let creds = args.device.credentials;
 
+      if (args.device.deviceId == undefined) {
```
drcoolsanjeev, line 2557: I think it should be `===` instead of `==`

PR #5112, opened 2020-08-05
Forwarder priority documentation

k-wall:
### Type of change

- Documentation

### Description

Documentation for the ability to specify a forwarder's priority.

### Checklist

- [ ] Update/write design documentation in `./documentation/design`
- [ ] Write tests and make sure they pass
- [ ] Update documentation
- [ ] Check RBAC rights for Kubernetes / OpenShift roles
- [ ] Try your changes from Pod inside your Kubernetes and OpenShift cluster, not just locally
- [ ] Reference relevant issue(s) and close them after merging
- [ ] Update CHANGELOG.md

PR #5137, opened 2020-08-17
[Docs] Forwarder priority documentation

lulf:
Re-created from https://github.com/EnMasseProject/enmasse/pull/5112

### DIFF
--- a/documentation/modules/ref-address-example-forwarders.adoc
+++ b/documentation/modules/ref-address-example-forwarders.adoc
@@ -15,7 +15,7 @@ To use an address forwarder, you must first configure a connector to the remote 
 
 Address forwarding is supported only in the `standard` address space type, and only for the `queue` and `subscription` address types. With the `queue` address type, you can forward messages to a remote AMQP server or from a remote AMQP server to a local queue. With the `subscription` address type, you can create a forwarder to a remote AMQP address, but you cannot create a forwarder that copies messages to the subscription. That is, the `subscription` address type supports forwarding in the `out` direction only, as shown in the example.
 
-For outgoing forwarders, it is possible to assign a `priority` value to the consumer associated with the forwarder. This allows you to influence the routing of messages within the AMQP network by favouring one link over another. If the forwarder has
+For outgoing forwarders, you can assign a `priority` value to the consumer associated with the forwarder. This allows you to influence message routing within the AMQP network by favoring one link over another. If the forwarder has
 a lower priority than other consumers, messages will only flow through the forwarder when the other consumers don't have credit or cannot accept a message for other reasons (e.g. messages selectors). The value must be an integral number in the range
 -2^31 to 2^31-1.  Values lower than zero reduces the forwarder's priority relative to other consumers while values greater than zero raise it.  If the priority is unassigned, a default value of 0 is used.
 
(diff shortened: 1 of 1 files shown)


documentation/modules/ref-address-example-forwarders.adoc
```diff
@@ -15,6 +15,10 @@ To use an address forwarder, you must first configure a connector to the remote
 
 Address forwarding is supported only in the `standard` address space type, and only for the `queue` and `subscription` address types. With the `queue` address type, you can forward messages to a remote AMQP server or from a remote AMQP server to a local queue. With the `subscription` address type, you can create a forwarder to a remote AMQP address, but you cannot create a forwarder that copies messages to the subscription. That is, the `subscription` address type supports forwarding in the `out` direction only, as shown in the example.
 
+For outgoing forwarders, it is possible to assign a `priority` value to the consumer associated with the forwarder. This allows you to influence the routing of messages within the AMQP network by favouring one link over another. If the forwarder has
```
hmaclean73, line 18:
```suggestion
For outgoing forwarders, you can assign a `priority` value to the consumer associated with the forwarder. This allows you to influence message routing within the AMQP network by favoring one link over another. If the forwarder has
```

documentation/modules/ref-address-example-forwarders.adoc
```diff
@@ -15,6 +15,10 @@ To use an address forwarder, you must first configure a connector to the remote
...
+a lower priority than other consumers, messages will only flow through the forwarder when the other consumers don't have credit or cannot accept a message for other reasons (e.g. messages selectors). The value must be an integral number in the range
```
hmaclean73, line 19:
```suggestion
a lower priority than other consumers, messages will only flow through the forwarder when the other consumers do not have credit or cannot accept a message for other reasons (for example, message selectors). The value must be an integral number in the range
```

documentation/modules/ref-address-example-forwarders.adoc
```diff
@@ -15,6 +15,10 @@ To use an address forwarder, you must first configure a connector to the remote
...
+-2^31 to 2^31-1.  Values lower than zero reduces the forwarder's priority relative to other consumers while values greater than zero raise it.  If the priority is unassigned, a default value of 0 is used.
```
hmaclean73, line 20:
```suggestion
-2^31 to 2^31-1.  Values lower than zero reduce the forwarder's priority relative to other consumers, while values greater than zero raise it.  If the priority is unassigned, a default value of 0 is used.
```

documentation/modules/ref-address-example-forwarders.adoc
```diff
@@ -35,14 +39,17 @@ spec:
   - name: f1 <1>
     remoteAddress: remote1/clients/me/1 <2>
     direction: out <3>
+    priority: 10 <4>
+
 ----
 <1> (Required) Specifies the name of the forwarder, which is used to ensure a unique identity.
 <2> (Required) Specifies the remote address to forward messages to. The address must be prefixed with the connector name and must be identical to the address matching patterns defined on the connector.
 <3> (Required) Specifies the direction of message flow, which is either `out` or `in`. A value of `out` forwards messages to the remote endpoint. A value of `in` forwards messages from the remote endpoint.
+<4> (Optional) Specifies the priority to be assigned to the consumer used by the forwarder (integral number in the range -2^31 to 2^31-1 with 0 being the default).  Applied to outgoing forwarders only.
```
hmaclean73, line 48:
```suggestion
<4> (Optional) Specifies the priority assigned to the consumer used by the forwarder (integral number in the range -2^31 to 2^31-1 with a default of 0).  Applies to outgoing forwarders only.
```

Issue #5130, opened 2020-08-11
Addresses fail to become ready if globalMaxSize is omitted from the infraconfig


k-wall:
**Describe the bug**

If globalMaxSize is omitted from the standardinfraconfig or brokeredinfraconfig, addresses associated with addressspaces using that infraconfig will fail to become ready.

Internally, the agent fails like this:

```
2020-08-11T15:19:22.832Z agent INFO  [broker-4fdeff9-seeu-0] all connectors exist
2020-08-11T15:19:22.841Z agent INFO  [broker-4fdeff9-seeu-0] Adding/updating address settings: myqueue
2020-08-11T15:19:22.847Z agent ERROR [broker-4fdeff9-seeu-0] Failed to create new address setting myqueue: null
2020-08-11T15:19:22.847Z agent ERROR [broker-4fdeff9-seeu-0] Failed to retrieve broker address setting: myqueue null
2020-08-11T15:19:22.847Z agent ERROR [broker-4fdeff9-seeu-0] failed to retrieve addresses: null
2020-08-11T15:19:22.847Z agent WARN  Rescheduling failed serialized func (null) in 5000ms.
```

Tracing the Agent's interaction with the Broker, the addAddressSetting management function fails on the Broker with  `java.lang.IllegalArgumentException`.    Examining with a debugger reveals that the maxSizeByte argument is null.  The reason this is null is because the Agent's calculation has produced a `NaN` value.   The `NaN` value occurs when the `globalMaxSize` is not passed via the  `BROKER_GLOBAL_MAX_SIZE` environment variable.

```
java.lang.IllegalArgumentException
        at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
        at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.lang.reflect.Method.invoke(Method.java:498)
        at org.apache.activemq.artemis.core.server.management.impl.ManagementServiceImpl.invokeOperation(ManagementServiceImpl.java:817)
        at org.apache.activemq.artemis.core.server.management.impl.ManagementServiceImpl.handleMessage(ManagementServiceImpl.java:441)
        at org.apache.activemq.artemis.core.server.impl.ServerSessionImpl.handleManagementMessage(ServerSessionImpl.java:2034)
[...]

PR #5159, opened 2020-08-25
Ensure the global-max-size is always defined when creating address-settings

vbusch:
### Type of change

- Bugfix

### Description

Fix: issue #5130 

Moved the calculation of the containers size out of the init-container and into the broker container to get the correct value.  Added scenario C) below.

A) If infraconfig specify globalMaxSize, it is used. (pre-existing)
B) If infraconfig do not specify globalMaxSize, but have the broker memory defined.  The launch scripts use it to set globalMaxSize to 1/8th. (pre-existing)
C) If infraconfig doesn't specify the globalMax size or the broker memory, then it is calculated as follows:
1) in the broker launch scripts, if the container memory is < the max unbounded memory total, we will use 1/8th of the container memory.
2) This is the new cases added.  In the case where it is crc or minishift, the container memory is too large.  In that case, we let the broker startup without any max heap settings, and the default jvm max is used.  The broker will set the default globalMaxSize to 1/2 the jvm max memory.  The agent will then get the broker's default globalMaxSize from the broker, and use 1/2 its size.  This is consistent with scenario 1 where the jvm heap size gets set to 1/2 the container size, and the brokers default globalMaxSize is 1/2 the jvm heap size.  And the globalMaxSize that we use is 1/8th the container size (or 1/2 of the broker's default globalMaxSize).

### Checklist

- [ ] Update documentation
- [x] Update CHANGELOG.md


agent/lib/broker_controller.js
```diff
@@ -73,6 +73,7 @@ BrokerController.prototype.on_connection_open = function (context) {
     var self = this;
     this.broker = new artemis.Artemis(context.connection);
     this.id = context.connection.container_id;
+    this.initializeGlobalMaxSizeFromBrokerIfUnset();
```
k-wall, line 76: Wouldn't it be better if this returned a promise and then you chained the actions that follow to the result?

broker-plugin/plugin/src/main/sh/launch-broker.sh
```diff
@@ -16,6 +16,14 @@ echo "JAVA_OPTS=${JAVA_OPTS}"
 
 sed -i 's/@JAVA_OPTS@/${JAVA_OPTS}/g' ${BROKER_CUSTOM_DIR}/conf/artemis.profile
 
+if [ "${GLOBAL_MAX_SIZE}" == "-1" ]; then
+    CONTAINER_MEMORY=$(cat /sys/fs/cgroup/memory/memory.limit_in_bytes)
+    MAX_MEM_UNBOUNDED=$(cat /proc/meminfo | grep 'MemTotal:' | awk '{print $2*1024}')
+    if [ ${CONTAINER_MEMORY} -lt ${MAX_MEM_UNBOUNDED} ]; then
```
k-wall, line 22: I think a comment here explaining the need for this code would help.

lulf (dismissed):

PR #5168, opened 2020-08-27
Improve default redelivery settings when a deadletter address is set

lulf:
### Type of change

- Enhancement / new feature

### Description

Default address settings will now be set to original Artemis defaults. I am not sure if it is better to use those, or to come up with a different set of settings.

### Checklist

- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/agent/lib/broker_controller.js
+++ b/agent/lib/broker_controller.js
@@ -776,6 +776,13 @@ BrokerController.prototype.generate_address_settings = function (address, global
         if (notTopic) {
             if (address.deadletter) {
                 addressSettings.DLA = address.deadletter;
+                // Specify default redelivery settings by default when using a DLQ. These are based
+                // on Artemis defaults.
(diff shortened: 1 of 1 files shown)


obabec (approved):

k-wall:
@lulf the intent was that the status section of the address would show the redelivery settings that are applied to the address.  I think we should consider moving the default back to the address controller to make them visible to the user.

lulf:
@k-wall I had not considered reflecting the status changes back to the user. Although I agree it would be nice to reflect the setting in the status as well, I don't think the "root" delivery settings would be shown in the status either. I'm not sure if it is that important, but I can fix it. Fixing it in agent would make this work for both standard and brokered was my reason for this fix.

obabec:
@lulf I didn't think about that either but @k-wall is right we should fix this so redelivery settings are visible to user.

PR #5188, opened 2020-09-07
[iot console]- Patch iot devices mutation 

jbtrystram:
### Type of change

- Enhancement / new feature

### Description

Added a graphQL mutation to edit iot devices using json patch. Multiples devices can be edited, the same patch will be applied to all.

### DIFF
--- a/console/console-init/ui/mock-console-server/mock-console-server.js
+++ b/console/console-init/ui/mock-console-server/mock-console-server.js
@@ -1967,7 +1967,7 @@ function createIotProject(pj) {
 
 function deleteIotProject(iotProject) {
   let pjIndex = getIotProjectIndex(iotProject.name);
-  let devIndex = getIotDevicesProjectIndex(iotProject.name);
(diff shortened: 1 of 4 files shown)


console/console-init/ui/mock-console-server/mock-console-server.js
```diff
@@ -2135,8 +2133,34 @@ function updateIotDevice(iotProject, device) {
   return iotdevices[pjIndex].devices[devIndex];
 }
 
+function patchIotDevice(iotProject, id, jsonPatch) {
+  let pjIndex = getIotDevicesIndexForProject(iotProject);
+  let devIndex = iotdevices[pjIndex].devices.findIndex(d => d.deviceId === id);
+
+  if (devIndex < 0) {
+    return {
```
ajaypratap003, line 2141: You should through a error rather than return object, as we had in other patch queries. Error format should be same as we had. In UI we have a common function for handling the server error, if you will change the error format it will cause issue in UI.

PR #5196, opened 2020-09-29
Modify address name to be different from address

lulf:
The agent uses the .spec.address as the key in its internal map, but the
purge request uses the .metadata.name as the key when doing the lookup.
This causes purge to fail if .metadata.name and .spec.address differ.

This fix changes the console server to pass the .spec.address instead of
.metadata.name and also using this as the lookup in agent.

An alternative would be to change the mapping internally to use
.metadata.name as the key, but this would complicate other uses of the
map that assumes it is keyed by .spec.address (such as router stats
retrieval).

The systemtest is updated to test purge with one address that has .metadata.name different from .spec.address.

### DIFF
--- a/agent/lib/console_server.js
+++ b/agent/lib/console_server.js
@@ -154,24 +154,19 @@ function ConsoleServer(env, openshift) {
             reject(context, 'amqp:unauthorized-access', 'not authorized');
         } else if (context.message.subject === 'purge_address') {
             accept();
-            var name = context.message.body.address;
-            log.info('[%s] purging address %s', user, name);
+            var addr = context.message.body.address;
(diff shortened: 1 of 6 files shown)


lulf:
@enmasse-ci run tests profile=acceptance testcase=standard.web.FirefoxConsoleTest

lulf:
@enmasse-ci run tests profile=acceptance testcase=standard.web.FirefoxConsoleTest

lulf:
@enmasse-ci run tests profile=acceptance testcase=standard.web.FirefoxConsoleTest

k-wall (approved):
LGTM

PR #5210, opened 2020-10-26
Make broker behavior on reject configurable

### LINKED ISSUE #5209: Make broker behavior on rejected messages configurable
**Description**

At present, the broker behaviour on reject is explicitly configured to treat it as modified, which means that an infinite retry will occur. In some cases, it is desirable that the message is dropped (0.33) or a global dlq (0.32) is used instead.

This option should be made configurable in the StandardInfraConfig.

lulf:
Expose configuration options for

* Treating rejected deliveries as modified (boolean, default: true)
* Responding with modified on transient delivery errors (boolean, default: true)
* Minimum size of messages to be considered large (integer, default: -1)

This change exposes these three options in the StandardInfraConfig
broker section.

Fixes #5209


lulf:
@enmasse-ci run tests profile=isolated testcase=InfraTestStandard

lulf:
@enmasse-ci run tests profile=isolated testcase=InfraTestStandard

lulf:
@enmasse-ci run tests profile=isolated testcase=InfraTestStandard

PR #5254, opened 2021-02-22
added subscription source var

jankalinic:
Signed-off-by: Jan Kalinic <jkalinic@redhat.com>
added env variable  and modified  subscription source


kornys (approved):
LGTM

PR #5302, opened 2021-04-27
ENTMQMAAS-2711 Makre sure pool-files is used on every possible broker.xml

clebertsuconic:
I am not sure what is required for a commit on this repository. if anything extra is required please import the commit and do a separate PR. Thanks


k-wall:
@clebertsuconic  thanks for the patch,  I plan to make this configureable, with the 10 as a default.

PR #5329, opened 2022-09-13
Fix #5328 : trigger reconfiguration of address space on change of authentication service trust

### LINKED ISSUE #5328: address space controller reconciliation doesn't consider auth CA changes
**Describe the bug**

There is an issue the address space control, if the user doesn't provide their own trust on the authentication service CR, copies the container's trust store /etc/ssl/certs/ca-bundle.crt and stores this in a secret [authservice-ca.${INFRA_UUID}](https://github.com/EnMasseProject/enmasse/blob/2887f204cb3d60d36a1714d4d5ee27f0e1cf088c/address-space-controller/src/main/java/io/enmasse/controller/TemplateInfraResourceFactory.java#L67). There's nothing in the operator that reconciles the contents of this secret, so if the container's trust changes, this won't ever be reflected in the secret until the next update is made to the address space's configuration that causes the update branch of [CreateController#reconcileActive()](https://github.com/EnMasseProject/enmasse/blob/master/address-space-controller/src/main/java/io/enmasse/controller/CreateController.java#L180) causes the resources to be recreated.

The controller really ought to be considering the equality of the key material from the source with the desired on its update check.  ~~It should also be using the injected trust from the OpenShift platform rather than relying on the trust from the container image~~ (edit: retracted last part - that would have broken use-cases where the external auth service is dialling out to a service secured by a public CA).

k-wall:
### Type of change

- Bugfix

### Description

Fix ##5328 : Adds digest of the authentication service CA (which may be platform trust from the container) onto applied config, so that any change to trust will trigger a reconcile and the internal `authservice-ca` to be repopulated.

### Checklist

- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/address-space-controller/src/main/java/io/enmasse/controller/AppliedConfig.java
+++ b/address-space-controller/src/main/java/io/enmasse/controller/AppliedConfig.java
@@ -6,9 +6,12 @@
 package io.enmasse.controller;
 
 import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+import java.util.Base64;
 import java.util.HashMap;
 import java.util.Objects;
 
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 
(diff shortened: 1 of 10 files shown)


vbusch (approved):
LGTM

vbusch (approved):
LGTM

